PR #21, opened 2021-08-27
Add ADR for testing frameworks

gnovv:
Created an ADR to document our initial decision to use spec and gomega for our testing framework on cf-on-k8s projects.

Co-authored-by: Matt Royal <mroyal@vmware.com>


matt-royal (approved):
LGTM

Issue #31, opened 2021-08-30
[Explore]: How should we structure k8s resource labels?


matt-royal:
### Background

For the CRD spike we used labels like `apps.cloudfoundry.org/appGuid` and `apps.cloudfoundry.org/processGuid`. These were copied over from past Eirini work, and we want to reconsider the correct domain and label structure. We also want to ensure that our labels conform with [the kubernetes label recommendations](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/)

### Acceptance Criteria

Write up an ADR as a PR with a proposed set of labels for the following cases:
- app GUID
- build GUID
- package GUID
- process GUID
- process type

### Number of days this exploration is expected to take. This is what is “pointed”.
Timeboxed to 1 day 

_No response_

tcdowney:
Closing: This was addressed in PR #39 and the ADR for this decision can be found here: https://github.com/cloudfoundry/cf-k8s-controllers/blob/203306701a4825584a514138bcd8cf73c877d7e4/controllers/docs/architecture-decisions/0005-structuring-kubernetes-resource-labels.md

Issue #9, opened 2021-08-17
Automatically set the appGUID label on CFApp records


matt-royal:
We encode the relationships between resources in two ways: via "ref" fields under `spec` and via `metadata.labels`. Rather than rely on our code and/or kubectl users to always remember both, we want to require the spec field and automatically set the label via a MutatingAdmissionWebhook

## Blockers
- #4 

## Acceptance Criteria
```
GIVEN that I am targeting a k8s cluster
AND  that I have followed the README instructions to install the web hook
WHEN I apply the example CFApp resource in the repo (confirm ahead of time that it doesn't provide a GUID label)
THEN I should see it succeeds
WHEN I fetch the yaml for that record from the k8s API
THEN I should see that the GUID label has been set on the record and that it matches the `metadata.name` field
```

## Dev Notes
- We should be able to accomplish this via a MutatingAdmissionWebhook. The application that backs that hook should be the same that supports the ValidatingAdmissionWebhook from earlier issues (unless we have a good reason not to).
- This should apply to creation and updates of records. One alternative to explore: can we only handle creation and then make the label immutable?

matt-royal:
The in-flight work for this is in the branch `issues/9`

davewalter:
> One alternative to explore: can we only handle creation and then make the label immutable?

The `Default()` method does not know whether a resource is being created or updated, and there is no way to raise an error, so we chose to always set the `appGuid` label to the value of the `metadata.name` field. There does not appear to be a way to make a set of labels immutable.

kieron-dev:
It seems likely that we'll need https://github.com/open-policy-agent/gatekeeper, or similar, to disallow changes to parts of some resources based on role.

If you find you need immutable labels, that should be simple enough to set up as an OPA rule.

angelachin:
Yes, labels are not immutable and are not expected to be.

Should the sample resource have the label now deleted since the mutating webhook now creates it?

Otherwise everything looks good to me.

akrishna90:
@angelachin Thats a good catch. Have addressed it in this [PR](https://github.com/cloudfoundry/cf-k8s-controllers/pull/33)

angelachin:
Awesome, looks good to me. Seems like this issue can be closed out.

PR #32, opened 2021-08-30
Set the appGUID label on CFApp records

akrishna90:
#9 
- the cfApp label key will need to be updated later based on what is agreed on this [story](https://github.com/cloudfoundry/cf-k8s-controllers/issues/31)


davewalter (approved):
LGTM!

PR #48, opened 2021-09-16
Add ADR for managing builders

acosta11:
https://github.com/cloudfoundry/cf-crd-explorations/issues/77

## What is this change about?
ADR capturing our initial strategy around the use of kpack builders

NA

## Acceptance Steps
Read and confirm it aligns with the team's understanding of builder management.

If accepted, update the status to accepted and merge.

@akrishna90

### DIFF
--- a/docs/architecture-decisions/0004-managing-builders.md
+++ b/docs/architecture-decisions/0004-managing-builders.md
@@ -1,0 +1,63 @@
+# Managing Builders
+
+Date: 2021-09-15
+
+## Status
+
+In Review
+
+## Context
+
+### CF App Staging Buildpack and Stack selection
+The CF API allows users to specify a list of `buildpacks` and a `stack` when `lifecycletype` is set to `buildpacks` while creating apps
+or while initiating staging by creating builds.
+
+As the lifecycle field is optional on both create app and create build requests, we want a default available.
+
+If lifecycle data is specified, the precedence we follow is `build lifecycle > app lifecycle > default lifecycle`.
+
+On Kubernetes we use `Kpack` for staging. In order to do that we need a `kpack builder` referencing a set of `buildpacks` and `stacks`. To provide default `buildpacks` and `stacks` for app staging, we define a default kpack builder.
+
+Below are the possible permutation of lifecycle scenarios and kpack builders to use
+
+- App created with empty lifecycle & Build created with empty lifecycle
+    - Default builder is used
+- App created with lifecycle data & Build created with empty lifecycle
+  - Check if a builder exists within the same namespace referencing only the app-guid.
+    - if it exists, the builder referencing only the app-guid is reused.
+    - if its doesn't exist, a new builder with specified `buildpacks` and `stacks` is created.
+- Build created with lifecycle data
+    - A new builder with specified `buildpacks` and `stacks` is created
+
+### CF Stack and Buildpack Updates
+On CF for VMs, when we update a stack, running apps are updated to use the updated stack. Note that for now, we are excluding droplet rollback concerns from our decision.
+
+On CF for VMs, when we update a buildpack, running apps are not automatically rebuilt to use the new buildpack.
+
+After a stack is updated, subsequent builds will use the new stack, so we focus on the cases of running apps and assigning droplets for which there is not necessarily a required restage.
+
+## Decision
+### Default Cluster Builder
+Operator is responsible for creating a default cluster builder, and it has be named as `cf-kpack-cluster-builder`.
+Operator is also responsible for creating service-account & secrets for access to a single registry on each namespace.
+We expect the operators to follow `${namespace}-kpack-service-account` naming convention for service accounts.
+
+### Handling Stack and Buildpack Updates
+At kpack Builder update time (for both Stack and Buildpack), the kpack controller will see the changes, find affected kpack Images, rebuild them, and add a new status on the Image with the reason for rebuild. A CF controller will see the changes on the kpack Image status, find affected CFApps by builder guid label, and update the droplet image ref of the associated CFBuild with the newly built image ref.
+
+The CF controller must intentionally ignore changes triggered by buildpack updates. When a buildpack is updated, the operator is responsible for explicitly rebuilding apps that rely on that buildpack as desired.
+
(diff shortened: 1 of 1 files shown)


docs/architecture-decisions/0004-managing-builders.md
```diff
@@ -0,0 +1,63 @@
...
+### Handling Stack and Buildpack Updates
+At kpack Builder update time (for both Stack and Buildpack), the kpack controller will see the changes, find affected kpack Images, rebuild them, and add a new status on the Image with the reason for rebuild. A CF controller will see the changes on the kpack Image status, find affected CFApps by builder guid label, and update the droplet image ref of the associated CFBuild with the newly built image ref.
+
+The CF controller must intentionally ignore changes triggered by buildpack updates. When a buildpack is updated, the operator is responsible for explicitly rebuilding apps that rely on that buildpack as desired.
+
+### App-Specific Builders
+In the context above we have captured details about when builders other than the default cluster builder are created. That behavior may be implemented in a subsequent iteration alongside the lifecycle data on CFApp and CFBuild resources.
+At creation time for such a builder we will label it with `app-guid` and/or `build-guid`.
+Additionally, we will track the readiness of those builders using `status` conditions on the builder.
+We will not initially clean up builders or associated images.
```
tcdowney, line 54: I think in terms of the Builders themselves we could have "cheap" clean-up by just getting the `ownerReferences` correct.

* App owns "app lifecycle" builder
* Build owns "build lifecycle" builder

In terms of the images in the registry themselves I'm not sure if we need to manage them since they're owned by kpack IMO.

docs/architecture-decisions/0004-managing-builders.md
```diff
@@ -0,0 +1,63 @@
...
+## Decision
+### Default Cluster Builder
+Operator is responsible for creating a default cluster builder, and it has be named as `cf-kpack-cluster-builder`.
+Operator is also responsible for creating service-account & secrets for access to a single registry on each namespace.
+We expect the operators to follow `${namespace}-kpack-service-account` naming convention for service accounts.
+
+### Handling Stack and Buildpack Updates
+At kpack Builder update time (for both Stack and Buildpack), the kpack controller will see the changes, find affected kpack Images, rebuild them, and add a new status on the Image with the reason for rebuild. A CF controller will see the changes on the kpack Image status, find affected CFApps by builder guid label, and update the droplet image ref of the associated CFBuild with the newly built image ref.
+
+The CF controller must intentionally ignore changes triggered by buildpack updates. When a buildpack is updated, the operator is responsible for explicitly rebuilding apps that rely on that buildpack as desired.
```
tcdowney, line 48: > When a buildpack is updated, the operator is responsible for explicitly rebuilding apps that rely on that buildpack as desired.

Someone (I think @gnovv ?) pointed out that this works fine until the next stack update and then that image will have both the buildpack and stack updates. Do y'all know if there is anyway to configure a kpack image to ignore certain types of rebuilds?

tcdowney: @Shanfan @emalm we're thinking of ways here to maintain parity with CF for VMs "stemcell bumping"/stack updating workflow where droplets automatically get scheduled with the new base OS after an operator upgrades those. Traditionally that didn't include buildpack changes and apps needed to be manually restaged to pick up those changes. Alternatively we could just embrace kpack fully and accept any and all image updates it comes with. This has a chance of being more risky for app stability since we believe buildpack updates are inherently more risky for the app than stack updates. Do y'all have thoughts about these potential UX impacts?

tcdowney (approved):

docs/architecture-decisions/0004-managing-builders.md
emalm: I do think that updating only the base OS/"stack" layer(s) would be safest and most consistent with the current behavior of CF on VMs, if we can arrange it, because of the ABI compatibility promises we get from those OS-layer updates. Secondarily, I could see it being valuable to allow developers to opt an app into having its image rebuilt in response to a buildpack update.

That's an interesting point about kpack now managing both of those updates in an interleaved order, as opposed to the distinct stack-update and droplet-rebuild mechanisms in CF on VMs. I recall from years ago that one part of the CNB model would be the ability to update a complete app image just by updating the image data structure to swap in new versions of layers that were known to be compatible (such as ABI-compatible OS layers), and that that update process could happen efficiently in the OCI image registry itself since it didn't require any new layers to be generated or transferred beyond uploading the new base layers into the registry. Do you know whether kpack can do that yet, or if not, whether they're still thinking about stack or even dependency updates in terms of registry-side manipulations of metadata? If so, that could be a way of still keeping the safer stack updates separate from the more expensive and riskier re-application of the buildpack processing.

tcdowney: > Do you know whether kpack can do that yet, or if not, whether they're still thinking about stack or even dependency updates in terms of registry-side manipulations of metadata?

@matthewmcnew do you happen to know? Or apart from that any other way of automated kpack rebuilds watching for stack updates only?

matthewmcnew: Today, kpack (and CNBs with the pack cli) will perform a rebase operation by only writing a new image manifest on the registry which replaces the old stack layers. This is not possible for non-stack layers (buildpack provided layers) as they are not ABI compatible. Because of this kpack schedules a full rebuild with buildpack updates whereas stack updates do not require a full rebuild.

matthewmcnew: There is not a way to mark an Image to only receive stack updates. You could pin the buildpack version in the in the corresponding builder so that new buildpacks added to the store do not schedule builds (but, then the builder would not be updated either). Alternatively, you could use the `Build` resource instead of the `Image` resource to have more control over when images get rebuilt. This would require "rebase" builds to be explicitly scheduled when stack updates are needed.

docs/architecture-decisions/0004-managing-builders.md
matthewmcnew: > they're owned by kpack IMO.
kpack does not delete images in the registry itself as not all registries support image deletion.

PR #51, opened 2021-09-17
Implement CFProcess mutating webhook to add labels

### LINKED ISSUE #29: [Feature]: API Client can see that Process labels are set automatically
### Blockers/Dependencies

- #25 

### Background

As an API Client
I want the appGUID, processGUID and processType labels on CFProcess to be set automatically
So that I can look up CFProcess records by these fields

We encode the relationships between resources in two ways: via "ref" fields under `spec` and via `metadata.labels`. Rather than rely on our code and/or kubectl users to always remember both, we want to require the spec field and automatically set the label via a MutatingAdmissionWebhook

### Acceptance Criteria

GIVEN that I am targeting a k8s cluster
AND  that I have followed the README instructions to install the web hook
WHEN I apply the example CFApp and CFProcess resources from the `config/samples` directory
THEN I should see all operations succeed
WHEN I fetch the yaml for the CFProcess record from the k8s API
THEN I should see that the appGUID label has been set on the record and that it matches the `spec.appRef.name` field
AND  I should see that the processGUID label has been set on the record and that it matches the `metadata.name` field
AND  I should see that the processType label has been set on the record and that it matches the `spec.processType` field

### Dev Notes

- Remove the labels from the example resources if they are still present, and commit the change.
- This should apply to creation and updates of records

davewalter:
Resolves #29 

## What is this change about?
This PR implements a mutating webhook for the CFProcess resource to set the appGUID, processGUID and processType labels.

No

## Acceptance Steps
Please see the steps in #29


apis/workloads/v1alpha1/cfprocess_webhook_integration_test.go
```diff
@@ -0,0 +1,142 @@
...
+			cfProcessLookupKey := types.NamespacedName{Name: cfProcessGUID, Namespace: namespace}
+			createdCFProcess := new(v1alpha1.CFProcess)
+
+			g.Eventually(func() map[string]string {
+				err := k8sClient.Get(ctx, cfProcessLookupKey, createdCFProcess)
+				if err != nil {
+					return nil
+				}
+				return createdCFProcess.ObjectMeta.Labels
+			}, 10*time.Second, 250*time.Millisecond).ShouldNot(BeEmpty(), "CFProcess resource does not have any metadata.labels")
```
akrishna90, line 63: On integration tests for cfbuild mutating webhook, me and @acosta11 decided to push this into a `before` function and grouped all assertions inside a `it` func. The reasoning is that we expect the created records to have all the labels we expect, so there is value in checking existence of all labels at once. It also reduces test execution time.

Also since we are using the same `guid` to create records across tests, shouldn't these fail? I might be missing something here

akrishna90 (approved):
lgtm!

acosta11 (approved):
lgtm

Issue #43, opened 2021-09-15
[Feature]: API Client can see that a CF Build + kpack Image & kpack Build results in a CF Droplet


gnovv:
### Blockers/Dependencies

BLOCKED BY: https://github.com/cloudfoundry/cf-k8s-controllers/issues/42
BLOCKED BY: https://github.com/cloudfoundry/cf-k8s-controllers/issues/52

### Background

As an API Client
I want a CF Build along with its kpack Image & Build CRs to result in a CF Droplet automatically
So that I can look up the latest built image by inspecting the CF Build and its Droplet

### Acceptance Criteria

GIVEN that I am targeting a k8s cluster
AND that I have followed the README instructions to install the Controllers and set up the kpack Registry Credentials and staging CRs
WHEN I apply the example CFApp, CFPackage, and CFBuild resources in config/samples directory
THEN I should see all those operations succeed
WHEN I fetch the yaml for the CF Build CR after its kpack Build completes successfully
THEN I should see that it now contains the Droplet's image

### Dev Notes
- Droplet processTypes and ports will be updated at the same time in a future story https://github.com/cloudfoundry/cf-k8s-controllers/issues/44
- Feel free to refer to the spike implementations of the [Build Controller](https://github.com/cloudfoundry/cf-crd-explorations/blob/main/controllers/build_controller.go) and the [cf-to-kpack Build Controller](https://github.com/cloudfoundry/cf-crd-explorations/blob/main/controllers/cf_kpack_build_controller.go) for inspiration
- Updating the Droplet's Process fields will be handled in a followup issue
 - For now ignore when a kpack Image is rebuilt and a new kpack Build is created after the first one, do not overwrite the droplet image

gnovv:
- [ ] Happy Path
  - [ ] Unit Test
  - [ ] Implementation
- [ ] Unhappy Path
  - [ ] Unit Test
  - [ ] Implementation
- [ ] Decide if Integration Tests are required
- [ ] Test in real cluster
- [ ] Create PR
- [ ] Merge PR

PR #64, opened 2021-09-30
Set CFDroplet fields on CFBuild.Status when Kpack Image succeeds

akrishna90:
#43 

## What is this change about?
CFBuild Controller can now watch kpack images for completion. Based on the completion status, it now updates CFDroplet fields on CFBuild's status.  

No.

## Acceptance Steps
Follow acceptance steps on issue #43

### DIFF
--- a/controllers/workloads/cfbuild_controller.go
+++ b/controllers/workloads/cfbuild_controller.go
@@ -17,9 +17,16 @@ limitations under the License.
 package workloads
 
 import (
+	"context"
+	"strings"
+
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+
 	workloadsv1alpha1 "code.cloudfoundry.org/cf-k8s-controllers/apis/workloads/v1alpha1"
 	cfconfig "code.cloudfoundry.org/cf-k8s-controllers/config/cf"
-	"context"
 	"github.com/go-logr/logr"
 	buildv1alpha1 "github.com/pivotal/kpack/pkg/apis/build/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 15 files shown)


controllers/workloads/cfbuild_controller.go
```diff
@@ -90,25 +94,64 @@ func (r *CFBuildReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+				r.Log.Error(err, "Error when updating CFBuild status")
+				return ctrl.Result{}, err
+			}
+		}
 	}
-
 	return ctrl.Result{}, nil
 }
 
 func (r *CFBuildReconciler) createKpackImageAndUpdateStatus(ctx context.Context, cfBuild *workloadsv1alpha1.CFBuild, cfApp *workloadsv1alpha1.CFApp, cfPackage *workloadsv1alpha1.CFPackage) error {
```
tcdowney, line 152: I see internally this calls `createKpackImageIfNotExists` -- what is the behavior if a kpackImage does exist? Should we be calling a `createOrUpdate` type method instead?

reference/cf-k8s-controllers.yaml
```diff
@@ -1323,7 +1323,7 @@ subjects:
 ---
 apiVersion: v1
 data:
-  cf_k8s_controllers_config.yaml: '"kpackImageTag": "gcr.io/cf-relint-greengrass/cf-crd-staging-spike/"'
+  cf_k8s_controllers_config.yaml: 'kpackImageTag: gcr.io/cf-relint-greengrass/cf-k8s-controllers/kpack/beta'
```
tcdowney, line 1326: What is this reference directory used for?

controllers/workloads/cfbuild_controller.go
```diff
@@ -90,25 +94,64 @@ func (r *CFBuildReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+		// Retrieves and Checks Kpack Image Status Condition for Type "Succeeded"
+		// If NotFound error - Ignore and return
+		// If Found, check Succeeded status condition
+		// If Succeeded is True - Update Status Conditions and Droplet fields on CFBuild
+		// If Succeeded is False - Update Status Conditions on CFBuild
+		var kpackImage buildv1alpha1.Image
+		err = r.Client.Get(ctx, types.NamespacedName{Name: cfBuild.Name, Namespace: cfBuild.Namespace}, &kpackImage)
+		if err != nil {
+			r.Log.Error(err, "Error when fetching Kpack Image")
+			//Ignore Image NotFound errors to account for eventual consistency
```
tcdowney, line 124: What triggers the reconciler to rerun in this case?

tcdowney (commented):
Looks reasonable to me, just left a few comments to better my understanding. :)

gnovv (approved):
lgtm!

reference/cf-k8s-controllers.yaml
akrishna90: this directory will store runnable app images created as a result of kpack build.
I updated the path to reflect the codebase that is pushing images.

controllers/workloads/cfbuild_controller.go
akrishna90: kpackImage "exists" condition is encountered when we were successful in creating `kpack image` but encountered an error while updating the status on the `cfbuild`. Hence, when we reconcile again, we continue from where we left and just update the status. 
What would have changed on the CFBuild that would be valuable to run an update? 

`createOrUpdate` is not part of `client.Client` pkg. CFClient interface we have defined will have to change to accommodate for it.

PR #68, opened 2021-10-01
Separate int tests

gnovv:
After converting tests from spec to ginkgo, this PR separates unit and integration tests by moving integration tests to a separate package.

Acceptance:
make test

### DIFF
--- a/apis/workloads/v1alpha1/integration/webhook_suite_integration_test.go
+++ b/apis/workloads/v1alpha1/integration/webhook_suite_integration_test.go
@@ -1,4 +1,21 @@
-package workloads_test
+/*
+
+Copyright 2021.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package integration_test
 
 import (
 	"context"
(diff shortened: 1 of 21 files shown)


apis/workloads/v1alpha1/integration/webhook_suite_integration_test.go
```diff
@@ -0,0 +1,132 @@
...
+	k8sClient client.Client
+)
+
+func TestWorkloadsMutatingWebhooks(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Workloads Mutating Webhooks Integration Test Suite")
+}
+
+var _ = BeforeSuite(func() {
+	logf.SetLogger(zap.New(zap.WriteTo(os.Stderr), zap.UseDevMode(true)))
```
matt-royal, line 57: Now that we're back in Ginkgo-land, `os.Stderr` should be `GinkgoWriter` here

controllers/networking/integration/suite_integration_test.go
```diff
@@ -0,0 +1,79 @@
...
+	k8sClient client.Client
+)
+
+func TestNetworkingControllers(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Networking Controllers Integration Suite")
+}
+
+var _ = BeforeSuite(func() {
+	logf.SetLogger(zap.New(zap.WriteTo(os.Stderr), zap.UseDevMode(true)))
```
matt-royal, line 33: Ditto that this should be `GinkgoWriter`

controllers/workloads/integration/cfbuild_controller_integration_test.go
```diff
@@ -264,3 +265,10 @@ var _ = Describe("CFBuildReconciler Integration Tests", func() {
 		})
 	})
 })
+
+func setKpackImageStatus(kpackImage *buildv1alpha1.Image, conditionType string, conditionStatus string) {
```
matt-royal, line 269: Where did this come from? This wasn't removed from elsewhere and no usages of it seem to have been added.

matt-royal (changes requested):

controllers/workloads/integration/cfbuild_controller_integration_test.go
gnovv: It gets used in both the unit and integration tests, but I didn't want to add it to the sharedtest utils because other tests are depending on that package too.

PR #75, opened 2021-10-06
Disable webhooks when running locally

Birdrock:
## What is this change about?
Disable webhooks when running locally so we don't need to worry about tls - https://book.kubebuilder.io/cronjob-tutorial/running.html

>If you want to run the webhooks locally, you’ll have to generate certificates for serving the webhooks, and place them in the right directory (/tmp/k8s-webhook-server/serving-certs/tls.{crt,key}, by default).

No.

## Acceptance Steps
`make run` after preliminary setup

@gnovv @acosta11

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,7 +61,7 @@ build: generate fmt vet ## Build manager binary.
 	go build -o bin/manager main.go
 
 run: manifests generate fmt vet ## Run a controller from your host.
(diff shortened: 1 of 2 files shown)


gnovv (approved):
lgtm!

davewalter (approved):
lgtm

Issue #89, opened 2021-10-12
[Feature]: API Client can see that the deletion of a CFRoute results in the deletion of the Contour HTTPProxy resources and the deletion of any Services


davewalter:
### Background

As an API Client
I want all created HTTPProxy resources to be cleaned up when I delete a CFRoute that does not have any destinations
So that I do not orphan any resources

As an API Client
I want all created HTTPProxy and Service resources to be cleaned up when I delete a CFRoute that does have at least one destination
So that I do not orphan any resources

### Acceptance Criteria

GIVEN that I am targeting a K8s cluster
AND that I have followed the README instructions to install the Controllers
WHEN I apply the `cfdomain.yaml` and `cfroute.yaml` resources in config/samples directory
THEN I should see all those operations succeed
AND I should see that `kubectl get httpproxy my-host.cf-apps.io  -o yaml` includes an owner reference to the CFRoute
AND I should see that `kubectl get httpproxy 84afc0f2-3dc2-4709-9fc8-8340be584a43  -o yaml` includes an owner reference to the CFRoute
AND I should see that `kubectl get service s-745cfe64-ac7a-4d4a-a627-c481e8f03be8-web -o yaml` includes an owner reference to the CFRoute
WHEN I delete the CFRoute resource with `kubectl delete cfroute 84afc0f2-3dc2-4709-9fc8-8340be584a43`
THEN I should see that `kubectl describe httpproxy my-host.cf-apps.io` does not return a valid resource
AND I should see that `kubectl describe httpproxy 84afc0f2-3dc2-4709-9fc8-8340be584a43 ` does not return a valid resource
AND I should see that `kubectl describe service s-745cfe64-ac7a-4d4a-a627-c481e8f03be8-web` does not return a valid resource

### Dev Notes

* ~This will require the use of a finalizer on the CFRoute resource. See [the routecontroller implementation](https://github.com/cloudfoundry/cf-k8s-networking/blob/d211c748db00fe83ee6892a9a804c31ee6a1c78a/routecontroller/controllers/networking/route_controller.go#L72-L88) for inspiration.~
  * After discussion with the team, we decided to start by using owner references as that means that Kubernetes will delete child resources without us having to implement our own finalizer.
* This will probably want to modify the current Service creation logic to use a GUID for the resource name instead of the current combination of app GUID and process type so that we can blindly delete the Service instead of figuring out whether it is still in use by another route.
  * This comes at the cost of a non-deterministic name for the Service.
[...]

PR #106, opened 2021-10-18
Route delete

davewalter:
#89 

## What is this change about?
This PR adds logic to the CFRoute reconcile loop to handle deleting Contour HTTPProxies and Services resources created by that route.

No

## Acceptance Steps
See #89 

@acosta11 @gnovv


gnovv (approved):
lgtm, except for the build test issue

Issue #90, opened 2021-10-12
[Feature]: API Client can see that the deletion of a destination from a CFRoute results in the deletion of the corresponding Service for that destination and the removal of the route entry from the sub-HTTPProxy


davewalter:
### Background

As an API Client
I want the Service related to a destination to be deleted when the corresponding destination is removed from a CFRoute
And I want the reference to the Service to be removed from the corresponding sub-HTTPProxy
So that I do not orphan resources

### Acceptance Criteria

GIVEN that I am targeting a K8s cluster
AND that I have followed the README instructions to install the Controllers
WHEN I apply the `cfdomain.yaml` and `cfroute.yaml` resources in `config/samples` directory
THEN I should see all those operations succeed
WHEN I edit the `cfroute.yaml` file to remove the destination and apply it to my cluster
THEN I should see that `kubectl describe service s-745cfe64-ac7a-4d4a-a627-c481e8f03be8-web` does not return a valid resource
AND I should see that `kubectl describe httpproxy 84afc0f2-3dc2-4709-9fc8-8340be584a43` returns a resource with an empty or missing `spec.routes` section

### Dev Notes

* We might need to update the HTTPProxy to remove the reference to the Service before deleting the Service to avoid an invalid status on the HTTPProxy.

PR #117, opened 2021-10-21
API Client can see that the deletion of a destination from a CFRoute results in the deletion of the corresponding Service for that destination and the removal of the route entry from the sub-HTTPProxy

matt-royal:
#90 

## What is this change about?
When a Destination is removed from a CFRoute, delete the Service associated with that Destination.

No

## Acceptance Steps
Follow the AC in #90 

@akrishna90 @gnovv and @davewalter have all worked on this


akrishna90 (approved):
LGTM!

PR #132, opened 2021-10-29
Merge api repo

Birdrock:
## Is there a related GitHub Issue?
No.

## What is this change about?
Subtree merging [cf-k8s-api](https://github.com/cloudfoundry/cf-k8s-api) into `cf-k8s-controllers`.

## Does this PR introduce a breaking change?
:laughing: :rofl: :sob: 

## Acceptance Steps
This will require some rework.

## Tag your pair, your PM, and/or team
@davewalter @acosta11 @gnovv @akrishna90 @matt-royal @gcapizzi @kieron-dev @danail-branekov @georgethebeatle

PR #140, opened 2021-11-01
Add a temporary top-level README

davewalter:
#132 

## What is this change about?
This PR creates a temporary top-level README and moves the LICENSE and NOTICE files up to the top-level of the repo.

No

## Acceptance Steps
N/A

@acosta11


acosta11 (approved):
LGTM

If anything, we should see if there's a quick fix to the ignore patterns to avoid running tests because the NOTICE and LICENSE files moved.

PR #142, opened 2021-11-02
Add a POST endpoint to add destinations to a route

akrishna90:
https://github.com/cloudfoundry/cf-k8s-api/issues/84

## What is this change about?
Adds a POST endpoint to add Destinations to Route.

No

## Acceptance Steps
Follow acceptance steps on the issue.

@matt-royal

### DIFF
--- a/api/apis/route_handler.go
+++ b/api/apis/route_handler.go
@@ -23,6 +23,7 @@ const (
 	RouteGetListEndpoint         = "/v3/routes"
 	RouteGetDestinationsEndpoint = "/v3/routes/{guid}/destinations"
 	RouteCreateEndpoint          = "/v3/routes"
+	RouteAddDestinationsEndpoint = "/v3/routes/{guid}/destinations"
 )
 
 //counterfeiter:generate -o fake -fake-name CFRouteRepository . CFRouteRepository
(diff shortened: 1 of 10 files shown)


api/apis/route_handler.go
```diff
@@ -290,9 +229,143 @@ func (h *RouteHandler) routeCreateHandler(w http.ResponseWriter, r *http.Request
...
+			}
+			domainGUIDToDomainRecord[currentDomainGUID] = domainRecord
+		}
+		routeRecords[i] = routeRecord.UpdateDomainRef(domainRecord)
+	}
+
+	return routeRecords, nil
+}
+
+func getDomainsForRoutes(ctx context.Context, domainRepo CFDomainRepository, client client.Client, routeRecords []repositories.RouteRecord) ([]repositories.RouteRecord, error) {
```
davewalter, line 353: Is this function still used anywhere?

davewalter (approved):
Looks good overall. I just had the one question about whether a private function in the handler is still being used.

api/apis/route_handler.go
akrishna90: Yes, it is used here https://github.com/cloudfoundry/cf-k8s-controllers/blob/5b0dc1a2a9cbf2be31247eb03264de3f9be60eac/api/apis/app_handler.go#L572

It might be a bit confusing as this function is not related to the PR but shows up as a change because we did a bit of formatting and moved all helper functions in this file to the very bottom.

Issue #165, opened 2021-10-11
[Feature]: App Developer can Apply a Minimal Space Manifest to Create New Apps


tcdowney:
### Background

**As an** App Developer
**I want** to be able to provide configuration for my App in the form of a (semi) declarative manifest
**So that** I can configure my App without scripting imperative CLI commands or orchestrating a bunch of API endpoint calls

Users can provide their own manifest along with their App Source code (or in a different directory via the `--manifest` and `-f` flags) and modern CF CLIs will interpret `cf push` flags and apply those on top of it. Even if the user doesn't provide a manifest, the CLI will construct one and use it for the bulk of the app creation/configuration flow of `cf push`.

There are many attributes that can be supported in an App Manifest:
* https://v3-apidocs.cloudfoundry.org/version/3.108.0/index.html#the-manifest-schema
* https://docs.cloudfoundry.org/devguide/deploy-apps/manifest-attributes.html

For the sake of this story, we aim to support the bare minimum that a manifest might contain, an application name.
```yaml
---
version: 1
applications:
  - name: app1
```

---

### Happy Path (App does not exist)

**GIVEN** I have a manifest that specifies an App that does not currently exist
**WHEN I** send it to the CF API Shim via the following request
```
curl "https://api-shim.example.org/v3/spaces/<space-guid>/actions/apply_manifest" \
  -X POST \
  -H "Authorization: bearer [token]" \
  -H "Content-type: application/x-yaml" \
  --data-binary @/path/to/manifest.yml
```

where the file `@/path/to/manifest.yml` looks like
```yaml
---
version: 1
applications:
  - name: app1
```

**THEN** I see a `CFApp` named `app1` created in the appropriate namespace in the `STOPPED` state.

```
HTTP/1.1 202 Accepted
Location: https://api-shim.example.org/v3/jobs/sync-space.apply_manifest-<space-guid>
```

---

### Happy Path (App DOES exist)

**GIVEN** I have a manifest that specifies an App that does currently exist
**WHEN I** send it to the CF API Shim via the following request
```
curl "https://api-shim.example.org/v3/spaces/<space-guid>/actions/apply_manifest" \
  -X POST \
  -H "Authorization: bearer [token]" \
  -H "Content-type: application/x-yaml" \
  --data-binary @/path/to/manifest.yml
```

where the file `@/path/to/manifest.yml` looks like
```yaml
---
version: 1
applications:
  - name: app1
```

[...]

PR #201, opened 2021-11-03
Add POST /v3/spaces/:guid/actions/apply_manifest endpoint

matt-royal:
#165 

## What is this change about?
Add an initial implementation of the apply manifest endpoint. This version only deals with creating apps that are enumerated, and ignores all fields but name. Only 1 app is allowed, and requests with more than 1 will fail.

No

## Acceptance Steps
Follow the AC in the story

@akrishna90 @davewalter

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,5 @@
+cover.out
+bin
+testbin
+.idea/
+~/.*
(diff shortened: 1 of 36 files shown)


gnovv (approved):
lgtm, verified on Kind cluster

Issue #210, opened 2021-11-04
[Feature]: API Client can hit a minimal Space Manifest Diff endpoint (`POST v3/spaces/<space-guid>/manifest_diff`)


tcdowney:
### Background

Newer versions of the CF CLI (v8+) hit the `POST v3/spaces/<space-guid>/manifest_diff` endpoint when a [space manifest](https://v3-apidocs.cloudfoundry.org/version/3.109.0/index.html#the-manifest-schema) is included with an app as part of `cf push`. We need to support a minimal version of this endpoint for the CLI to be able to continue with the push workflow.

**As a** client of the CF k8s API Shim
**I want** a minimal version of the `POST v3/spaces/<space-guid>/manifest_diff` endpoint to exist
**So that** I can use my existing client code to push apps

### Space Exists
**GIVEN** A CF Space exists and I have access to it
**WHEN I** I apply _any_ space manifest

```console
curl "https://api-shim.example.org/v3/spaces/<space-guid>/manifest_diff" \
  -X POST \
  -H "Content-Type: application/x-yaml" \
  -H "Authorization: bearer [token]" \
  --data-binary @/path/to/manifest.yml
```

**THEN I** get back an empty diff

```json
HTTP/1.1 202 OK
Content-Type: application/json

{
  "diff": []
}
```

---

### Space Not Accessible
**GIVEN** A CF Space does not exist / I don't have access to it
**WHEN I** I apply _any_ space manifest

```console
curl "https://api-shim.example.org/v3/spaces/<fake-space-guid>/manifest_diff" \
  -X POST \
  -H "Content-Type: application/x-yaml" \
  -H "Authorization: bearer [token]" \
  --data-binary @/path/to/manifest.yml
```

**THEN I** get back a 404

```json
HTTP/1.1 404 Not Found
Content-Type: application/json

{
  "errors": [
    {
      "detail": "Space not found",
      "title": "CF-ResourceNotFound",
      "code": 10010
    }
  ]
}
```

### Dev Notes

* Docs: https://v3-apidocs.cloudfoundry.org/version/3.109.0/index.html#create-a-manifest-diff-for-a-space-experimental
* Cloud Controller Code for how the diff works: https://github.com/cloudfoundry/cloud_controller_ng/blob/41036982a1d0ab82737c0f10f855bef650a7c5ef/app/actions/space_diff_manifest.rb

tcdowney:
[...]

PR #214, opened 2021-11-04
Add minimal space manifest diff endpoint

tcdowney:
* https://github.com/cloudfoundry/cf-k8s-controllers/issues/210

## What is this change about?
Newer CF CLIs hit this endpoint to be able to display a diff of the current state of their app with the manifest that they're applying. We first need to implement the ability to generate manifests before we can implement true diffing, so for now this endpoint will always return an empty diff. For more information check out the issue: https://github.com/cloudfoundry/cf-k8s-controllers/issues/210

Nope

## Acceptance Steps
Follow the A/C on the issue.

@matt-royal @gnovv

### DIFF
--- a/api/apis/space_manifest_handler.go
+++ b/api/apis/space_manifest_handler.go
@@ -17,12 +17,14 @@ import (
 
 const (
 	SpaceManifestApplyEndpoint = "/v3/spaces/{spaceGUID}/actions/apply_manifest"
+	SpaceManifestDiffEndpoint  = "/v3/spaces/{spaceGUID}/manifest_diff"
 )
 
 type SpaceManifestHandler struct {
 	logger              logr.Logger
 	serverURL           url.URL
 	applyManifestAction ApplyManifestAction
+	spaceRepo           CFSpaceRepository
 	buildClient         ClientBuilder
 	k8sConfig           *rest.Config // TODO: this would be global for all requests, not what we want
 }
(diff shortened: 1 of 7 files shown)


api/apis/space_manifest_handler.go
```diff
@@ -34,19 +36,22 @@ func NewSpaceManifestHandler(
...
 		applyManifestAction: applyManifestAction,
+		spaceRepo:           spaceRepo,
 		buildClient:         buildClient,
 		k8sConfig:           k8sConfig,
 	}
 }
 
 func (h *SpaceManifestHandler) RegisterRoutes(router *mux.Router) {
 	router.Path(SpaceManifestApplyEndpoint).Methods("POST").HandlerFunc(h.applyManifestHandler)
+	router.Path(SpaceManifestDiffEndpoint).Methods("POST").HandlerFunc(h.validateSpaceVisible(h.diffManifestHandler))
```
tcdowney, line 54: Eventually we'd like to apply this same check to the `applyManifestHandler`, which is why I made the wrapping handler, but it was expensive to update its integration tests due to the coupling between Spaces and HNS resources and HNC. It's out of scope for this story (which is a blocker for `cf push` to proceed) so I decided to skip it for now, but I feel like this is pointing at another issue with the tight coupling between orgs/spaces and HNC.

gnovv (approved):
I think you forgot to update the api doc!
Otherwise looks good to me

tcdowney:
Thanks @gnovv , I rebased and updated the docs. 👍

Issue #160, opened 2021-09-17
[Feature]: API client can assign a space role to a user with `POST /v3/roles`


gcapizzi:
### Background

**As an** admin
**I want** to assign space roles to users
**So that** I can control permissions in the space

### Acceptance Criteria

**GIVEN** a space exists
**AND** a user other than the admin exists
**AND** a `ClusterRole` exists
**WHEN** I make the following request:
```sh
curl "https://api.example.org/v3/roles" \
  -X POST \
  -H "Content-Type: application/json" \
  -d '{
      "type": "<the ClusterRole name>",
      "relationships": {
        "user": {
          "data": {
            "guid": "<the username>"
          }
        },
        "space": {
          "data": {
            "guid": "<the space GUID>"
          }
        }
      }
    }'
```
**THEN** I get the following response:
```
HTTP/1.1 201 Created
Content-Type: application/json
```

```json
{
   "guid": "40557c70-d1bd-4976-a2ab-a85f5e882418",
   "created_at": "2019-10-10T17:19:12Z",
   "updated_at": "2019-10-10T17:19:12Z",
   "type": "<the ClusterRole name>",
   "relationships": {
      "user": {
         "data": {
            "guid": "<the username>"
         }
      },
      "space": {
         "data": {
            "guid": "<the space GUID>"
         }
      },
      "organization": {
         "data": null
      }
   },
   "links": {
      "self": {
         "href": "https://api.example.org/v3/roles/40557c70-d1bd-4976-a2ab-a85f5e882418"
      },
      "space": {
         "href": "https://api.example.org/v3/spaces/05c5da3b-6cbc-421c-87c3-20bb3c41ab7c"
      }
   }
}
``` 
**AND** I see a `RoleBinding` in the space namespace for the specified user

### Dev Notes

* [`POST /v3/roles`](https://v3-apidocs.cloudfoundry.org/version/3.106.0/index.html#create-a-role)

PR #215, opened 2021-11-05
Introduce `POST /v3/roles` endpoint to create space roles

danail-branekov:
https://github.com/cloudfoundry/cf-k8s-controllers/issues/160

## What is this change about?
Introduce the `POST /v3/roles` API endpoint. Whenever called, the API would create a rolebinding in a space namespace that binds a user to a role. Note that the API would fail to create a space role if the user does not have a role (i.e. a role binding) in the parent org namespace (this is how the API should behave according to the CF API documentation)

Yes

K8s role names differ from the CF roles, for example, `space_developer` maps to `cf-k8s-controllers-space-developer`. In order to express this mapping we have introduced a new [configuration yaml](https://github.com/cloudfoundry/cf-k8s-controllers/blob/wip-create-role/api/config/base/config/role_mappings_config.yaml) providing that mapping. We did not want to put that config in `cf_k8s_api_config.yaml` as that would mean that we would have to carry those mappings across all kustomization overlays. 
That is why we now have two config files and we changed the meaning of the `CONFIG` environment variable to reference a directory, rather than a single file. The `main` would read the content of the directory and gradually fill the configuration object.
We have amended the `.envrc` accordingly so that `make run` works.

## Acceptance Steps
* Create org
`curl -X POST /v3/organizations
   -H 'Content-Type: application/json'
   -d '{"name":"my-org"}'`
* Create a rolebinding in the org namespaces that binds a user to whatever role
`kubectl create rolebinding <rolebinding-name> -n <org-ns-name> --clusterrole=some-role --user=some-user`
* Create space 
`curl -X POST /v3/spaces
   -H 'Content-Type: application/json'
   -d '{
"name": "space-name",
"relationships": {
  "organization": {
    "data": {
      "guid": "<org-guid>"
    }
  }
}
}'`

* Create the role:
```
curl "https://api.example.org/v3/roles" \
  -X POST \
  -H "Content-Type: application/json" \
  -d '{
      "type": "cf-k8s-controllers-space-developer",
      "relationships": {
        "user": {
          "data": {
            "guid": "some-user>"
          }
        },
        "space": {
          "data": {
            "guid": "<the space GUID>"
          }
        }
      }
    }'
``` 

*See success response like:

```
{
   "guid": "40557c70-d1bd-4976-a2ab-a85f5e882418",
   "created_at": "2019-10-10T17:19:12Z",
   "updated_at": "2019-10-10T17:19:12Z",
   "type": "<the ClusterRole name>",
   "relationships": {
      "user": {
         "data": {
            "guid": "<the username>"
         }
      },
      "space": {
         "data": {
            "guid": "<the space GUID>"
         }
      },
      "organization": {
         "data": null
      }
   },
   "links": {
      "self": {
         "href": "https://api.example.org/v3/roles/40557c70-d1bd-4976-a2ab-a85f5e882418"
      },
      "space": {
         "href": "https://api.example.org/v3/spaces/05c5da3b-6cbc-421c-87c3-20bb3c41ab7c"
      }
   }
}
```

@cloudfoundry/cf-k8s


api/repositories/role_repository.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+
+	if !hasOrgBinding {
+		return RoleRecord{}, ErrorMissingRoleBindingInParentOrg
+	}
+
+	roleBinding := rbacv1.RoleBinding{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: role.Space,
+			Name:      calculateRoleBindingName(role),
```
kieron-dev, line 90: It makes sense encoding subject and role name in the rolebinding name here. Although breaking the convention of using a guid in the name, it makes sense to have a deterministic name here since this name is immutable in CF. Hence we protect from duplicate rolebindings using built-in k8s name unqiueness checking.

api/repositories/role_repository_test.go
```diff
@@ -0,0 +1,218 @@
...
+		ctx = context.Background()
+		authorizedInChecker = new(fake.AuthorizedInChecker)
+		Expect(k8sClient.Create(context.Background(), &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: rootNamespace}})).To(Succeed())
+		roleRepo = repositories.NewRoleRepo(k8sClient, authorizedInChecker, map[string]string{"space_developer": "cf-space-dev-role"})
+
+		roleRecord = repositories.RoleRecord{}
+	})
+
+	Describe("CreateSpaceRole", func() {
+		createSubnamespaceAnchor := func(name, parent string) *hnsv1alpha2.SubnamespaceAnchor {
```
kieron-dev, line 41: Given we set SpaceNameLabel, this is creating a `space` subnamespaceanchor. Consider renaming.

api/repositories/role_repository_test.go
```diff
@@ -0,0 +1,218 @@
...
+		var (
+			orgAnchor   *hnsv1alpha2.SubnamespaceAnchor
+			spaceAnchor *hnsv1alpha2.SubnamespaceAnchor
+			createdRole repositories.RoleRecord
+			createErr   error
+		)
+
+		BeforeEach(func() {
+			authorizedInChecker.AuthorizedInReturns(true, nil)
+			orgAnchor = createSubnamespaceAnchor(uuid.NewString(), rootNamespace)
```
kieron-dev, line 76: We are creating an org anchor, but setting the `SpaceNameLabel` on it. Suggest modifying this to remove future confusion

api/repositories/role_repository_test.go
kieron-dev: There is also some duplication with org/space creation in the org_repository_test code. Potential refactor.

api/tests/e2e/e2e_suite_test.go
```diff
@@ -59,6 +67,16 @@ var _ = BeforeSuite(func() {
 	ensureServerIsUp()
 })
 
+var _ = BeforeEach(func() {
+	serviceAccountName = generateGUID("user")
+	token := obtainServiceAccountToken(serviceAccountName)
```
kieron-dev, line 72: `obtainServiceAccountToken` does a fair bit of work, so we're introducing a significant overhead into each test here.

Consider whether this is necessary for every e2e test. If so, it might make more sense to be shared in a synchronised before suite.

api/tests/e2e/roles_test.go
```diff
@@ -0,0 +1,134 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ = Describe("Roles", func() {
+	var (
+		ctx      context.Context
+		userName string
+	)
+
+	createRoleWithHeaders := func(roleName, userName, spaceGUID string, headers map[string]string) (*http.Response, error) {
```
kieron-dev, line 26: Is this a bit premature? We're currently using the privileged client to create the role binding, so the auth header is not used.

api/tests/e2e/roles_test.go
```diff
@@ -0,0 +1,134 @@
...
+			createBinding(org.GUID, userName, "basic-user")
+		})
+
+		AfterEach(func() {
+			deleteSubnamespace(org.GUID, space.GUID)
+			deleteSubnamespace(rootNamespace, org.GUID)
+		})
+
+		It("creates a role binding", func() {
+			response, err := createRoleWithHeaders("space_developer", userName, space.GUID, nil)
```
kieron-dev, line 98: `nil`. See comment above.

kieron-dev (changes requested):
See inline comments.

Also, where are the changes to the .envrc mentioned in one of the commits?

kieron-dev (approved):

PR #217, opened 2021-11-05
fixed port bug on LRP

gnovv:
Adds some test refactors and fixed a bug with the LRP Health Check when the CFProcess has no ports

### DIFF
--- a/controllers/controllers/workloads/cfprocess_controller.go
+++ b/controllers/controllers/workloads/cfprocess_controller.go
@@ -103,6 +103,10 @@ func (r *CFProcessReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 			}
 		}
 
+		var lrpHealthCheckPort int32 = 0
+		if len(cfProcess.Spec.Ports) > 0 {
+			lrpHealthCheckPort = cfProcess.Spec.Ports[0]
+		}
 		_, err = controllerutil.CreateOrPatch(ctx, r.Client, &lrp, func() error {
 			lrp.Spec.GUID = cfProcess.Name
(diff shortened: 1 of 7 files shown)


controllers/controllers/workloads/cfprocess_controller.go
```diff
@@ -103,6 +103,10 @@ func (r *CFProcessReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 			}
 		}
 
+		var lrpHealthCheckPort int32 = 0
```
Birdrock, line 106: Isn't this redundant? The zero value for int32 is already 0 if I recall correctly.

controllers/controllers/workloads/cfprocess_controller.go
```diff
@@ -103,6 +103,10 @@ func (r *CFProcessReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
+		if len(cfProcess.Spec.Ports) > 0 {
+			lrpHealthCheckPort = cfProcess.Spec.Ports[0]
```
Birdrock, line 108: Out of scope for this, but I still feel it's pretty strange to just pick the first element as the health check. We don't even know what type of traffic whatever is listening on that port expects.

Birdrock (approved):
Isn't this redundant? The zero value of int32 is 0, iirc.

controllers/controllers/workloads/cfprocess_controller.go
gnovv: I guess I prefer to be explicit for now, we can change it later if people have concerns

Issue #101, opened 2021-10-15
[Feature]: API Client can see the FQDN and URI of a CFRoute in status fields


davewalter:
### Background

**As an** API Client
**I want** to see the FQDN for a `CFRoute` in its `status.fqdn` field
**And I want** to see the URI for a `CFRoute` in its `status.uri` field
**So that** I do not have to retrieve the corresponding `CFDomain` resource to determine these values

### Acceptance Criteria

**GIVEN** that I am targeting a K8s cluster
**AND** that I have followed the README instructions to install the Controllers
**WHEN** I apply the `cfdomain.yaml` and `cfroute.yaml` resources in config/samples directory
**THEN** I should see all those operations succeed
**AND** I should see that `kubectl get cfroute 84afc0f2-3dc2-4709-9fc8-8340be584a43 -o yaml` includes the FQDN for the route (`my-host.cf-apps.io`) in the `status.fqdn` field
**AND** I should see that `kubectl get cfroute 84afc0f2-3dc2-4709-9fc8-8340be584a43 -o yaml` includes the URI for the route (`my-host.cf-apps.io`) in the `status.uri` field

**GIVEN** that I am targeting a K8s cluster
**AND** that I have followed the README instructions to install the Controllers
**AND** that I have copied the `cfroute.yaml` resource from the config/samples directory to a temporary file and modified it to specify the path `/example`
**WHEN** I apply the `cfdomain.yaml` resource in config/samples directory and the `cfroute.yaml` resource from the temporary directory
**THEN** I should see all those operations succeed
**AND** I should see that `kubectl get cfroute 84afc0f2-3dc2-4709-9fc8-8340be584a43 -o yaml` includes the FQDN for the route (`my-host.cf-apps.io`) in the `status.fqdn` field
**AND** I should see that `kubectl get cfroute 84afc0f2-3dc2-4709-9fc8-8340be584a43 -o yaml` includes the URI for the route (`my-host.cf-apps.io/example`) in the `status.uri` field

### Dev Notes
* This should be done by the CFRouteController
  * The CFRouteController may need to be updated to watch for CFDomain changes (example from our spike: https://github.com/cloudfoundry/cf-crd-explorations/blob/8cf10d5aafec5695394c1f4eebf21ce481ca800b/controllers/process_controller.go#L227-L241)
  * BuildController / kpack code will have some real code examples

PR #219, opened 2021-11-06
Set the FQDN and URI status fields on a CFRoute

davewalter:
#101 

## What is this change about?
This PR updates the CFRoute reconcile loop to set the FQDN and URI status fields

No

## Acceptance Steps
See #101


gnovv (approved):
ltgm!

Issue #164, opened 2021-10-11
[Feature]: API Client can retrieve Jobs for space manifest application via `GET /v3/jobs/:guid` (minimal version)


tcdowney:
### Blockers/Dependencies

* If we encode meaning into the `Job` guid we need to make sure that https://github.com/cloudfoundry/cf-k8s-api/issues/144 returns the correct guids.

### Background

In CF for VMs the Apply Space Manifest endpoint is asynchronous. When the CLI calls it it gets back a `Job` guid in the response for it to track.

It then polls the `GET /v3/jobs/:guid` endpoint and will get back responses like this:

```
{
  "created_at": "2021-10-11T18:23:24Z",
  "errors": null,
  "guid": "7689eeca-bfc2-410e-b91e-d8682ce83743",
  "links": {
    "self": {
      "href": "https://api.boulder-finger.capi.land/v3/jobs/7689eeca-bfc2-410e-b91e-d8682ce83743"
    },
    "space": {
      "href": "https://api.boulder-finger.capi.land/v3/spaces/8269b7d4-364b-4dcb-960f-38567d854736"
    }
  },
  "operation": "space.apply_manifest",
  "state": "PROCESSING",
  "updated_at": "2021-10-11T18:23:24Z",
  "warnings": null
}
```

Until it is `COMPLETE`:

```
{
  "created_at": "2021-10-11T18:23:24Z",
  "errors": null,
  "guid": "7689eeca-bfc2-410e-b91e-d8682ce83743",
  "links": {
    "self": {
      "href": "https://api.boulder-finger.capi.land/v3/jobs/7689eeca-bfc2-410e-b91e-d8682ce83743"
    },
    "space": {
      "href": "https://api.boulder-finger.capi.land/v3/spaces/8269b7d4-364b-4dcb-960f-38567d854736"
    }
  },
  "operation": "space.apply_manifest",
  "state": "COMPLETE",
  "updated_at": "2021-10-11T18:23:28Z",
  "warnings": null
}
```

Since our manifest applications are synchronous and we do not have any other use for the `Job` resource at this time, this is a place where we can cut scope and return static value.

---

**As a** client of the CF API Shim
**I want** the `GET /v3/jobs/:guid` endpoint to return successfully
**So that** I can reuse my existing client codebase

### Acceptance Criteria

**GIVEN** I have the API shim deployed
**WHEN I** make the following request

```
[...]

PR #224, opened 2021-11-09
API Client can retrieve Jobs for space manifest application

akrishna90:
#164 

## What is this change about?
Adds the job endpoint which always returns a static response with `state` set to `COMPLETE`

No.

## Acceptance Steps
Follow acceptance steps on #164 

@Birdrock

### DIFF
--- a/api/apis/job_handler.go
+++ b/api/apis/job_handler.go
@@ -1,0 +1,66 @@
+package apis
+
+import (
+	"encoding/json"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"code.cloudfoundry.org/cf-k8s-controllers/api/presenter"
+
+	"github.com/go-logr/logr"
+	"github.com/gorilla/mux"
+)
+
+const (
+	JobGetEndpoint  = "/v3/jobs/{guid}"
+	syncSpacePrefix = "sync-space.apply_manifest-"
+)
+
+type JobHandler struct {
+	logger    logr.Logger
+	serverURL url.URL
+}
+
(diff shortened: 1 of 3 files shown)


api/apis/job_handler.go
```diff
@@ -0,0 +1,66 @@
...
+
+	"code.cloudfoundry.org/cf-k8s-controllers/api/presenter"
+
+	"github.com/go-logr/logr"
+	"github.com/gorilla/mux"
+)
+
+const (
+	JobGetEndpoint  = "/v3/jobs/{guid}"
+	syncSpacePrefix = "sync-space.apply_manifest-"
```
tcdowney, line 17: I know we chatted about this a bit on Slack, but wanted to put some context here since I know this looks weird:

`space.apply_manifest` is the operation that these jobs get assigned by Cloud Controller:
https://github.com/cloudfoundry/cloud_controller_ng/blob/4a23a419862ef9b7272f411a4435e33fc7305f30/app/jobs/space_apply_manifest_action_job.rb#L56

I proposed appending `sync-` to it because this is a synchronous manifest application and used hyphens because those align with the general guid structure.

Open to other suggestions or thoughts on what a good sentinel value here could be as well.

tcdowney (approved):
LGTM 🙂

api/apis/job_handler.go
Birdrock: With that context, I don't think there's much we can do. We could use an underscore, but the guid would still have hyphens.

akrishna90: Looks like we would always end up with a mixture of hypens, underscores and dashes. I'll leave the implementation as is.

PR #225, opened 2021-11-09
Manifest can Update Environment Variables for an App

matt-royal:
#156 

## What is this change about?
The environment variables provided in a manifest will now update an application's environment variables. This is done as an additive patch, and any environment variables on the app that weren't specified in the app will retain their existing value

No. There is a behavioral change though. Previously, only Apps with environment variables would have Secrets created for them. Now all Apps have a Secret. Also, the Secret now has an OwnerRef back to the App so it will be garbage collected now.

## Acceptance Steps
See #156 

@akrishna90


matt-royal:
Had to rebase the branch against main... waiting for tests to pass here before merging

api/repositories/app_repository_test.go
```diff
@@ -330,163 +334,180 @@ var _ = Describe("AppRepository", func() {
...
-				emptyAppRecord = AppRecord{}
-			)
+			Expect(k8sClient.Create(testCtx, &corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{Name: spaceGUID},
+			})).To(Succeed())
+
+			appCreateMessage = initializeAppCreateMessage(testAppName, spaceGUID)
+		})
+
+		AfterEach(func() {
```
tcdowney, line 354: It looks like there is a test failing due to some test pollution related to this repo. I'm wondering if we're missing a delete `"test-app-name"` call here in this `AfterEach`?

tcdowney (changes requested):
LGTM apart from it looks like there is a potential test pollution issue.

api/repositories/droplet_repository_test.go
```diff
@@ -140,6 +140,7 @@ var _ = Describe("DropletRepository", func() {
 						Expect(createdAt).To(BeTemporally("~", time.Now(), timeCheckThreshold*time.Second))
 					})
 
+					// TODO: Fix these failing UpdatedAt assertions or remove these tests
```
matt-royal, line 143: I eventually figured out that this happens when I run the envtest tests without specifying the `KUBEBUILDER_ASSETS` env var. I'll remove the comment.

api/repositories/app_repository_test.go
matt-royal: My theory has been that we should be able to just delete the namespace that contains the resources, and it will cascade the delete to the resources in the namespace. However, this doesn't seem to work consistently in test, which is confusing.

matt-royal:
This PR is superceded by https://github.com/cloudfoundry/cf-k8s-controllers/pull/249 so I'm closing it

Issue #229, opened 2021-11-10
[Chore]: Update ServiceAccounts and Secrets for `cf namespace`


Birdrock:
### Background

Currently, Kpack expects to operate in the `default` namespace, while Eirini expects to operate in the `cf` namespace. This means that one cannot get resources reconciled end to end.

### Action to take

Update Kpack configuration and resources to operate in the `cf` namespace.

### Impact

This is a blocker for further development on Process stories.

georgethebeatle:
This one looks obsolete, shall we close it? @Birdrock

Birdrock:
@georgethebeatle Yep.

PR #230, opened 2021-11-10
Align Kpack and Eirini workloads namespace

Birdrock:
## What is this change about?
#229 

Existing installations will need to update the `kpack-service-account` ServiceAccount and `image-registry-credentials` Secret.

## Acceptance Steps
Run all tests.


tcdowney (approved):
lgtm

Issue #150, opened 2021-09-28
[Feature]: Developer should only see spaces they belong to


gcapizzi:
### Blockers

* cloudfoundry/cf-k8s-api#98 
* cloudfoundry/cf-k8s-controllers#168

### Background

At the moment, `GET /v3/spaces` returns the full list of spaces present on the foundation. We should only return the spaces the user making the request belongs to, e.g. spaces the user has some role in.

We want to keep this behaviour behind a **feature flag** for now, to avoid having a mix of authenticated and unauthenticated endpoints exposed.

### Acceptance Criteria

**GIVEN** a `cf` root namespace exists
**AND** the `cf/org1`, `cf/org2` and `cf/org3` subnamespaces exist
**AND** the `cf/org1/space1`, `cf/org2/space2` and `cf/org3/space3` subnamespaces exist
**AND** `user` has `RoleBinding`s (to any role for now) in `org1` and `org3`
**AND** `user` has `RoleBinding`s (to any role for now) in `space1` and `space3`
**WHEN I** `GET /v3/spaces` with a correct `Authorization` header
**THEN I** only see `space1` and `space3` in the `resources` field

### Dev Notes

*  Org Managers can list all spaces inside their org. While we don't need to implement the Org Manager role yet, let's take this into account while implementing this.
* It is impossible in CF to have a space role without having a role in the corresponding org.

PR #238, opened 2021-11-12
Add authorization to the fetch spaces endpoint

danail-branekov:
https://github.com/cloudfoundry/cf-k8s-controllers/issues/150

## What is this change about?
Add authorization to the `GET /v3/spaces` endpoint. User credentials are expected to be put into the `Authorization` request header.

Sort of, the create space (`POST /v3/spaces`) now implicitly requires authentication

## Acceptance Steps
See [the story](https://github.com/cloudfoundry/cf-k8s-controllers/issues/150) for acceptance

@mnitchev


api/apis/space_handler.go
```diff
@@ -87,8 +101,24 @@ func (h *SpaceHandler) SpaceListHandler(w http.ResponseWriter, r *http.Request)
 	orgUIDs := parseCommaSeparatedList(r.URL.Query().Get("organization_guids"))
 	names := parseCommaSeparatedList(r.URL.Query().Get("names"))
 
-	spaces, err := h.spaceRepo.FetchSpaces(ctx, orgUIDs, names)
+	spaceRepo, err := h.spaceRepoProvider.SpaceRepoForRequest(r)
+	if err != nil {
+		if authorization.IsUnauthorized(err) {
```
tcdowney, line 106: Question for my own understanding:

This just checks that the user submitted an auth header right? If they have correctly submitted an `Authorization` header it will get past this and if they lack the requisite RBAC then the call on line 119 to `FetchSpaces` will fail?

If that's the case, I'm wondering if instead of an unknown error we should check to see if the request failed due to authorization reasons and return a `403` instead in that case.

tcdowney (dismissed):

api/apis/apis_suite_test.go
```diff
@@ -114,6 +114,18 @@ func expectUnknownKeyError() {
     }`)
 }
 
+func expectUnauthorizedError() {
+	expectJSONResponse(http.StatusUnauthorized, `{
+        "errors": [
+            {
+                "detail": "No auth token was given, but authentication is required for this endpoint",
```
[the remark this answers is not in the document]
kieron-dev: We're changing this to "Authentication error" as suggested.

api/apis/space_handler.go
danail-branekov: > ... and if they lack the requisite RBAC then the call on line 119 to FetchSpaces will fail?
Not really. In order to list authorised spaces efficiently we lean on the privileged client to get them all in one shot and filter out the ones the authenticated user has been bound to. We cannot rely on RBAC as namespaces are cluster-wide resources, and therefore users (and the user-scoped client) are not expected to have permissions to list them. 

TL;DR, the call on line 119 cannot fail due to authorisation issues because that line would use the privileged client that can do anything.

tcdowney: Ah yeah, I see.

I guess that's more of an issue for the future endpoints then. Thanks for the clarification!

mnitchev:
We've resolved the propagation issue. We decided to make the propagation of roles configurable in the role mappings. This means that roles like the OrgAuditor, OrgUser and OrgBillingManager that don't have permissions to list spaces within an org will not be propagated down to the space namespace, hence fixing the listing problem.

If there is no more feedback that needs to be addressed, this PR can be merged

georgethebeatle (approved):

georgethebeatle:
@tcdowney can you please approve this PR so that we can merge. I can see that the conversation is resolved, so I am guessing you just forgot to click the button.

Issue #120, opened 2021-10-25
[Chore]: Flaky test: CFRouteReconciler Integration Tests when the CFRoute does not include any destinations [It] eventually reconciles the CFRoute to a root Contour HTTPProxy which includes a proxy for a route destination 


danail-branekov:
### What happened?

While running the github actions pipeline I got the following flaky error:

```
Running Suite: Networking Controllers Integration Suite
=======================================================
Random Seed: 1635176308
Will run 11 of 11 specs

2021-10-25T15:38:47.750Z	INFO	controller.cfdomain	Starting EventSource	{"reconciler group": "networking.cloudfoundry.org", "reconciler kind": "CFDomain", "source": "kind source: /, Kind="}
2021-10-25T15:38:47.750Z	INFO	controller.cfdomain	Starting Controller	{"reconciler group": "networking.cloudfoundry.org", "reconciler kind": "CFDomain"}
2021-10-25T15:38:47.750Z	INFO	controller.cfroute	Starting EventSource	{"reconciler group": "networking.cloudfoundry.org", "reconciler kind": "CFRoute", "source": "kind source: /, Kind="}
2021-10-25T15:38:47.750Z	INFO	controller.cfroute	Starting Controller	{"reconciler group": "networking.cloudfoundry.org", "reconciler kind": "CFRoute"}
2021-10-25T15:38:48.151Z	INFO	controller.cfroute	Starting workers	{"reconciler group": "networking.cloudfoundry.org", "reconciler kind": "CFRoute", "worker count": 1}
2021-10-25T15:38:48.151Z	INFO	controller.cfdomain	Starting workers	{"reconciler group": "networking.cloudfoundry.org", "reconciler kind": "CFDomain", "worker count": 1}
2021-10-25T15:38:48.177Z	ERROR	controllers.CFRoute	failed to get CFDomain	{"error": "CFDomain.networking.cloudfoundry.org \"a53525b1-35a9-11ec-a37a-000d3ae29adf\" not found"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/home/runner/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.10.2/pkg/internal/controller/controller.go:311
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
[...]

PR #239, opened 2021-11-12
Fix flaky cfroutes controller flaky tests

danail-branekov:
https://github.com/cloudfoundry/cf-k8s-controllers/issues/120

## What is this change about?
CFRoutes controller integration tests frequently failed because of their `Eventually` expectations time out. This change gets rid of the 2 second timeout in the test and sets the default eventually timeout in the suite to 10 seconds. Hopefully this would be enough

No

## Acceptance Steps
Run tests, they should succeed

### DIFF
--- a/controllers/controllers/networking/integration/suite_integration_test.go
+++ b/controllers/controllers/networking/integration/suite_integration_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"path/filepath"
 	"testing"
+	"time"
 
 	networkingv1alpha1 "code.cloudfoundry.org/cf-k8s-controllers/controllers/apis/networking/v1alpha1"
 	. "code.cloudfoundry.org/cf-k8s-controllers/controllers/controllers/networking"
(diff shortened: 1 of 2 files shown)


controllers/controllers/networking/integration/suite_integration_test.go
```diff
@@ -33,6 +34,7 @@ func TestNetworkingControllers(t *testing.T) {
 }
 
 var _ = BeforeSuite(func() {
+	SetDefaultEventuallyTimeout(10 * time.Second)
```
tcdowney, line 37: Do all of these assertions have the potential to take this long, or is it just a subset?

I guess some of them depend on changes made by the Services reconciler and Contour's HTTPProxy controller which is why it's so variable?

cc @acosta11 and @davewalter

danail-branekov: Well, all the eventuallies in the test had a two second timeout so I guess they take some time. From my point of view, we should not bother with how long the eventually timeout is as long as there are no performance requirements. I personally prefer a test to always pass within 10 seconds, rather than occasionally fail due to insufficient eventually timeout.

Issue #147, opened 2021-09-30
[Feature]: CLI user can run `cf create-space` successfully


gcapizzi:
### Blockers/Dependencies

* cloudfoundry/cf-k8s-api#61
* cloudfoundry/cf-k8s-controllers#166
* cloudfoundry/cf-k8s-controllers#160

### Background

**As a** developer
**I want** to create a space using the `cf` CLI
**And** automatically become a `SpaceManager` and `SpaceDeveloper` for the new space
**So that** I can manage apps in my space

### Acceptance Criteria

**GIVEN** a `cf-k8s-api` instance exists
**AND** an organization exists
**AND** I have targeted the org with `cf target -o <org-name>`
**WHEN I** run `cf create-space <space-name>`
**AND** `cf spaces`
**THEN I** see the created space in the list

**GIVEN** a `cf-k8s-api` instance exists
**AND** an organization exists
**WHEN I** run `cf create-space -o <org-name> <space-name>`
**AND** `cf target -o <org-name>`
**AND** `cf spaces`
**THEN I** see the created space in the list

### Dev Notes

* The CLI will try to assign the [`OrgUser`](https://github.com/eirini-forks/cli/blob/93284695ec2b2d4cdc44d3601986401c342de73c/actor/v7action/role.go#L47), [`SpaceManager`](https://github.com/cloudfoundry/cli/blob/5ffa262bb2e51ba770d456fd169da89ca0b9801f/command/v7/create_space_command.go#L104) and [`SpaceDeveloper`](https://github.com/cloudfoundry/cli/blob/5ffa262bb2e51ba770d456fd169da89ca0b9801f/command/v7/create_space_command.go#L117) roles to the user creating the space. This will in turn need a valid username returned by the `CurrentUser()` method in `Config`. This in turn will need a `/whoami` enpoint. We have discussed these things [here](https://github.com/cloudfoundry/cf-crd-explorations/issues/72#issuecomment-905614411) and [here](https://github.com/cloudfoundry/cf-crd-explorations/issues/71#issuecomment-905534136).

kieron-dev:
See commit above https://github.com/eirini-forks/cli/commit/190558df25c702fee15619542c18d3b77df51f13. This has passing tests.

[...]

PR #241, opened 2021-11-15
Support cli create space

mnitchev:
#147 

## What is this change about?
In order for the cli to be able to get the current user, the api needs to support a `/whoami` endpoint. Since the user can change who they are logged in as for the cluster at any time (for example with `gcloud`), the cli needs the endpoint to determine what the user is from the token or cert.

Additionally we spotted that the create role endpoint will be passed a `username` instead of a `guid`. In practice we'll probably always get a `username`, but in this PR we support both. This behavior is controlled by the [set_roles_by_username](https://v3-apidocs.cloudfoundry.org/version/3.106.0/index.html#list-of-feature-flags) feature flag, which is on by default.

No

## Acceptance Steps
1. `cf api <the-api-endpoint>`
2. `cf login`
3. notice that the user is set to the actual user you're logged in to in Kubernetes
4. run any other command like `cf spaces` for example
5. notice the cli says the real user here as well

@kieron-dev


tcdowney (changes requested):
Overall sounds good to me. I'm worried that the context for why this `/whoami` endpoint exists might get lost over time, though.

Can y'all add an entry for it to the rough API docs file we've got? I think what you've put in the PR description makes sense.

https://github.com/cloudfoundry/cf-k8s-controllers/blob/main/api/docs/api.md

kieron-dev:
> Can y'all add an entry for it to the rough API docs file we've got? 

Hi @tcdowney. Thanks for the review. I've added a section to that doc.

tcdowney (approved):
LGTM

PR #242, opened 2021-11-16
Remove unused config.go

davewalter:
No

## What is this change about?
This PR removes the unused duplicate config.go from the `api` directory.

No

## Acceptance Steps
Everything still works.


tcdowney (approved):
LGTM

I assume this is the one we actually use?
https://github.com/cloudfoundry/cf-k8s-controllers/blob/main/api/config/config.go

davewalter:
> I assume this is the one we actually use? https://github.com/cloudfoundry/cf-k8s-controllers/blob/main/api/config/config.go

Yes. When I remove it I get a Go compiler error:
```
go vet ./...
api/payloads/app.go:4:2: no required module provides package code.cloudfoundry.org/cf-k8s-controllers/api/config; to add it:
	go get code.cloudfoundry.org/cf-k8s-controllers/api/config
make: *** [vet] Error 1
```

Issue #177, opened 2021-09-28
[Feature]: API Client can list domains via GET /v3/domains (with filtering by name)


davewalter:
### Background

As a client of the API shim
I want to be able to list Domains
So that I can see all the domains that I have access to

As a client of the API shim
I want to be able to filter the list Domains by domain name
So that I can see a subset of the domains that I have access to

## Acceptance Scenarios

**GIVEN** I have the API shim running locally
**THEN** I can make perform the following scenarios

---

### Happy Path (when at least one domain exists)

**GIVEN** I have applied a CFDomain CR to the cluster (see cf-k8s-controllers samples)
**WHEN** I make the following request
```
curl "https://api.example.org/v3/domains" \
  -X GET \
  -H "Authorization: bearer <some-placeholder-token>" \
```

**THEN** I get a response that looks like the following

```json
HTTP/1.1 200 OK
Content-Type: application/json

{
  "pagination": {
    "total_results": 1,
    "total_pages": 1,
    "first": {
      "href": "https://api.example.org/v3/domains?page=1"
    },
    "last": {
      "href": "https://api.example.org/v3/domains?page=1"
    },
    "next": null,
    "previous": null
  },
  "resources": [
    {
      "guid": "5b5032ab-7fc8-4da5-b853-821fd1879201",
      "created_at": "2019-03-08T01:06:19Z",
      "updated_at": "2019-03-08T01:06:19Z",
      "name": "cf-apps.io",
      "internal": false,
      "router_group": null,
      "supported_protocols": ["http"],
      "metadata": {
        "labels": {},
        "annotations": {}
      },
      "relationships": {
        "organization": {
          "data": null
        },
        "shared_organizations": {
          "data": []
        }
      },
      "links": {
        "self": {
          "href": "https://api.example.org/v3/domains/5b5032ab-7fc8-4da5-b853-821fd1879201"
        },
        "route_reservations": {
          "href": "https://api.example.org/v3/domains/5b5032ab-7fc8-4da5-b853-821fd1879201/route_reservations"
        },
        "router_group": null
      }
    },
  ]
}
```

---

### Happy Path (when no domains exist)

**GIVEN** that no CFDomain resources exist
**WHEN** I make the following request
```
curl "https://api.example.org/v3/domains" \
  -X GET \
  -H "Authorization: bearer <some-placeholder-token>" \
```

**THEN** I get a response that looks like the following

```json
HTTP/1.1 200 OK
[...]

PR #245, opened 2021-11-16
Fix GET /v3/domains names param parsing

tcdowney:
- https://github.com/cloudfoundry/cf-k8s-controllers/issues/177

## What is this change about?
Hitting `GET /v3/domains` without a `names` filter would always return an empty list of domains instead of all visible domains. This change adds some tests and improves the param parsing to correctly interpret omitting the `names` param as empty.

No

## Acceptance Steps
1. Apply a `CFDomain` CR to the cluster
2. `curl /v3/domains` and see the domain in the response


acosta11 (approved):
LGTM

Issue #209, opened 2021-11-04
[Chore]: Bump API Version to latest capi-release and include cf-k8s suffix


PureMunky:
### Background

When connecting to the API using the CLI this warning is presented.

“Warning: Your targeted API’s version (3.90.0) is less than the minimum supported API version (3.99.0). Some commands may not function correctly.”

### Action to take

Update the API to display the latest version: `3.108.0+cf-k8s`

### Impact

This is a non-breaking interaction but the CLI response would be seen by a user.

tcdowney:
@gerg do y'all have any recommendations for what we should do here? From a certain point of view what the CLI is saying is accurate -- some commands may not function correctly. 🙂 

We'll never be fully compatible so we can never truly say we meet a given v3 API version, but for what we _do_ support we are as compatible as possible and our users should already be aware of our V3 API Shim's limitations.

The way I see it, we could either:
1. Add some automation to bump this every time there is a capi-release
2. Artificially say we are something higher like `4.0.0`. Do you know if that would negatively affect the CLI?
3. ??? open to suggestions

For the short-term we're going to manually bump to `3.99.0`, but wanted to get some thoughts for a longer term solution.

tcdowney:
@PureMunky I had a discussion with some folks on the CLI and CAPI sides.

They brought up a good point that operators sometimes refer to this version when reasoning about CVEs and whether or not they're affected, so we probably shouldn't completely shadow the version since that could be misleading.

We could suffix it with some descriptor like `3.108.0-k8s-adapter` or something to disambiguate it. I tried this out and the CLI was fine with it since that's still valid semver.

cc @emalm as well

emalm:
@tcdowney nice, would an identifier like `3.108.0+cf-k8s` that uses the [build metadata segment of semver](https://semver.org/#spec-item-10) be the best fit, if the CLI parses it without complaint?

tcdowney:
@emalm yeah, CLI is handling that well!

With an unsupported version with the suffix it correctly still warns:
```
Warning: Your targeted API's version (3.98.0+cf-k8s) is less than the minimum supported API version (3.99.0). Some commands may not function correctly.
```

With the latest version with the suffix it happily proceeds:
```
API endpoint:   https://api.brook-piper.capi.land
API version:    3.108.0+cf-k8s
```

PR #247, opened 2021-11-16
Bump v3 API Version

tcdowney:
Addresses https://github.com/cloudfoundry/cf-k8s-controllers/issues/209

## What is this change about?
Bumps the presented v3 API version so that the `cf` CLI stops complaining and appends a `+cf-k8s` suffix to the version to disambiguate it from Cloud Controller

No

## Acceptance Steps
1. Hit the root `/` endpoint and see the version updated
2. Target the API with the `cf` CLI and see that it doesn't error


akrishna90 (approved):
LGTM

PR #253, opened 2021-11-17
Change ginkgo flags to v2 counterparts

danail-branekov:
- Replace deprecated ginkgo command line arguments with the new ones
- Set the e2e slow test threshold
- remove the .test file that was accidentally committed


kieron-dev (approved):
LGTM

PR #254, opened 2021-11-17
Update the sample CFRoute appRef to point at the sample CFApp

davewalter:
No

## What is this change about?
This PR updates the sample CFRoute appRef to point at the sample CFApp. This means that applying all of the sample resources should result in a routable workload.

No

## Acceptance Steps
Deploy the sample resources to a suitably configured cluster running the API and controllers and verify that the CFRoute resource results in a service that correctly targets the pods for the workload created from the CFApp/CFPackage/CFBuild resources.

### DIFF
--- a/controllers/config/samples/cfroute.yaml
+++ b/controllers/config/samples/cfroute.yaml
@@ -14,5 +14,5 @@ spec:
     - guid: 8ad77ef4-53d5-117a-b640-0ae227a13f35
       port: 8080
       appRef:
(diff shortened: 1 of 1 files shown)


gnovv (approved):
lgtm

tcdowney (approved):
`14dcda7d-1fa1-4a91-b437-fbdba20e8c5a` is a nice guid

LGTM as well

Issue #222, opened 2021-11-08
[Feature]: API Client can hit _minimal_ log-cache API endpoints


tcdowney:
### Background

The CF CLI attempts to fetch app staging logs from log-cache as part of `cf push`.

It does this via the following process:

1. Discovers the address of the log-cache API via Cloud Controller's `/` root endpoint
2. Calls the log-cache API's `/api/v1/info` endpoint to discover version info
3. Calls the log-cache API's `/api/v1/read/<app-guid>` endpoint to fetch log envelopes

We will have more functional support for logs as part of a later story, but for our `cf push` MVP we just don't want these calls to error.

**As a** client of the API shim
**I want** minimal implementations of the log-cache endpoints to exist
**So that** I can use my existing code and scripts to push apps

### API shim root endpoint
**GIVEN** I have deployed the API shim
**WHEN I** hit the root discovery endpoint
```console
curl "https://api-shim.example.org"
```
**THEN I** I see a `log-cache` entry in the response self-referencing the API shim's address
```
{
  "links": {
    "self": {
      "href": "https://api-shim.example.org"
    },
    "bits_service": null,
    "cloud_controller_v2": null,
    "cloud_controller_v3": {
      "href": "https://api-shim.example.org/v3",
      "meta": {
        "version": "3.90.0+cf-k8s"
      }
    },
    "network_policy_v0": null,
    "network_policy_v1": null,
    "login": null,
    "uaa": null,
    "credhub": null,
    "routing": null,
    "logging": null,
     "log_cache": {
      "href": "https://api-shim.example.org/" <-------------------------- new entry
    },
    "log_stream": null,
    "app_ssh": null
  }
}
```

---

### log-cache info endpoint
**GIVEN** I have deployed the API shim
**WHEN I** hit the `/api/v1/info` endpoint
```console
curl "https://api-shim.example.org/api/v1/info"
```
**THEN I** I see the following hard-coded response
```json
{"version":"2.11.4+cf-k8s","vm_uptime":"0"}
```

---

### log-cache read endpoint
**GIVEN** I have deployed the API shim
**WHEN I** hit the `/api/v1/read/<app-guid>` endpoint
```console
curl "https://api-shim.example.org/api/v1/read/<app-guid>?descending=true&limit=1&start_time=-6795364578871345152"
```
**THEN I** I see the following hard-coded response
```json
{"envelopes":{"batch":[]}}
```

[...]

PR #255, opened 2021-11-17
Add minimal log-cache API endpoints to support cf push

tcdowney:
- https://github.com/cloudfoundry/cf-k8s-controllers/issues/222

## What is this change about?
Adds basic log-cache endpoints that return hard-coded empty responses to support `cf push`. No logs will be returned, but at least the CLI won't error.

No

## Acceptance Steps
Follow the A/C on https://github.com/cloudfoundry/cf-k8s-controllers/issues/222.

### DIFF
--- a/api/apis/log_cache_handler.go
+++ b/api/apis/log_cache_handler.go
@@ -1,0 +1,40 @@
+package apis
+
+import (
+	"net/http"
+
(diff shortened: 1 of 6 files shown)


gnovv (approved):
lgtm

PR #258, opened 2021-11-17
Give the API shim the ability to create/update secrets

tcdowney:
## What is this change about?
Give the API shim the ability to create/update secrets so that it is able to create/update app env vars.

@matt-royal @akrishna90 @Birdrock

### DIFF
--- a/api/config/base/rbac/role.yaml
+++ b/api/config/base/rbac/role.yaml
@@ -18,6 +18,14 @@ rules:
   - ""
   resources:
   - secrets
+  verbs:
+  - create
+  - patch
+  - update
+- apiGroups:
+  - ""
+  resources:
+  - secrets
   - serviceaccounts
   verbs:
   - get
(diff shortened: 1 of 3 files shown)


gnovv (approved):
lgtm

api/config/base/rbac/role.yaml
```diff
@@ -23,6 +23,9 @@ rules:
   - get
   - list
   - watch
+  - create
```
tcdowney, line 26: Are these generated somewhere else?

tcdowney: I just realized when I run `make build-reference-api` it deletes them.

akrishna90 (changes requested):
@tcdowney - the file you have edited is autogenerated, so these changes are not preserved. 
Please add them as kubebuilder annotations

api/config/base/rbac/role.yaml
akrishna90: Add them as kubebuilder annotations, you'll be fine

tcdowney: Thanks I added them to the apps repo and pushed a new commit! 👍

akrishna90 (approved):

PR #259, opened 2021-11-17
Fix the app-guid label applied to the app env var secret

davewalter:
No

## What is this change about?
This PR updates the `app-guid` label applied to the app env var secret to match similar labels applied to other resources.

No

## Acceptance Steps
Push an app with an env var and check the label on the secret that is created.

### DIFF
--- a/api/repositories/app_repository.go
+++ b/api/repositories/app_repository.go
@@ -34,7 +34,7 @@ const (
 	Kind            string = "CFApp"
 	APIVersion      string = "workloads.cloudfoundry.org/v1alpha1"
 	TimestampFormat string = time.RFC3339
-	CFAppGUIDLabel  string = "apps.cloudfoundry.org/appGuid"
+	CFAppGUIDLabel  string = "workloads.cloudfoundry.org/app-guid"
 )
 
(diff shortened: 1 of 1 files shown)


api/repositories/app_repository.go
```diff
@@ -34,7 +34,7 @@ const (
 	Kind            string = "CFApp"
 	APIVersion      string = "workloads.cloudfoundry.org/v1alpha1"
 	TimestampFormat string = time.RFC3339
-	CFAppGUIDLabel  string = "apps.cloudfoundry.org/appGuid"
+	CFAppGUIDLabel  string = "workloads.cloudfoundry.org/app-guid"
```
tcdowney, line 37: Nothing referenced this label?

tcdowney (approved):

api/repositories/app_repository.go
davewalter: I couldn't find any other references to `appGuid` in the code. The tests use the "production" code so they still passed.

tcdowney: must have just been a convenience label we added 👍

PR #263, opened 2021-11-18
Generate Version 4 UUIDs

Birdrock:
No.

## What is this change about?
Use Version 4 UUID.

No.

## Acceptance Steps
Run tests as normal

@emalm


davewalter (commented):
Could we get rid of the `generateGUID()` functions everywhere and just call `uuid.NewString()` directly?

Birdrock:
Definitely can. I'll make that change.

Issue #266, opened 2021-11-18
[Feature]: API Client can filter apps by name and space guid


tcdowney:
### Blockers/Dependencies

* Follow up to https://github.com/cloudfoundry/cf-k8s-api/issues/6

### Background

The CF CLI hits the `GET /v3/apps?names=<app-name>&space_guids=<space-guid> endpoint as part of push to be able to retrieve an app from the API while only knowing its name. Currently our `GET /v3/apps` endpoint returns all apps when the CLI expects a single app to be returned, so this endpoint causes push to display info/act on the wrong app.

**As a** Client of the cf-k8s API shim
**I want** to be able to to filter a list of apps by names and space guids
**So that** I can retrieve a limited subset of apps and be able to look up a particular app my name/space.

### Multiple Apps Exist with Same Name
**GIVEN** I have multiple apps with the same name (but in different spaces)
**WHEN I** make the following request
```
curl "https://api.example.org/v3/apps?names=<app-name>&space_guids=<space-guid>" \
  -X GET \
  -H "Authorization: bearer [token]"
```
**THEN I** get a response containing a list of 1 app that matches the name/space specified

```json
HTTP/1.1 200 OK
Content-Type: application/json

  {
    "pagination": {
      "total_results": 1,
      "total_pages": 1,
      "first": {
        "href": "https://api.example.org/v3/apps?page=1"
      },
      "last": {
        "href": "https://api.example.org/v3/apps?page=1"
      },
      "next": null,
      "previous": null
    },
    "resources": [
      {
        "guid": "1cb006ee-fb05-47e1-b541-c34179ddc446",
        "name": "my_app",
        "state": "STARTED",
        "created_at": "2016-03-17T21:41:30Z",
        "updated_at": "2016-03-18T11:32:30Z",
        "lifecycle": {
          "type": "buildpack",
          "data": {
            "buildpacks": ["java_buildpack"],
            "stack": "cflinuxfs3"
          }
        },
        "relationships": {
          "space": {
            "data": {
              "guid": "space-guid-1"
            }
          }
        },
        "links": {
          "self": {
            "href": "https://api.example.org/v3/apps/1cb006ee-fb05-47e1-b541-c34179ddc446"
          },
          "space": {
            "href": "https://api.example.org/v3/spaces/2f35885d-0c9d-4423-83ad-fd05066f8576"
          },
          "processes": {
[...]

PR #270, opened 2021-11-18
Add ability to filter GET /v3/apps by name and space guid

tcdowney:
Issue: https://github.com/cloudfoundry/cf-k8s-controllers/issues/266

## What is this change about?
Add ability to filter GET /v3/apps by name and space guid. The CLI requires these filters to be able to fetch an app by name during `cf push`.

No

## Acceptance Steps
Apply multiple apps in different namespaces and play around with the filters. E.g.

```
curl "localhost:9000/v3/apps?names=tim-node&space_guids=49c31b07-bf89-45c0-8eef-401efd0fa7e5" | jq .
```

@gnovv @akrishna90 @davewalter @acosta11


api/repositories/app_repository.go
```diff
@@ -156,22 +161,70 @@ func (f *AppRepo) CreateApp(ctx context.Context, client client.Client, appCreate
...
+				}
+			}
+		} else if spaceGUIDFilterSpecified {
+			for _, spaceGUID := range message.SpaceGuids {
+				if appBelongsToSpace(app, spaceGUID) {
+					filtered = append(filtered, app)
+				}
+			}
+		}
 	}
```
tcdowney, line 208: I'm not too happy with the logic I came up with here to try and perform the filter union.

The most natural thing IMO would be to start with the whole list and remove that don't match each particular filter. Since Go doesn't have a built-in remove I implemented my own and saw how many times I was iterating over the collection of apps and just settled with this for now.

As we add more filtering and more complex filters across endpoints we might want to revisit this. Maybe use generics 😈

gnovv: This seems fine for now. 
For the CRD explore we took a similar approach as you described:
https://github.com/cloudfoundry/cf-crd-explorations/blob/8cf10d5aafec5695394c1f4eebf21ce481ca800b/cfshim/filters/app_filter.go#L13

gnovv (approved):
lgtm

PR #271, opened 2021-11-18
Fix flaky app filtering test

tcdowney:
## What is this change about?
Fixes a flaky test I introduced in https://github.com/cloudfoundry/cf-k8s-controllers/pull/270 😔 

- The order that the apps get returned by the K8s API is
non-deterministic
- Issue https://github.com/cloudfoundry/cf-k8s-controllers/issues/266

@gnovv


gnovv (approved):
D:

Issue #170, opened 2021-10-04
[Feature]: App developer can restart an application w/ downtime


Birdrock:
### Background

**As a** App developer
**I want** to be able to issue a restart command
**So that** restart my workload

### Happy Path

**GIVEN** my app exists
**WHEN I** issue a start command
```
curl "https://api.example.org/v3/apps/<guid>/actions/restart" \
  -X POST \
  -H "Authorization: bearer <token>"
```

**THEN I** get a response similar to the following
```
HTTP/1.1 200 OK
Content-Type: application/json

{
  "guid": "1cb006ee-fb05-47e1-b541-c34179ddc446",
  "name": "my_app",
  "state": "STARTED",
  "created_at": "2016-03-17T21:41:30Z",
  "updated_at": "2016-03-18T11:32:30Z",
  "lifecycle": {
    "type": "buildpack",
    "data": {
      "buildpacks": ["java_buildpack"],
      "stack": "cflinuxfs3"
    }
  },
  "relationships": {
    "space": {
      "data": {
        "guid": "2f35885d-0c9d-4423-83ad-fd05066f8576"
      }
    }
  },
  "links": {
    "self": {
      "href": "https://api.example.org/v3/apps/1cb006ee-fb05-47e1-b541-c34179ddc446"
    },
    "space": {
      "href": "https://api.example.org/v3/spaces/2f35885d-0c9d-4423-83ad-fd05066f8576"
    },
    "processes": {
      "href": "https://api.example.org/v3/apps/1cb006ee-fb05-47e1-b541-c34179ddc446/processes"
    },
    "packages": {
      "href": "https://api.example.org/v3/apps/1cb006ee-fb05-47e1-b541-c34179ddc446/packages"
    },
    "environment_variables": {
      "href": "https://api.example.org/v3/apps/1cb006ee-fb05-47e1-b541-c34179ddc446/environment_variables"
    },
    "current_droplet": {
      "href": "https://api.example.org/v3/apps/1cb006ee-fb05-47e1-b541-c34179ddc446/droplets/current"
    },
    "droplets": {
      "href": "https://api.example.org/v3/apps/1cb006ee-fb05-47e1-b541-c34179ddc446/droplets"
    },
    "tasks": {
      "href": "https://api.example.org/v3/apps/1cb006ee-fb05-47e1-b541-c34179ddc446/tasks"
    },
    "start": {
[...]

PR #273, opened 2021-11-18
Increase response timeout for the API

tcdowney:
Not specifically, but it was uncovered by https://github.com/cloudfoundry/cf-k8s-controllers/issues/170

## What is this change about?
- By default Contour/Envoy set a 15 second timeout for the API shim to
respond to requests
  - https://projectcontour.io/docs/main/config/api/#projectcontour.io/v1.TimeoutPolicy
- Certain endpoints like app restart and package upload are known to take longer than this
- Eventually we should deal with timeouts more holistically, but for now let's just set it to something high

No

## Acceptance Steps
Restarting an app should be less flaky. Prior to this you might see behavior and errors like this:
```
Waiting for app node to start...

REQUEST: [2021-11-18T15:47:52-07:00]
POST /v3/apps/c3147553-5639-4094-8d95-00f789734577/actions/restart HTTP/1.1
Host: api.example.com
Accept: application/json
Authorization: [PRIVATE DATA HIDDEN]
User-Agent: cf/8.0.0+a5e5680e0.2021-11-15 (go1.17.1; amd64 darwin)
[application/json Content Hidden]

RESPONSE: [2021-11-18T15:48:07-07:00]
HTTP/1.1 504 Gateway Timeout
Content-Length: 24
Content-Type: text/plain
Date: Thu, 18 Nov 2021 22:48:07 GMT
Server: envoy
upstream request timeout

Unexpected Response
Response Code: 504
Code: 0, Title: , Detail: upstream request timeout
```

After this change running `cf curl -X POST /v3/apps/$(cf app APP_NAME --guid)/actions/restart` should succeed more reliably.

@davewalter @acosta11 @akrishna90

### DIFF
--- a/api/config/base/ingress.yaml
+++ b/api/config/base/ingress.yaml
@@ -15,3 +15,5 @@ spec:
     services:
     - name: cf-k8s-api-svc
       port: 80
+    timeoutPolicy:
+      response: "5m"
--- a/api/reference/cf-k8s-api.yaml
+++ b/api/reference/cf-k8s-api.yaml
@@ -304,6 +304,8 @@ spec:
     services:
     - name: cf-k8s-api-svc
       port: 80
+    timeoutPolicy:
+      response: 5m
   virtualhost:
     fqdn: ""
     tls:


akrishna90 (approved):
Good job hunting down the root cause of the issue ! 👏🏽

PR #276, opened 2021-11-19
Add authentication middleware

gcapizzi:
This adds a middleware that will block any request coming from users with missing or invalid authentication credentials.

### DIFF
--- a/api/apis/authentication_middleware.go
+++ b/api/apis/authentication_middleware.go
@@ -1,0 +1,49 @@
+package apis
+
+import (
+	"net/http"
+
+	"code.cloudfoundry.org/cf-k8s-controllers/api/repositories/authorization"
+	"github.com/go-http-utils/headers"
+)
+
+type AuthenticationMiddleware struct {
+	identityProvider         IdentityProvider
+	unauthenticatedEndpoints map[string]interface{}
+}
+
+func NewAuthenticationMiddleware(identityProvider IdentityProvider) *AuthenticationMiddleware {
+	return &AuthenticationMiddleware{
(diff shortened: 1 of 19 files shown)


tcdowney:
Nice!
```
09:51 $ cf curl /v3/processes/51f740bd-494d-11ec-8560-ce6c6a3b098d/stats

{"resources":[{"type":"web","index":0,"state":"RUNNING","host":null,"instance_ports":[],"uptime":null,"mem_quota":null,"disk_quota":null,"fds_quota":null,"isolation_segment":null,"details":null}]}

09:51 $ curl localhost:9000/v3/processes/51f740bd-494d-11ec-8560-ce6c6a3b098d/stats

{"errors":[{"detail":"Authentication error","title":"CF-NotAuthenticated","code":10002}]}
```

tcdowney:
@gcapizzi If you're using `curl` instead of the `cf` CLI what would be the best way to manually set up the `Authorization` header?

On CF for VMs I would use `curl -H "Authorization: $(cf oauth-token)" ...`.

gcapizzi:
> @gcapizzi If you're using `curl` instead of the `cf` CLI what would be the best way to manually set up the `Authorization` header?
> 
> On CF for VMs I would use `curl -H "Authorization: $(cf oauth-token)" ...`.

We don't have a good way at the moment 😕  We could implement `cf oauth-token` though! I'll add a story and we can see when to prioritise it.

danail-branekov (commented):
lgtm

danail-branekov (approved):

Issue #252, opened 2021-11-17
[Chore]: Custom Ginkgo fail handler in e2e tests to collect K8S logs


danail-branekov:
### Background

Currently, if a e2e test fails for an unexpected reason it is very hard to reason about the failure as Ginkgo output does not contain K8S logs. In order to improve e2e tests flake hunting and general troubleshooting we could implement a fail handler that collects the controller and api logs on a test failure and prints them to the Ginkgo writer. 

As reference, we implemented a [similar fail handler](https://github.com/cloudfoundry-incubator/eirini-controller/blob/2aadf7a1feba960048c96d6204eef5598c8f8a79/tests/fail_handler.go) in Eirini back in time.

### Action to take

see the description above

### Impact

better test failure troubleshooting

PR #278, opened 2021-11-22
Capture Kubernetes logs on e2e failures

gcapizzi:
This introduces a Ginkgo fail handler that will capture logs from `Pod`s we are interested in on end-to-end test failures. This can get quite verbose, so we'll possibly have to tweak the number of lines printed, but it's super useful to understand the reason tests are failing, especially for flakes.

### DIFF
--- a/api/tests/e2e/helpers/fail_handler.go
+++ b/api/tests/e2e/helpers/fail_handler.go
@@ -1,0 +1,107 @@
+package helpers
+
+import (
+	"context"
+	"fmt"
+	"io/ioutil"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	controllerruntime "sigs.k8s.io/controller-runtime"
+)
+
+const (
+	logTailLines = 50
+)
+
+type podContainerDescriptor struct {
+	Namespace  string
+	LabelKey   string
+	LabelValue string
+	Container  string
+}
+
+func E2EFailHandler(message string, callerSkip ...int) {
+	config, err := controllerruntime.GetConfig()
+	Expect(err).NotTo(HaveOccurred())
+
+	clientset, err := kubernetes.NewForConfig(config)
+	Expect(err).NotTo(HaveOccurred())
+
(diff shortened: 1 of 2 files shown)


georgethebeatle:
Does this close this chore: https://github.com/cloudfoundry/cf-k8s-controllers/issues/252 ?

tcdowney (approved):
LGTM once tests pass

api/tests/e2e/helpers/fail_handler.go
```diff
@@ -0,0 +1,107 @@
...
+	podLogOpts := corev1.PodLogOptions{TailLines: int64ptr(logTailLines), Container: container}
+	req := clientset.CoreV1().Pods(pod.Namespace).GetLogs(pod.Name, &podLogOpts)
+
+	logStream, err := req.Stream(context.Background())
+	if err != nil {
+		return "", err
+	}
+	defer logStream.Close()
+
+	logBytes, err := ioutil.ReadAll(logStream)
```
tcdowney, line 97: Unrelated to this PR, but wanted to run a thought by y'all. 

Do you think this approach could be used to implement the log-cache API READ envelope endpoint? It seems promising... pod logs are a subresource of `Pods` so could be RBAC-ed. That'd handle the access control question. And we could then stream them into something that converts them into the loggregator envelope format

Right now we've just got a hard-coded implementation that always returns an empty array of envelopes, but we're thinking about what the future of this looks like.
https://github.com/cloudfoundry/cf-k8s-controllers/issues/222

danail-branekov: I doubt how reliable would such an implementation be. 
Pods are being managed by K8s, meaning that when a pod is killed, its logs are lost. Therefore, if an app is constantly crashing, its logs are going to be immediately lost after the crash so and endpoint that just reads running pods logs would not be much helpful.

tcdowney: Right now it's `cf logs` returns no logs so we're looking for ways for it to return _some_. I think in the scenario you describe even cf-for-k8s with it's log-caches and fluentds had trouble with it since they didn't get logs before the containers crashed.

Rather than re-implement the whole CF logging subsystem, though, I anticipate advising operators to use existing K8s log aggregators for longer-term storage.

gcapizzi: I don't understand how log streaming works in log-cache given their response is a JSON object 🤔

Issue #151, opened 2021-10-22
[Chore]: `make build-reference` should be a target of another commonly executed make target so we don't forget to run it


kieron-dev:
### Background

There is a github action on cf-k8s-api, and cf-k8s-controllers repositories, which ensure the reference YAML is up to date.

Unfortunately the developer needs to remember to manually run `make build-reference` to ensure this is the case.

It would be better if `build-reference` was a target of another commonly run target so by the time all tests have been run with make and passed, then the reference YAML would automatically be up to date.

### Impact

Saves a failed test run on each PR.

PR #282, opened 2021-11-23
Ensure build-reference gets run so that check always succeeds

georgethebeatle:
#151

## What is this change about?
Make sure the yaml reference is updated automatically when e2e tests are run. This way we will avoid yaml drift failures.

No

## Acceptance Steps
make test-e2e should update the yaml reference 

@cloudfoundry/cf-k8s


kieron-dev (approved):
LGTM

PR #293, opened 2021-11-26
Only list apps the user has access to

gcapizzi:
This is a first shot at the implementation of apps filtering per user. The current version is quite naive, as it tries to list apps in every namespace on the system, skipping the ones that return a `403 Forbidden`. We plan to optimise this by restricting the listing to the namespaces the user has bindings into, but we will do this in a following PR.


danail-branekov (approved):
lgtm

PR #294, opened 2021-11-26
Make checks a bit faster

georgethebeatle:
No

## What is this change about?
Remove dependency of e2e tests to other tests thus making execution more
parallel. Pushing of the docker images now depends on all tests, since
they run in a more parallel fashion.

No

## Acceptance Steps
Checks are green.

@cloudfoundry/cf-k8s


danail-branekov (approved):
lgtm

Issue #100, opened 2021-10-14
[Chore]: Test Flake App Validating Webhook


gnovv:
### What happened?

```
Random Seed: 1634250989
Will run 6 of 6 specs

••••2021-10-14T22:36:38.521Z	DEBUG	controller-runtime.webhook.webhooks	received request	{"webhook": "/mutate-workloads-cloudfoundry-org-v1alpha1-cfapp", "UID": "64e4c2eb-e02a-4e64-9e38-68b287809ad2", "kind": "workloads.cloudfoundry.org/v1alpha1, Kind=CFApp", "resource": {"group":"workloads.cloudfoundry.org","version":"v1alpha1","resource":"cfapps"}}
2021-10-14T22:36:38.521Z	INFO	cfapp-resource	Mutating CFApp webhook handler	{"name": "test-app-guid"}
2021-10-14T22:36:38.522Z	DEBUG	controller-runtime.webhook.webhooks	wrote response	{"webhook": "/mutate-workloads-cloudfoundry-org-v1alpha1-cfapp", "code": 200, "reason": "", "UID": "64e4c2eb-e02a-4e64-9e38-68b287809ad2", "allowed": true}
2021-10-14T22:36:38.523Z	DEBUG	controller-runtime.webhook.webhooks	received request	{"webhook": "/validate-workloads-cloudfoundry-org-v1alpha1-cfapp", "UID": "993c15f8-a1f5-4cfa-8a69-e85c853c5237", "kind": "workloads.cloudfoundry.org/v1alpha1, Kind=CFApp", "resource": {"group":"workloads.cloudfoundry.org","version":"v1alpha1","resource":"cfapps"}}
2021-10-14T22:36:38.523Z	INFO	cfapp-validate	Validate	{"name": "test-app-guid"}
2021-10-14T22:36:38.523Z	INFO	cfapp-validate	CFApp with the same spec.name exists	{"name": "test-app-guid"}
2021-10-14T22:36:38.523Z	DEBUG	controller-runtime.webhook.webhooks	wrote response	{"webhook": "/validate-workloads-cloudfoundry-org-v1alpha1-cfapp", "code": 403, "reason": "{\"code\":1,\"message\":\"CFApp with the same spec.name exists\"}", "UID": "993c15f8-a1f5-4cfa-8a69-e85c853c5237", "allowed": false}

------------------------------
• Failure in Spec Setup (BeforeEach) [0.006 seconds]
[...]

PR #305, opened 2021-11-29
Make the cfapp validation test more resilient

danail-branekov:
https://github.com/cloudfoundry/cf-k8s-controllers/issues/100

## What is this change about?
Randomize the names of the objects that the test create in order to avoid name collisions during parallel test execution

No

## Acceptance Steps
All tests should pass

@mnitchev


mnitchev (approved):
LGTM

Issue #158, opened 2021-10-18
[Feature]: Developer should only be able to list apps they have access to


gcapizzi:
### Background

**As an** admin
**I want** to restrict app visibility for developers
**So that** I can maintain access to apps segregated

We want to keep this behaviour behind a **feature flag** for now, to avoid having a mix of authenticated and unauthenticated endpoints exposed.

### Acceptance Criteria

**GIVEN** the spaces `space1`, `space2` and `space3` exist
**AND** the apps `app1` and `app2` exist in `space1`
**AND** the apps `app3` and `app4` exist in `space2`
**AND** the apps `app5` and `app6` exist in `space3`
**AND** `user` is a `SpaceDeveloper` in `space1` and `space3`
**WHEN I** `GET /v3/apps` as `user`
**THEN I** see that only `app1`, `app2`, `app5` and `app6` are part of the response

### Dev Notes

* This should drive out the implementation of a user-specific K8s client to talk to the API, _in addition_ to the privileged client. We will probably need the privileged client to figure out in which spaces the developer has bindings, but we should use the user client to list the apps in each individual space, so to leverage Kubernetes authorization.
* This should also drive the creation of a `SpaceDeveloper` role. For now, we only want to add the `list` permission strictly necessary for this story, not fully reproduce CF's `SpaceDeveloper` yet.

georgethebeatle:
PR #num-here adds authorization checks to the app listing endpoint, but does so in a pretty inefficient way (lists apps in every namespace on the cluster). What's left to be done in order to close that story is to improve this as described in option 2 in the [proposal document](https://docs.google.com/document/d/1M02YxnUfNRZw8-5W9TfWj3jghCPG1M-Pracgk6BSV1I/edit#).

gcapizzi:
@georgethebeatle isn't this story done after #414 has been merged?

PR #306, opened 2021-11-29
Add authorization info to the request context

georgethebeatle:
#158

## What is this change about?
This is a preliminary refactoring that has the intent of making the list apps authorization logic easier to implement.
The general idea is so stop passing the auth header down to the repositories but instead add a type safe parsed
`authorization.Info` object to the request context so that any repository that may need user information can look it
up and use it however it needs. 

No

## Acceptance Steps
All tests pass. This is a refactoring preserving current behaviour.

@cloudfoundry/cf-k8s


danail-branekov (approved):
lgtm

Issue #285, opened 2021-11-24
`SubnamespaceAnchorValidation` test doesn't get an error sometimes


gcapizzi:
```
• [FAILED] [0.187 seconds]
SubnamespaceanchorValidation updating a space the new space name already exists in the root CF namespace [It] fails
/home/runner/work/cf-k8s-controllers/cf-k8s-controllers/controllers/webhooks/workloads/integration/subnamespaceanchor_validation_test.go:214

  Begin Captured GinkgoWriter Output >>
    2021-11-24T11:51:29.498Z	DEBUG	controller-runtime.webhook.webhooks	received request	{"webhook": "/validate-hnc-x-k8s-io-v1alpha2-subnamespaceanchor", "UID": "1a0cb90e-3240-4e1e-8016-e716b12d5340", "kind": "hnc.x-k8s.io/v1alpha2, Kind=SubnamespaceAnchor", "resource": {"group":"hnc.x-k8s.io","version":"v1alpha2","resource":"subnamespaceanchors"}}
    2021-11-24T11:51:29.599Z	DEBUG	controller-runtime.webhook.webhooks	wrote response	{"webhook": "/validate-hnc-x-k8s-io-v1alpha2-subnamespaceanchor", "code": 200, "reason": "", "UID": "1a0cb90e-3240-4e1e-8016-e716b12d5340", "allowed": true}
    2021-11-24T11:51:29.608Z	DEBUG	controller-runtime.webhook.webhooks	received request	{"webhook": "/validate-hnc-x-k8s-io-v1alpha2-subnamespaceanchor", "UID": "08e2eb65-5d05-475a-9110-af4127f3943f", "kind": "hnc.x-k8s.io/v1alpha2, Kind=SubnamespaceAnchor", "resource": {"group":"hnc.x-k8s.io","version":"v1alpha2","resource":"subnamespaceanchors"}}
    2021-11-24T11:51:29.608Z	DEBUG	controller-runtime.webhook.webhooks	wrote response	{"webhook": "/validate-hnc-x-k8s-io-v1alpha2-subnamespaceanchor", "code": 200, "reason": "", "UID": "08e2eb65-5d05-475a-9110-af4127f3943f", "allowed": true}
[...]

PR #310, opened 2021-11-30
Prevent duplication org/space naming using leases

kieron-dev:
#285

## What is this change about?
Listing existing names in a validation webhook to prevent duplicate names is inherently racy. The change uses the idea of leases discussed in [this proposal](https://docs.google.com/document/d/1fYIulNhfXlp_y2Ah5EtoKDDX83s6V-WRu162xpziLZ8/edit#) to implement the webhook robustly.

No

## Acceptance Steps
See repro in issue #285.


controllers/webhooks/workloads/subnamespaceanchor_validation_test.go
```diff
@@ -324,41 +351,183 @@ var _ = Describe("SubnamespaceanchorValidation", func() {
...
+			When("the org has a space label as well", func() {
+				BeforeEach(func() {
+					anchor.Labels[workloads.SpaceNameLabel] = "my-space"
+				})
+
+				It("allows the deletion anyway", func() {
 					Expect(response.Allowed).To(BeTrue())
 				})
+
+				It("does not attempt to deregister any names", func() {
```
danail-branekov, line 506: We should probably have the `does not attempt to deregister any names` test in the previous context (`the anchor has no org or space labels`)

controllers/coordination/name_registry.go
```diff
@@ -0,0 +1,123 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      hashName(r.entityType, name),
+			Namespace: namespace,
+		},
+	}
+
+	copy := lease.DeepCopy()
+	copy.Spec.HolderIdentity = &unlockedIdentity
+
+	if err := r.client.Patch(ctx, copy, client.MergeFrom(lease)); err != nil {
```
danail-branekov, line 110: @georgethebeatle and I were wondering whether we should allow unlocking names that are not locked, i.e. use the `test-replace` patch operation similar to the one in `TryLockName`. 
On one hand, not doing that makes the unlock operation idempotent, on the other, if you are unlocking an unlocked object, that might be an unexpected state (like when you find your car opened, you would start wondering whether someone else has the keys :))
What do you think?

danail-branekov (approved):
LGTM, just consider whether my comments make sense. Otherwise, feel free to merge

controllers/webhooks/workloads/subnamespaceanchor_validation_test.go
kieron-dev: Thanks! We've added some more checks for not calling the registry in the failure cases.

controllers/coordination/name_registry.go
kieron-dev: This seems reasonable. We've made the change :)

Issue #302, opened 2021-11-26
[Chore]: Increase test parallelism


georgethebeatle:
### Background

At one point the ginkgo test parallelism was bumped down in order to reduce the number of flakes we were observing. However flakes due to parallelism may be a manifestation of code or test problems, that better be fixed. What's more, increased parallelism means faster CI.

### Action to take

Remove the restriction of ginkgo nodes.

### Impact

We have faster, more stable tests.

danail-branekov:
Today we addressed the [limit of two ginkgo nodes](https://github.com/cloudfoundry/cf-k8s-controllers/blob/dfe90f800c2a00fb5934b2df34e861b2031b7034/scripts/run-tests.sh#L31) when running tests.

It turned out that on CI tests are always run with 2 nodes so this limitation is only relevant to local runs. Also, we believe that the flaky tests are in controllers only (the ones that are testing uniqueness validating webhooks) so we preserved the two nodes limit to controllers tests only. The rationale is that we would like to get local run speedups, but at the same time minimise the probability of seeing flakes when testing locally.
It should be mentioned that controller tests still flake with two nodes but for some reason that happens much more frequently when running with more nodes. Once the validating webhook flakes are addressed, we should remove the nodes limit completely.

PR #312, opened 2021-11-30
Only limit ginkgo nodes for controllers tests

danail-branekov:
https://github.com/cloudfoundry/cf-k8s-controllers/issues/302

## What is this change about?
Get rid of limiting Ginkgo nodes (except for controllers tests) during test runs to make running tests faster

No

## Acceptance Steps
Run all tests, see them pass


mnitchev (approved):
LGTM

Issue #317, opened 2021-12-01
[Chore]: Make cf-k8s-controllers KpackImageTag and PackageRegistryBase tolerant to trailing slashes


Birdrock:
### Background

Current, `KpackImageTag` and `PackageRegistryBase` must be specified without a trailing slash. It should be tolerant to these specified with or without a trailing slash.

### Action to take

Add toleration for trailing slashes.

### Impact

Medium impact, as it will mitigate potential configuration issues. Small effort to resolve.

### Dev Notes

The relevant code snippets are here:
* https://github.com/cloudfoundry/cf-k8s-controllers/blob/9a978623c7153f7b7575f2269438926b1b3174fe/api/apis/package_handler.go#L213
* https://github.com/cloudfoundry/cf-k8s-controllers/blob/9a978623c7153f7b7575f2269438926b1b3174fe/controllers/controllers/workloads/cfbuild_controller.go#L203
* Could handle via the config struct or with ytt template validation, etc- up to implementor

Birdrock:
Resolved with #318

PR #318, opened 2021-12-02
Join paths with Go path lib

Birdrock:
#317 

## What is this change about?
Uses the path lib to join image paths.

No.

## Acceptance Steps
Run tests as normal.


Birdrock:
Can add test cases for trailing slash or no trailing slash, but that seems like it is just testing the go std lib.

gnovv (approved):
lgtm!

Issue #330, opened 2021-12-06
[Feature]: App Developer should not need to specify `--destination-protocol` when calling `cf map-route`


tcdowney:
### Background

**As an** App Developer
**I want** to be able to map `http` routes without explicitly specifying `--destination-protocol`
**So that** I can use my existing `cf` CLI scripts and workflows

### Acceptance Criteria

---
### Happy path: Omitting the `--destination-protocol` flag
**GIVEN** I have an app deployed
**WHEN I** call `cf map-route MY_APP MY_DOMAIN -n MY_HOST` without the `--destination-protocol` flag
**THEN I** see that the command succeeds and I see that an `http1` destination was added to my route

---
### Happy path: Setting the `--destination-protocol` to `http1`
**GIVEN** I have an app deployed
**WHEN I** call `cf map-route MY_APP MY_DOMAIN -n MY_HOST --destination-protocol http1`
**THEN I**  see that the command succeeds and I see that an `http1` destination was added to my route

---
### Unhappy path: Setting the `--destination-protocol` to `http`
**GIVEN** I have an app deployed
**WHEN I** call `cf map-route MY_APP MY_DOMAIN -n MY_HOST --destination-protocol http`
**THEN I** see that the command **fails** with an error saying that `protocol must be 'http1'`

---

### `http` vs `http1`
Apparently `http` was never a valid option here. It should be `http1` and we should prevent people from creating new routes with destinations set to `http`.

Relevant convo on Slack: https://cloudfoundry.slack.com/archives/C07C04W4Q/p1638814089030400

### Where to default
This brings back the age old question of where to do the defaulting. Cloud Controller defaults it based on the parent domain's type:

* https://github.com/cloudfoundry/cloud_controller_ng/blob/main/app/models/runtime/route_mapping_model.rb#L5
* https://github.com/cloudfoundry/cloud_controller_ng/blob/ed09b45936879975e9784927431cc5387e1221d4/app/models/runtime/route.rb#L187

IMO for this case it's fine to do it in the shim and CRD users should be forced to specify the protocol we want. Our API should allow for CLI users to omit, it though.

PR #338, opened 2021-12-08
Properly handle route destination protocol

davewalter:
#330 

## What is this change about?
This PR properly handles route destination protocol

- move defaulting of the protocol value from the presenter to the
  message
- allow the client to specify a destination protocol
- clean up confusing test values (http vs http1)
- remove label from sample CFDomain resource

No

## Acceptance Steps
See #330 

@acosta11

### DIFF
--- a/api/repositories/route_repository.go
+++ b/api/repositories/route_repository.go
@@ -29,7 +29,8 @@ type DestinationRecord struct {
 	AppGUID     string
 	ProcessType string
 	Port        int
-	// Weight and Protocol intentionally omitted as experimental features
+	Protocol    string
+	// Weight intentionally omitted as experimental features
 }
 
 type RouteRecord struct {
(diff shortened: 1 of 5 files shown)


api/repositories/route_repository.go
```diff
@@ -29,7 +29,8 @@ type DestinationRecord struct {
 	AppGUID     string
 	ProcessType string
 	Port        int
-	// Weight and Protocol intentionally omitted as experimental features
+	Protocol    string
```
tcdowney, line 32: When we fetch a route back out how is this set to `http1`? I don't see it on the CRD definition:
https://github.com/cloudfoundry/cf-k8s-controllers/blob/3e602c6042ab48fb01853e7089ab114de80b7396/controllers/apis/networking/v1alpha1/cfroute_types.go#L33

Is this just being presented as `http1` somewhere?

acosta11 (approved):
LGTM

acosta11:
Note this was merged a bit prematurely; We plan to fast follow with an update to the Route CRD to include the field as well instead of handling everything in the presenter. That change should address Tim's comment.

davewalter:
See #341 for the fast-follow PR.

PR #343, opened 2021-12-09
Temporarily pin buildpacks

davewalter:
No

## What is this change about?
This PR temporarily pins the buildpacks in the `cf-default-buildpacks` clusterstore to versions that are known to work with kpack v0.4.1.

No

## Acceptance Steps
Deploy as normal with the `scripts/install-dependencies.sh` script and verify that you can successfully push an app.

@cloudfoundry/cf-k8s


julian-hj (approved):
🤏

PR #346, opened 2021-12-10
Generate workload ingress secret in install-dependencies

danail-branekov:
No

## What is this change about?
Generate the workload ingress secret when running `install-dependencies.sh` as [described](https://github.com/cloudfoundry/cf-k8s-controllers/#configure-workload-ingress-tls-certificate-secret) in the README

No

## Acceptance Steps
1. Run `install-dependencies.sh`
2. See the `cf-k8s-workloads-ingress-cert` secret is create

### DIFF
--- a/scripts/generate-workload-ingress-certs-secret.sh
+++ b/scripts/generate-workload-ingress-certs-secret.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+
+echo "Will now generate tls.ca tls.crt and tls.key files"
+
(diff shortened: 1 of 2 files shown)


davewalter:
Hi @danail-branekov,

This shouldn't be required for deployment to a kind cluster. We added kustomize artifacts to disable system and workload TLS when deployed with `make deploy-kind`. For other types of clusters, the user should be creating a certificate based on the desired DNS names. Was there a problem that you encountered?

Regards,
Dave

Issue #334, opened 2021-12-07
[Feature]: App Developer can delete their apps with `cf delete APP_NAME` / `DELETE /v3/apps/:guid` endpoint


tcdowney:
### Blockers/Dependencies

There's a lot to unpack here, so going to try and break this out into separate stories. Let's consider this one blocked on https://github.com/cloudfoundry/cf-k8s-controllers/issues/335.

The goal of this initial story will be for the CLI command to succeed, the `CFApp` to be deleted, and no `Pods` should be running for it.

### Background

**As an** App Developer
**I want** I want to be able to use `cf delete APP_NAME` to delete my App
**So that** I can remove any trace of my app from the cluster

### Acceptance Criteria

---
### When the app exists
**GIVEN** I have an app deployed and running on the cluster
**WHEN I** run `cf delete APP_NAME`
**THEN I** see the command succeed and I can no longer see the corresponding `CFApp` on the cluster

```bash
$ cf delete dora
Really delete the app dora? [yN]: y
Deleting app dora in org org / space space as admin...
OK
```

---
### When the app no longer exists
**GIVEN** I already deleted my app (or never created it)
**WHEN I** run `cf delete APP_NAME`
**THEN I** see the command succeeds (it tries to be idempotent)

```bash
$ cf delete dora
Really delete the app dora? [yN]: y
Deleting app dora in org org / space space as admin...
OK
```

---
### DELETE /v3/apps/:guid Happy Path

**Request:**
```
curl "https://api.example.org/v3/apps/:app-guid" \
  -X DELETE \
  -H "Authorization: bearer [token]"
```

**Response:**
```
HTTP/1.1 202 Accepted
Location: https://api.example.org/v3/jobs/app.delete-<app-guid>
```

---

### DELETE /v3/apps/:guid App Doesn't Exist
**Request:**
```
curl "https://api.example.org/v3/apps/:non-existant-app-guid" \
  -X DELETE \
  -H "Authorization: bearer [token]"
```

**Response:**
```json
HTTP/1.1 404 Not Found

{
  "errors": [
    {
      "code": 10010,
      "detail": "App not found",
      "title": "CF-ResourceNotFound"
    }
  ]
}
```

---

### Requests made during `cf delete`
```
15:21 $ cf delete dora -f -v | grep -i REQUEST: -A2
REQUEST: [2021-12-07T15:22:34-07:00]
[...]

PR #360, opened 2021-12-13
Implement DELETE /v3/apps/:guid

davewalter:
#334

## What is this change about?
This PR implements the `DELETE /v3/apps/:guid` endpoint for the `cf delete` command.

No

## Acceptance Steps
See #334

@Birdrock

### DIFF
--- a/api/apis/app_handler.go
+++ b/api/apis/app_handler.go
@@ -31,6 +31,7 @@ const (
 	AppStartEndpoint             = "/v3/apps/{guid}/actions/start"
 	AppStopEndpoint              = "/v3/apps/{guid}/actions/stop"
 	AppRestartEndpoint           = "/v3/apps/{guid}/actions/restart"
+	AppDeleteEndpoint            = "/v3/apps/{guid}"
 	invalidDropletMsg            = "Unable to assign current droplet. Ensure the droplet exists and belongs to this app."
 
 	AppStartedState = "STARTED"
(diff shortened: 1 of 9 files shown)


api/apis/app_handler.go
```diff
@@ -659,6 +661,48 @@ func (h *AppHandler) appRestartHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func (h *AppHandler) appDeleteHandler(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	vars := mux.Vars(r)
+	appGUID := vars["guid"]
+
+	authInfo, ok := authorization.InfoFromContext(ctx)
```
tcdowney, line 669: Nice, glad to see we're using this in new endpoints 🙂

api/repositories/app_repository.go
```diff
@@ -363,6 +368,17 @@ func (f *AppRepo) SetAppDesiredState(ctx context.Context, authInfo authorization
...
 
+func (f *AppRepo) DeleteApp(ctx context.Context, authInfo authorization.Info, message AppDeleteMessage) error {
+	cfApp := &workloadsv1alpha1.CFApp{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      message.AppGUID,
+			Namespace: message.SpaceGUID,
+		},
+	}
+
+	return f.privilegedClient.Delete(ctx, cfApp)
```
tcdowney, line 379: Would it work still if we used the user client like we do in AppList?

https://github.com/cloudfoundry/cf-k8s-controllers/pull/360/files#diff-433641cf784e6cc94960970a9af8ae3a9c392337e21bc68d90485dd5c9a31b10R200

tcdowney (approved):
Overall looks good to me. There does look like an opportunity here to use a user client instead of the privileged client to perform the delete, but I'd understand if you'd rather tackle that more holistically.

Eventually I think we want to implement authz as part of building out a new endpoint, but we haven't really discussed that much. I was just inspired by the AppRepo constructor. 😛

akrishna90 (commented):
@Birdrock - update the api docs!

api/repositories/app_repository.go
davewalter: @tcdowney It works with some minor RBAC tweaks.

Issue #291, opened 2021-11-25
[Chore] Enable the dupl linter


mnitchev:
We recently enabled golangci-lint, but only use the default linters. The dupl linter checks for duplication in our code. Since we ran it and found that there are a few places where we have a lot of duplications, we feel that this might be a useful thing to check. [Link to the linter](https://github.com/mibk/dupl

danail-branekov:
I tried enabling the dupl linter and got lots of warnings. After looking at them, I could classify duplications like this:

## API duplications

### Rest handlers boilerplate
Those code duplications are caused by the huge amount of boilerplate around request handling, e.g.:
1. Get the authInfo from the request + its error handling
2. Talk to the underlying repository + related error handling
3. send the result from the repo to the presenter + related error handling


### Repositories duplications
Repositories implement similar patterns:
* They all have some sort of `Get` and `List` methods
* Those methods always talk to the k8s client by calling its `Get` or `List`
* Format the result from the k8s client and return it

## Controllers duplications
Those duplications are mostly around subnamespaces which makes sense. Orgs and spaces are backed up by the same subnamespace anchor, hence the orgs implementation is quite similar to the spaces one. 

## Error case tests
This might be a natural consequence of the duplications listed above. Generally, similar handlers code requires similar test code, hence the duplication


In order to address the duplications we could:
* For rest handlers we could extract an abstract rest handler that gets the authinfo from the context and use template methods (or even strategies) to proceed with the thing the handler is expected to do. For example, a `GetXXX` handler could have functions injected during wiring one talking to the repo and another talking to the presenter
* For controllers, I guess introducing an org/space custom resources would make the controllers duplications similar to the rest api handlers ones, we could tackle them in a similar manner
* For the errors tests - I guess the actions above would solve most of them

**However**, I do believe that the refactoring above would make the code quite unreadable (and maybe hard to maintain) so I would probably limit the refactoring down to just extracting the authinfo in a wrapper. For the rest of the duplications, we could probably turn the linter off. But if we turn the linter off for those packages, does it make sense to enable it all?

PR #361, opened 2021-12-14
Extract authInfo handling to reduce code duplication

danail-branekov:
https://github.com/cloudfoundry/cf-k8s-controllers/issues/291

## What is this change about?
Extract handling the authorization info in API handlers into a dedicated wrapper. This has the consequence that the duplicated code (+related test code) around that is gone. There is sort of a drawback though - handlers' functions no longer comply to the `http.HandlerFunc` interface as they have the additional `authInfo` parameter. 

To be honest, I can't really decide whether that change is worth it so I am looking forwards comments, suggestions and opinions.

No

## Acceptance Steps
Run the tests - they should pass


matt-royal (approved):
Looks good to me

matt-royal:
Merged this now in the interest of avoiding future merge conflicts

Issue #335, opened 2021-12-07
[Feature]: API Client can hit _MINIMAL_ `GET /v3/jobs/:guid` for app deletion jobs


tcdowney:
### Blockers/Dependencies

Related to https://github.com/cloudfoundry/cf-k8s-controllers/issues/334.

The `DELETE /v3/apps/:guid` endpoint deletes apps asynchronously and returns a job for clients to poll.

### Background

**As a** client of the API Shim
**I want** the `GET /v3/jobs/:guid` endpoint to support `app.delete` jobs
**So that** can delete apps without having to change my client code

#### Happy Path
**GIVEN** I am requesting an `app.delete` Job
**WHEN I** make the following request

```
curl "https://api.example.org/v3/jobs/app.delete-<app-guid>" \
  -X GET \
  -H "Authorization: bearer [token]"
```
**THEN I** see the following response

```json
{
  "guid": "app.delete-<app-guid>",
  "created_at": "2021-12-07T22:38:33Z",
  "updated_at": "2021-12-07T22:38:33Z",
  "operation": "app.delete",
  "state": "COMPLETE",
  "errors": [

  ],
  "warnings": [

  ],
  "links": {
    "self": {
      "href": "https://api.plain-tail.capi.land/v3/jobs/app.delete-<app-guid>"
    }
  }
}
```

---

#### Unhappy Path

Investigate and propose a solution for when/if we should return a `404 Job not found`.

### Dev Notes

Should we make the endpoint try and see if all traces of the app are gone and return the state as `PROCESSING` if it's still deleting?

Consider creating a followup `tech-debt` issue with a more robust implementation if you've got ideas to improve this.

Job docs: https://v3-apidocs.cloudfoundry.org/version/3.111.0/index.html#the-job-object

PR #363, opened 2021-12-14
Implement minimal GET app delete job

acosta11:
[#335](https://github.com/cloudfoundry/cf-k8s-controllers/issues/335)

## What is this change about?
Add a minimal implementation of the GET jobs endpoint for app delete jobs.

No

## Acceptance Steps
See related issue.

paired with @Birdrock

### DIFF
--- a/api/apis/job_handler.go
+++ b/api/apis/job_handler.go
@@ -1,6 +1,7 @@
 package apis
 
 import (
+	"encoding/json"
 	"net/http"
 	"net/url"
 	"strings"
@@ -14,6 +15,7 @@ import (
 const (
 	JobGetEndpoint  = "/v3/jobs/{guid}"
 	syncSpacePrefix = "sync-space.apply_manifest-"
+	appDeletePrefix = "app.delete-"
 )
 
 type JobHandler struct {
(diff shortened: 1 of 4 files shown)


api/apis/job_handler.go
```diff
@@ -34,26 +36,27 @@ func (h *JobHandler) jobGetHandler(w http.ResponseWriter, r *http.Request) {
...
 	}
-}
 
-func getSpaceGUID(jobGUID string) string {
-	if strings.HasPrefix(jobGUID, syncSpacePrefix) {
-		spaceGUID := strings.Replace(jobGUID, syncSpacePrefix, "", 1)
-		return spaceGUID
-	}
-	return ""
+	w.WriteHeader(http.StatusOK)
```
davewalter, line 58: I think we normally omit this as this is the default status code.

davewalter (approved):
Looks good. Just the comment about the explicit 200 OK response code.

api/apis/job_handler.go
acosta11: Good call I think I missed this when rebasing on main along with some other artifacts of letting the branch languish over the weekend. Will remove.

matt-royal:
Since it seemed like the StatusOK line was the only issue, @julian-hj and I went ahead and removed it and force pushed over the existing branch. We'll merge it once all the tests pass

Issue #362, opened 2021-12-14
[Bug]: Hitting the logging endpoints returns 401 errors


tcdowney:
### What happened?

We added some minimal "log-cache"-like endpoints in https://github.com/cloudfoundry/cf-k8s-controllers/issues/222 that always return an empty set of log envelopes. We don't ever want to enforce auth(n/z) on the `/api/v1/info` discovery endpoint and we don't currently care to enforce any auth(n/z) on the log envelope endpoint (`/api/v1/read/<app-guid>`) since it always returns a hard-coded empty list.

Currently when the CLI tries to hit both of these endpoints it gets back a `401` error due to the [`AuthenticationMiddleware`](https://github.com/cloudfoundry/cf-k8s-controllers/blob/95e7a474d0474b2d8b64a73bfe267f45ffb80ab7/api/apis/authentication_middleware.go#L24) that was recently added.

### What you expected to happen

Both the `/api/v1/info` and `/api/v1/read/<app-guid>` endpoints should successfully return their hard-coded responses with or without valid auth.

### Acceptance Criteria

---

### `/api/v1/info`
GIVEN I have the project deployed
WHEN I make the following request
```
curl https://example.com/api/v1/info
```
THEN I get back the following response (and not a `401`)
```json
{"version":"2.11.4+cf-k8s","vm_uptime":"0"}
```

---

### `/api/v1/read/<app-guid>`
GIVEN I have the project deployed
WHEN I make the following request
```
curl https://example.com/api/v1/read/<app-guid>
```
THEN I get back the following hard-coded response (and not a `401`)
```json
{"envelopes":{"batch":[]}}
```

### How to reproduce it (as minimally and precisely as possible)

Perform a `cf push` and you will see a number of calls to log-cache fail with `401` errors.

### Anything else we need to know?

TBD, I am planning on logging two issues:
1. Investigate why the `AuthenticationMiddleware` doesn't like the requests that the CLI is making (I _think_ it is sending some form of `bearer` token still)
2. Update the "basic app logs" story to mention that `/api/v1/read/<app-guid>` need auth(n/z) added back to it.

### Environment

Revision of codebase: Latest (as of logging this issue it is 95e7a474d0474b2d8b64a73bfe267f45ffb80ab7)
Kubernetes version (use `kubectl version`): N/A
Cloud provider or hardware configuration: N/A
Others: N/A

tcdowney:
[...]

PR #364, opened 2021-12-14
Skip AuthenticationMiddleware for log-cache endpoints

tcdowney:
* https://github.com/cloudfoundry/cf-k8s-controllers/issues/362

## What is this change about?
Right now the CF CLI is omitting the authorization header for the log-cache endpoints for some reason. I've got an open Slack thread about that and will link the issue for it once it exists.

As a mitigation this change omits our hard-coded log-cache endpoints in the `AuthenticationMiddleware`.
- `/api/v1/info` is a discovery endpoint and doesn't require
authentication in general
- `/api/v1/read/<app-guid>` eventually will require authentication, but
doesn't right now since we are returning a hard-coded empty list of
envelopes. Eventually we will want Auth here.

No

## Acceptance Steps
Follow the A/C on https://github.com/cloudfoundry/cf-k8s-controllers/issues/362 or try to `cf push` an app. You shouldn't see `401` errors anymore while trying to hit log-cache.

### DIFF
--- a/api/apis/authentication_middleware.go
+++ b/api/apis/authentication_middleware.go
@@ -2,6 +2,7 @@ package apis
 
 import (
 	"net/http"
+	"regexp"
 
 	"code.cloudfoundry.org/cf-k8s-controllers/api/authorization"
 	"github.com/go-http-utils/headers"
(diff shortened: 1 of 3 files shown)


api/apis/authentication_middleware.go
```diff
@@ -74,3 +76,14 @@ func (a *AuthenticationMiddleware) Middleware(next http.Handler) http.Handler {
 		next.ServeHTTP(w, r)
 	})
 }
+
+// Eventually we will want to authenticate the log-cache read endpoint and go back to using the simple
+// unauthenticatedEndpoints map. For now though we need to do a more complicated regexp match against
+// the path since the log-cache read endpoint includes an arbitrary guid as part of its path
+var logCacheReadEndpointRegexp = regexp.MustCompile(`\/api\/v1\/read\/[0-9a-fA-F\-]*$`)
```
tcdowney, line 83: I'm open to suggestions here around other ways of doing this. The `unauthenticatedEndpoints` map didn't work in this case because `r.URL.Path` isn't a static path for this endpoint, but I wanted to preserve that map because I think we can go back to using it and it alone once we fix the CLI and can require authentication for this endpoint.

tcdowney: I like the simplicity / performance of the map versus a composite regexp that could match all of these endpoints.

Birdrock: I think this is fine as a temporary measure.

Birdrock (changes requested):
Looks like there are some go mod issues.

tcdowney:
Thanks @Birdrock , I threw away the `go.mod` changes. They must have came about when I was trying to get `golangci-lint` working on my machine.

davewalter (approved):

Issue #337, opened 2021-12-07
[Feature]: Deleting a `CFApp` removes any `CFRoute` destinations that refer to the app


tcdowney:
### Background

**As an** App Developer
**I want** Any route destinations that refer to my app to be removed when I delete the app
**So that** Traffic isn't misrouted to an app that no longer exists

### Acceptance Criteria

**GIVEN** I have `CFRoute`s with destinations that refers to my app
**WHEN I** run `cf delete MY_APP`
**THEN I** see that the `CFRoutes` are updated to no longer have a destination that points at my app

### Dev Notes

* Could have a finalizer on the `CFApp` that uses some of this logic to find Routes and update them: https://github.com/cloudfoundry/cf-k8s-controllers/blob/3e602c6042ab48fb01853e7089ab114de80b7396/api/apis/app_handler.go#L521
* Could have the `RouteController` watch for app deletions
* The App reconcile logic is only responsible for updating the CFRoute, not the underlying resources realizing that resource's intent

tcdowney:
Closing: This was addressed by https://github.com/cloudfoundry/cf-k8s-controllers/pull/365.

PR #365, opened 2021-12-15
Deleting CFApp removes CFRoute Destinations that match the app

akrishna90:
#337 

## What is this change about?
Deleting a CFApp deletes matching Destinations for the app on CFRoute

No.

## Acceptance Steps
follow acceptance on #337 

@acosta11

### DIFF
--- a/controllers/controllers/shared/index.go
+++ b/controllers/controllers/shared/index.go
@@ -1,0 +1,30 @@
+package shared
+
+import (
+	"context"
+
+	networkingv1alpha1 "code.cloudfoundry.org/cf-k8s-controllers/controllers/apis/networking/v1alpha1"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+const DestinationAppName = "destinationAppName"
+
+func SetupIndexWithManager(mgr manager.Manager) error {
+
+	// Generate indexes for CFRoute on field spec.Destination.AppRef.Name for efficient querying.
+	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &networkingv1alpha1.CFRoute{}, DestinationAppName,
+		func(rawObj client.Object) []string {
+			route := rawObj.(*networkingv1alpha1.CFRoute)
+			var destinationAppNames []string
+			for _, destination := range route.Spec.Destinations {
+				destinationAppNames = append(destinationAppNames, destination.AppRef.Name)
+			}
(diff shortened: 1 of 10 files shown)


controllers/controllers/workloads/cfapp_controller.go
```diff
@@ -181,6 +191,82 @@ func getDesiredInstanceCount(processType string) int {
...
+					cfRoute.Spec.Destinations = cfRoute.Spec.Destinations[:len(cfRoute.Spec.Destinations)-1]
+				}
+			}
+		}
+		err = r.Client.Patch(ctx, &cfRoute, client.MergeFrom(originalCFRoute))
+		if err != nil {
+			r.Log.Error(err, "failed to patch cfRoute to remove a destination")
+			return ctrl.Result{}, err
+		}
+	}
```
tcdowney, line 244: I think it would improve readability and make it more clear what's going on here to move this into another function called like `removeRouteDestinations` or something

controllers/controllers/shared/index.go
```diff
@@ -0,0 +1,30 @@
...
+
+	// Generate indexes for CFRoute on field spec.Destination.AppRef.Name for efficient querying.
+	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &networkingv1alpha1.CFRoute{}, DestinationAppName,
+		func(rawObj client.Object) []string {
+			route := rawObj.(*networkingv1alpha1.CFRoute)
+			var destinationAppNames []string
+			for _, destination := range route.Spec.Destinations {
+				destinationAppNames = append(destinationAppNames, destination.AppRef.Name)
+			}
+			return destinationAppNames
```
tcdowney, line 24: Out of curiosity, how does this work with an array of strings?

tcdowney (approved):
Overall LGTM. Got one minor suggestion and just a general question about how the index works.

controllers/controllers/shared/index.go
akrishna90: @tcdowney - in case of array, the field selector tries to match with at-least one value in the array. 
more info and example - https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/client#FieldIndexer

tcdowney: Awesome! Learned something new 😁

controllers/controllers/workloads/cfapp_controller.go
akrishna90: Thanks for the review. I have addressed this

PR #384, opened 2021-12-21
Add dev support for local container registry

julian-hj:
Nope

## What is this change about?
It adds a local container registry option to the deploy-on-kind developer script.
- `deploy-on-kind.sh` can now take a second argument that tells it to
  deploy and use a local container registry.
- this requires some hackery to get the registry address to work from
  within the kind node, but saves us from having to specify an external
  registry.

no

## Acceptance Steps
To use it:

```
./scripts/deploy-on-kind.sh cluster local
```

The existence of the second argument tells the script to wire up the local registry.

Once the script is done, create a CFDomain object, and your cluster will be ready to push and run apps.

@Birdrock @tcdowney @PureMunky PTAL

### DIFF
--- a/controllers/config/overlays/kind/controllersconfig/cf_k8s_controllers_config.yaml
+++ b/controllers/config/overlays/kind/controllersconfig/cf_k8s_controllers_config.yaml
@@ -1,5 +1,4 @@
 kpackImageTag: gcr.io/cf-relint-greengrass/cf-k8s-controllers/kpack/beta
-
 clusterBuilderName: cf-kpack-cluster-builder
 cfProcessDefaults:
   memoryMB: 500
--- a/scripts/deploy-on-kind.sh
+++ b/scripts/deploy-on-kind.sh
@@ -48,6 +48,7 @@ EOF
 }
 
 ensure_local_registry() {
+  helm repo add twuni https://helm.twun.io
   helm upgrade --install localregistry twuni/docker-registry --set service.type=NodePort,service.nodePort=30050,service.port=30050
 
   # TODO-maybe we don't need to add the /etc/hosts hack if we configure the mirror below to redirect to 127.0.0.1?
--- a/scripts/deploy-on-kind.sh
+++ b/scripts/deploy-on-kind.sh
@@ -77,7 +77,7 @@ EOF
 deploy_cf_k8s_controllers() {
   pushd $ROOT_DIR >/dev/null
   {
-    if [ -n "$use_local_registry" ]; then
+    if [[ -n "$use_local_registry" ]]; then
       export DOCKER_SERVER="localregistry-docker-registry.default.svc.cluster.local:30050"
       export DOCKER_USERNAME="whatevs"
       export DOCKER_PASSWORD="whatevs"
@@ -94,7 +94,7 @@ deploy_cf_k8s_controllers() {
     fi
     kind load docker-image --name "$cluster" "$IMG_CONTROLLERS"
     make install-crds
-    if [ -n "$use_local_registry" ]; then
+    if [[ -n "$use_local_registry" ]]; then
       make deploy-controllers-kind-local
     else
       make deploy-controllers-kind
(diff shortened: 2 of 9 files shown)


controllers/config/overlays/kind/controllersconfig/cf_k8s_controllers_config.yaml
```diff
@@ -1,4 +1,5 @@
 kpackImageTag: gcr.io/cf-relint-greengrass/cf-k8s-controllers/kpack/beta
+
```
tcdowney, line 2: Is this space intentional?

scripts/deploy-on-kind.sh
```diff
@@ -75,13 +111,21 @@ deploy_cf_k8s_api() {
       make docker-build-api
     fi
     kind load docker-image --name "$cluster" "$IMG_API"
-    make deploy-api-kind-auth
+    if [ -n "$use_local_registry" ]; then
+      make deploy-api-kind-local
+    else
+      make deploy-api-kind-auth
```
tcdowney, line 117: I'm wondering if we should have the `authEnabled: true` bit from `make deploy-api-kind-auth` in the API config that `make deploy-api-kind-local` uses. That's the feature flag we're using for whether or not endpoints should use the user's provided auth token when interacting with the K8s API.

https://github.com/cloudfoundry/cf-k8s-controllers/blob/main/api/config/overlays/kind-auth-enabled/apiconfig/cf_k8s_api_config.yaml#L11

tcdowney (approved):
Overall LGTM. Added a couple comments.

scripts/deploy-on-kind.sh
matt-royal: We just tried to set `authEnabled: true` for the local kind cluster, and we ran into issues creating orgs:

```
$ cf create-org o
Creating org o as kubernetes-admin...
OK

TIP: Use 'cf target -o "o"' to target new org

$ cf create-space -o o s
Organization 'o' not found.
FAILED
```

Any idea what's going on? The SubnamespaceAnchor was created for the org, but for whatever reason we can't see it as our user.

matt-royal: We're thinking this is a bug in the current implementation of org creation and listing. The code appears to create the org as a privileged client, but the fetch it as the actual user. Still, it seems like it should work since our user is an admin, but it isn't 🤷

controllers/config/overlays/kind/controllersconfig/cf_k8s_controllers_config.yaml
julian-hj: definitely not..just a leftover crumb from previous hacking. I took it back out.

scripts/deploy-on-kind.sh
danail-branekov: > Any idea what's going on?

No, but I remember having this issue back in time but was in a hurry and forgot to create an issue. I worked it around via 

```
create create-org o
cf target -o o
cf create-space s
cf target -o o -s s
```

No idea why that worked though...

scripts/deploy-on-kind.sh
```diff
@@ -47,9 +47,42 @@ EOF
   kind export kubeconfig --name "$cluster" --kubeconfig "$HOME/.kube/$cluster.yml"
 }
 
+ensure_local_registry() {
+  helm repo add twuni https://helm.twun.io
+  helm upgrade --install localregistry twuni/docker-registry --set service.type=NodePort,service.nodePort=30050,service.port=30050
+
+  # TODO-maybe we don't need to add the /etc/hosts hack if we configure the mirror below to redirect to 127.0.0.1?
```
danail-branekov, line 54: Instead of adding TODOs in the code I would suggest having a chore to look into configuring the mirror

scripts/deploy-on-kind.sh
julian-hj: @danail-branekov We found that we consistently hit the issue with the option enabled, and didn't see it without, even when we used the four steps above.

Maybe there is some kind of caching of org lists in the client? We noticed that the org is created by the privileged user and then retrieved by the not-privileged user in the case where the option is enabled. I also saw that an old org previously created *did* show up in `cf orgs` but a new one did not.

julian-hj: @danail-branekov I filed #392 to describe the `cf orgs` bug. Oddly, it appears to affect any org without spaces in it, but once there is a space in the org, it becomes visible/usable.

Issue #331, opened 2021-12-06
[Feature]: App Dev can see `PORT` related env vars in their app's environment


tcdowney:
### Blockers

We should do https://github.com/cloudfoundry/cf-k8s-controllers/issues/376 first so that we can directly reference the `CFRoute` status.

### Background

Currently many apps deployed to CF look for the `VCAP_APP_HOST`, `VCAP_APP_PORT`, and `PORT` env vars to determine what host / port to listen on. We're not setting these right now and all `CFRoutes` are defaulting to port `8080` even if the app container is exposing some other port.

A bit more info for these env vars:

* `VCAP_APP_HOST` - this is always hardcoded in CF to be `0.0.0.0`. I think this should be the same for K8s.
* `VCAP_APP_PORT` - this is set to the first port that is returned by the [ProcessModel#open_ports function](https://github.com/cloudfoundry/cloud_controller_ng/blob/ed09b45936879975e9784927431cc5387e1221d4/app/models/runtime/process_model.rb#L566-L584)
* `PORT` - this is the same as `VCAP_APP_PORT`

### Acceptance Criteria

**As an** App Developer
**I want** the system to be able to project the `VCAP_APP_HOST`, `VCAP_APP_PORT`, and `PORT` env vars into my app
**So that** can use my existing CF env var aware apps with modification

---

### `CFRoute` exists and has port info in its `Status`
**GIVEN** I have a `CFRoute` that has a destination for the App/`CFProcess` in its `Status`
**WHEN I** inspect the `EiriniLRP` for the `CFProcess` and the running `Pod`
**THEN I** see that `VCAP_APP_HOST` is set to `0.0.0.0` and `VCAP_APP_PORT`/`PORT` are set to a value that matches the destination port in the `CFRoute`'s `Status`

---

### `CFRoute` does not exists or otherwise lacks port info in its `Status`
**GIVEN** I do not have a `CFRoute` with the `CFProcess` in its destinations
**WHEN I** inspect the `EiriniLRP` for the `CFProcess` and the running `Pod`
**THEN I** see that `VCAP_APP_HOST` is set to `0.0.0.0` and `VCAP_APP_PORT`/`PORT` are set to `8080` by default

### Dev Notes

---

[...]

PR #390, opened 2021-12-23
Expose PORT, VCAP_APP_PORT, VCAP_APP_HOST env to app

julian-hj:
#331

## What is this change about?

Previously, we were not setting the PORT, VCAP_APP_PORT, VCAP_APP_HOST environment variables when creating an LRP, leaving some apps to choose their own port, which didn't work.

This change adds the standard environment variables.
If there is an existing route with a destination that matches our app
and a port set, we use that port. Otherwise we use port 8080

No

## Acceptance Steps

See #331 

@acosta11

### DIFF
--- a/controllers/controllers/workloads/cfprocess_controller.go
+++ b/controllers/controllers/workloads/cfprocess_controller.go
@@ -21,6 +21,7 @@ import (
 	"crypto/sha1"
 	"errors"
 	"fmt"
+	"sort"
 	"strconv"
 
 	networkingv1alpha1 "code.cloudfoundry.org/cf-k8s-controllers/controllers/apis/networking/v1alpha1"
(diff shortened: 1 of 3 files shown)


controllers/controllers/workloads/cfprocess_controller.go
```diff
@@ -240,11 +248,37 @@ func (r *CFProcessReconciler) fetchLRPsForProcess(ctx context.Context, cfProcess
...
+	err := r.Client.List(ctx, &cfRoutesForProcess, client.InNamespace(cfApp.GetNamespace()), client.MatchingFields{shared.DestinationAppName: cfApp.Name})
+	if err != nil {
+		return 0, err
+	}
+	// Filter those destinations
+	for _, cfRoute := range cfRoutesForProcess.Items {
+		for _, destination := range cfRoute.Status.Destinations {
+			if destination.ProcessType == cfProcess.Spec.ProcessType && destination.Port != 0 {
+				// Just use the first candidate port
+				return destination.Port, nil
```
tcdowney, line 263: Do you know if the order of these destinations is deterministic?

tcdowney: I think if we include `AppGuid` in the `if` check there should only be a single match here.

controllers/controllers/workloads/cfprocess_controller.go
```diff
@@ -240,11 +248,37 @@ func (r *CFProcessReconciler) fetchLRPsForProcess(ctx context.Context, cfProcess
...
-func secretDataToEnvMap(secretData map[string][]byte) map[string]string {
+func (r *CFProcessReconciler) getPort(ctx context.Context, cfProcess workloadsv1alpha1.CFProcess, cfApp workloadsv1alpha1.CFApp) (int, error) {
+	// Get Routes for the process
+	var cfRoutesForProcess networkingv1alpha1.CFRouteList
+	err := r.Client.List(ctx, &cfRoutesForProcess, client.InNamespace(cfApp.GetNamespace()), client.MatchingFields{shared.DestinationAppName: cfApp.Name})
+	if err != nil {
+		return 0, err
+	}
+	// Filter those destinations
+	for _, cfRoute := range cfRoutesForProcess.Items {
```
tcdowney, line 259: Do you know if the order of these routes deterministic?

matt-royal (approved):
lgtm

tcdowney (approved):
lgtm

controllers/controllers/workloads/cfprocess_controller.go
tcdowney: Looks good now that we're sorting by creation timestamp. 👍

Issue #374, opened 2021-12-17
[Bug]: golang apps fail to start without a Procfile


julian-hj:
### What happened?

With a deployed cf-k8s installation:
- cd ~/workspace/cf-acceptance-tests/assets/golang
- cf push go

```
Instances starting...

name:              go
requested state:   started
routes:
last uploaded:     Thu 16 Dec 16:17:44 PST 2021
stack:             io.buildpacks.stacks.bionic
buildpacks:

type:            golang
sidecars:
instances:       0/0
memory usage:    500M
start command:   /layers/tanzu-buildpacks_go-build/targets/bin/golang
There are no running instances of this process.
```
If we add a `Procfile` with `web: golang` in it, we get a different bad behavior:

```
Waiting for app go to start...

Instances starting...
Error unmarshalling the following into a cloud controller error: upstream connect error or disconnect/reset before headers. reset reason: connection termination
FAILED
```
The app *does* start up, but it ends up creating 2 CFProcesses. One is type `web` and actually runs. The other is type `golang` and has 0 instances.

### What you expected to happen

We pushed the same app (with no Procfile) to a cf-for-vms environment, and it succeeded. That app ended up with a single process named `web` that also had the correct command (`./bin/golang`)

We either need for the buildpack to make a `web` process for us, or we need to handle the single process that isn't named `web` and use that instead.

### Acceptance Criteria

GIVEN a simple golang app with no Procfile
WHEN I cf push
THEN I get a running app and no errors

### How to reproduce it (as minimally and precisely as possible)

see "what happened" above

### Anything else we need to know?

We think this is because the go buildpack isn't making a process with `processType: web` by default, and if it isn't there, we don't find the other (only) process and use that.

We could try to get the buildpack folk to give us a `web` process, or we could change our controller to create a web process by default when there is only one process in the container.

### Environment

Revision of codebase: 0.2.0ish
Kubernetes version (use `kubectl version`): 1.21
Cloud provider or hardware configuration: TCE on AWS, but I don't think it matters
Others:

tcdowney:
@julian-hj what happens if you call `cf scale APP_NAME --process --instances 1`? Does that process type start then?

[...]

PR #393, opened 2021-12-23
Ensure there is always a "web" process

matt-royal:
#374

## What is this change about?

This change ensures that a `web` process is always present, even when none is created as part of staging. This solves problems for go apps, which name their processes after the package being compiled.

This is a behavior change for apps of some languages, such as go, but it brings the behavior in line with cf-for-vms and cf-for-k8s so I wouldn't consider it breaking. Apps that worked before this change should work after it, and some apps that didn't work should work now.

## Acceptance Steps

Follow the steps in the issue


julian-hj (approved):
LGTM

Issue #391, opened 2021-12-23
[Bug]: Re-pushing a running app fails because the app doesn't restart


julian-hj:
### What happened?

`$ cf push ruby2`

\<app starts nicely and is running>
\<make some trivial change to the app>

```
$ cf push ruby2

...

Waiting for app ruby2 to start...

Unexpected Response
Response Code: 504
Code: 0, Title: , Detail: upstream request timeout
FAILED
```

This seems to just be an issue with desired state. If I start the app manually, it starts back up again just fine:

```
$ cf start ruby2
Starting app ruby2 in org o / space s as kubernetes-admin...

Waiting for app to start...

Instances starting...

name:              ruby2
requested state:   started
routes:            ruby2.vcap.me
last uploaded:     Thu 23 Dec 11:11:56 PST 2021
stack:             io.buildpacks.stacks.bionic
buildpacks:

type:           web
sidecars:
instances:      1/1
memory usage:   500M
     state     since                  cpu    memory   disk     details
#0   running   2021-12-23T19:23:40Z   0.0%   0 of 0   0 of 0
```

### What you expected to happen

The app should restart when I `cf push`

### Acceptance Criteria

GIVEN a running app
WHEN I make a change and re-push it
THEN I see that it restages and restarts gracefully

### How to reproduce it (as minimally and precisely as possible)

see "What Happened" above

### Environment

Revision of codebase: main branch circa 12/23/2021
Kubernetes version (use `kubectl version`): 1.21
Cloud provider or hardware configuration: kind
Others:

Issue #260, opened 2021-11-17
[Feature]: App developer can restart an application w/ downtime (Controller implementation)


gnovv:
### Blockers/Dependencies

Follow up to https://github.com/cloudfoundry/cf-k8s-controllers/issues/170

### Background

As an App developer
I want to be able to issue a restart command
So that restart my workload

Previously we implemented V3 restart app using the api shim and watching the app's pods in order to synchronously set the app's desiredState to started.

Now we want to implement this functionality on the Controller side so that it is asynchronous and more robust.

### Acceptance Criteria

**GIVEN** my app exists
**WHEN I** issue a start command
```
curl "https://api.example.org/v3/apps/<guid>/actions/restart" \
  -X POST \
  -H "Authorization: bearer <token>"
```

**THEN I** get a response similar to the following
```
HTTP/1.1 200 OK
Content-Type: application/json

{
  "guid": "1cb006ee-fb05-47e1-b541-c34179ddc446",
  "name": "my_app",
  "state": "STARTED",
  "created_at": "2016-03-17T21:41:30Z",
  "updated_at": "2016-03-18T11:32:30Z",
  "lifecycle": {
    "type": "buildpack",
    "data": {
      "buildpacks": ["java_buildpack"],
      "stack": "cflinuxfs3"
    }
  },
  "relationships": {
    "space": {
      "data": {
        "guid": "2f35885d-0c9d-4423-83ad-fd05066f8576"
      }
    }
  },
  "links": {
    "self": {
      "href": "https://api.example.org/v3/apps/1cb006ee-fb05-47e1-b541-c34179ddc446"
    },
    "space": {
      "href": "https://api.example.org/v3/spaces/2f35885d-0c9d-4423-83ad-fd05066f8576"
    },
    "processes": {
      "href": "https://api.example.org/v3/apps/1cb006ee-fb05-47e1-b541-c34179ddc446/processes"
    },
    "packages": {
      "href": "https://api.example.org/v3/apps/1cb006ee-fb05-47e1-b541-c34179ddc446/packages"
    },
    "environment_variables": {
      "href": "https://api.example.org/v3/apps/1cb006ee-fb05-47e1-b541-c34179ddc446/environment_variables"
    },
    "current_droplet": {
      "href": "https://api.example.org/v3/apps/1cb006ee-fb05-47e1-b541-c34179ddc446/droplets/current"
    },
    "droplets": {
[...]

PR #396, opened 2022-01-04
Re-pushing a running app fails - does not restart

Birdrock:
#391 

## What is this change about?
Fixes subsequent `cf push` behavior. Was hanging on watching for pod deletion. This updates the implementation as a result of changes from #260.

No.

## Acceptance Steps
Follow steps in #391

@akrishna90


akrishna90 (approved):
paired with @Birdrock, lgtm!

PR #397, opened 2022-01-04
Resolve TODO in deploy-on-kind.sh

julian-hj:
- reduce the number of hacks we use for local registry deployment by 1
- reduce "sleep" time after patching kind containerd instance from 30s
  to 10s.

No

## What is this change about?

Small tweaks to improve the deploy-on-kind.sh script when using a local container registry

No

## Acceptance Steps

```
./scripts/deploy-on-kind.sh xyz local
k apply -f controllers/config/samples/cfdomain.yaml
cd ../cf-acceptance-tests/assets/node
cf p node
curl node.vcap.me
```

### DIFF
--- a/scripts/deploy-on-kind.sh
+++ b/scripts/deploy-on-kind.sh
@@ -51,9 +51,6 @@ ensure_local_registry() {
   helm repo add twuni https://helm.twun.io
   helm upgrade --install localregistry twuni/docker-registry --set service.type=NodePort,service.nodePort=30050,service.port=30050
 
-  # TODO-maybe we don't need to add the /etc/hosts hack if we configure the mirror below to redirect to 127.0.0.1?
-  docker exec "${cluster}-control-plane" bash -c 'echo "127.0.0.1 localregistry-docker-registry.default.svc.cluster.local" >> /etc/hosts'
-
   # reconfigure containerd to allow insecure connection to our local registry
   docker cp ${cluster}-control-plane:/etc/containerd/config.toml /tmp/config.toml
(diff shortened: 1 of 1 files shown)


scripts/deploy-on-kind.sh
```diff
@@ -62,15 +59,15 @@ ensure_local_registry() {
 [plugins."io.containerd.grpc.v1.cri".registry]
   [plugins."io.containerd.grpc.v1.cri".registry.mirrors]
     [plugins."io.containerd.grpc.v1.cri".registry.mirrors."localregistry-docker-registry.default.svc.cluster.local:30050"]
-      endpoint = ["http://localregistry-docker-registry.default.svc.cluster.local:30050"]
+      endpoint = ["http://127.0.0.1:30050"]
```
tcdowney, line 62: Out of curiosity how does the networking work for this? Maybe I just need to read up more on KinD networking.. 😅

tcdowney (approved):
LGTM

scripts/deploy-on-kind.sh
julian-hj: 1. We set the port to 30050 so that we can map it to a node port with the same port number
2. We tell the helm chart that we want a nodeport with 30050
3. We tell containerd to mirror the k8s dns for the service to localhost (we used to just put the hostname in /etc/hosts but this PR makes that unnecessary)
4. we tell everybody to use localregistry-docker-registry.default.svc.cluster.local:30050. For the build service, kube dns resolves the address. For k8s itself, the containerd config above redirects it to the nodeport we opened.

This approach assumes that we will just have a single kind worker node, but it could be elaborated if we wanted to use multi-node clusters.

Issue #314, opened 2021-12-01
[Chore]: CFAppValidatingWebhook test doesn't get an error sometimes


georgethebeatle:
Running `CONTROLLERS_GINKGO_NODES=8 ./scripts/run-tests.sh controllers/webhooks/workloads/integration` eventually fails with an error like the one below.

**Note:** After fixing this flake, consider removing the [parallel nodes limitation](https://github.com/cloudfoundry/cf-k8s-controllers/blob/3f660c4515dd8c2382d99058b5a5e207108c9c1b/Makefile#L9) for controllers tests.

```
• [FAILED] [0.035 seconds]
CFAppValidatingWebhook
/home/georgethebeatle/workspace/cf-k8s-controllers/controllers/webhooks/workloads/integration/cfapp_validation_integration_test.go:16
  creating a new CFApp record
  /home/georgethebeatle/workspace/cf-k8s-controllers/controllers/webhooks/workloads/integration/cfapp_validation_integration_test.go:36
    another CFApp exists with a same name in the same namespace
    /home/georgethebeatle/workspace/cf-k8s-controllers/controllers/webhooks/workloads/integration/cfapp_validation_integration_test.go:94
      [It] should fail
      /home/georgethebeatle/workspace/cf-k8s-controllers/controllers/webhooks/workloads/integration/cfapp_validation_integration_test.go:105

  Begin Captured GinkgoWriter Output >>
    2021-12-01T10:45:48.615Z    DEBUG   controller-runtime.webhook.webhooks     received request        {"webhook": "/mutate-workloads-cloudfoundry-org-v1alpha1-cfapp", "UID": "ba4947
76-e1d4-4395-872b-97b9511f214c", "kind": "workloads.cloudfoundry.org/v1alpha1, Kind=CFApp", "resource": {"group":"workloads.cloudfoundry.org","version":"v1alpha1","resource":"cfapps"}
}
    2021-12-01T10:45:48.616Z    INFO    cfapp-resource  Mutating CFApp webhook handler  {"name": "anothertest-app-guid-79fe6fde-2bf0-4de1-8c57-e839c5426692"}
    2021-12-01T10:45:48.616Z    DEBUG   controller-runtime.webhook.webhooks     wrote response  {"webhook": "/mutate-workloads-cloudfoundry-org-v1alpha1-cfapp", "code": 200, "reason":
 "", "UID": "ba494776-e1d4-4395-872b-97b9511f214c", "allowed": true}
[...]

Issue #383, opened 2021-12-20
[Explore]: Prevent duplicate records (both name and GUID)


matt-royal:
### Background

We are currently taking several different approaches to avoid duplicate records:
- We guard against multiple apps in a space with the same name using a [validating web hook](https://github.com/cloudfoundry/cf-k8s-controllers/blob/main/controllers/webhooks/workloads/cfapp_validation.go)
- Several of our repo fetch methods in the api codebase [will error if they find duplicate results](https://github.com/cloudfoundry/cf-k8s-controllers/blob/203306701a4825584a514138bcd8cf73c877d7e4/api/repositories/app_repository.go#L440-L452)
- We have a ["leases" proposal](https://docs.google.com/document/d/1fYIulNhfXlp_y2Ah5EtoKDDX83s6V-WRu162xpziLZ8/edit?usp=sharing) to enforce uniqueness, but I don't believe it's implemented for any resources yet. It's also fairly heavy weight.
- There's a [proposal to use CF names as K8s names](https://docs.google.com/document/d/1swH9ZJgxluo1GIJnqCn54gNLosyd-aJqgynzHJkGH1I/edit), which would ensure uniqueness of name within a namespace. However, this has problems, as per the comments in the proposal

Generally at creation time we're not ensuring uniqueness of GUIDs (`metadata.name`) across namespaces, and of names (`spec.name`) within namespace.

### Acceptance Criteria

Explore the current solutions to this problem and synthesize a few options for how we can enforce global uniqueness of GUIDs and uniqueness of names within a space. We may use several different solutions at once depending on the type of resource, how dangerous it is to have duplicates, and how expensive and error prone the solution is.

### Number of days this exploration is expected to take. This is what is “pointed”.

3 days.

### Dev Notes

- Ideally we will prevent the resources from being created in the k8s API. Failing that, we should transition the duplicate record into an error state in the k8s API and ignore it in the CF API shim.

- The solution should prevent duplicates when using kubectl directly, not just when using the CF API

- Consider Open Policy Agent

kieron-dev:
Just to clarify we have the leases solution, wrapped in [NameRegistry](https://github.com/cloudfoundry/cf-k8s-controllers/blob/main/controllers/coordination/name_registry.go), for orgs, spaces and apps currently.

kieron-dev:
[...]

PR #401, opened 2022-01-05
Use the name registry to enforce unique app names

georgethebeatle:
#314

## What is this change about?
Make the app webhook that prevents app name duplication more robust by using the name registry rather than listing apps. The name registry implements locking based on lease objects.

No

## Acceptance Steps
Tests are green (less flaky)

@cloudfoundry/cf-k8s


kieron-dev (approved):
LGTM

tcdowney:
I didn't realize we had test flakiness due to this issue -- thought it was still mostly theoretical. 🤔  Makes sense to use the lease approach for Apps for now since App name uniqueness in a space is almost as important as Org/Space name uniqueness.

For context though, we're planning on doing an explore to investigate using Open Policy Agent Gatekeeper for resources where we want other forms of cross-resource validation more generally.

https://github.com/cloudfoundry/cf-k8s-controllers/issues/383

cc @matt-royal who wrote that explore / is also thinking about this

kieron-dev:
Yep, the tests for name uniqueness were falling after a low number of retries.

I don't think OPA would be able to enforce name uniqueness for the same reason the original validating webhook couldn't. If two processes attempt to create the same named app, and the webhooks each invoke OPA before k8s has created either app resource, both creates will be allowed as OPA can only checks for uniqueness within the existing set of app names which won't yet include the new name.

That webhook really needs to take a global lock and check for name uniqueness in a list that is always correct and complete when viewed with the lock. And the cheap way to do that is with k8s' built-in namespace/name uniqueness enforcement using a separate resource with no side effects like Lease.

PR #404, opened 2022-01-06
Install dependencies works for different environments

julian-hj:

No

## What is this change about?

 installations using gcr.io and local registry require different tags for
  the kpack configuration. This change swaps in the hostname of the
  docker registry instead of gcr.io if it has been set in the
  DOCKER_SERVER environment variable.

no

## Acceptance Steps

- `./scripts/deploy-on-kind.sh kind local`
- Check the status of your clusterbuilder. You should see that it becomes ready
- Do the same thing, only using gcr (omit the last arg to the script)

@akrishna90


akrishna90 (approved):
pairf with @julian-hj, lgtm!

PR #408, opened 2022-01-06
add options to deploy-on-kind.sh

julian-hj:

No

## What is this change about?

This PR helps to streamline the `deploy-on-kind.sh` script for developer workflows and make it more approachable with usage text. It also applies the default domain for kind automatically so that we don't have to do it manually each time we set up a cluster, and exposes port 30050 so that you can access the local docker registry if you're using that option.

New command line options:
  - -l for local registry
  - -c to skip everything except controller build and deployment
  - -a to skip everything except api build and deployment
  - -v for bash -x output
  - -h usage text
  
No, the default usage remains the same.

## Acceptance Steps
- run the following
  ```
  ./scripts/deploy-on-kind.sh kind -l
  cf api http://localhost --skip-ssl-validation
  cf login #choose "kind-kind"
  ```
- now push an app--it should just work
- now re-run the script with the -a or -c flag, and observe that it only updates the api shim or controller
- now re-run the script with the -v flag and observe that it produces debug output

### DIFF
--- a/scripts/deploy-on-kind.sh
+++ b/scripts/deploy-on-kind.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -euxo pipefail
+set -euo pipefail
 
 SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
 ROOT_DIR="$SCRIPT_DIR/.."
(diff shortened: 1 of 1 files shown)


gnovv (approved):
lgtm!

PR #389, opened 2021-12-22
Add Destinations field to CFRoute Status

acosta11:
## Is there a related GitHub Issue?
#376 

## What is this change about?
See related issue.

## Does this PR introduce a breaking change?
No, the new status field is not required.

## Acceptance Steps
See related issue.

## Tag your pair, your PM, and/or team
paired w/ @julian-hj

PR #432, opened 2022-01-10
Added auth to space delete endpoint

gnovv:
Co-authored-by: Andrew Costa <ancosta@vmware.com>

#389

## What is this change about?
Adds auth to space delete 

No

## Acceptance Steps
See issue

@acosta11


acosta11 (approved):
LGTM

If anyone has opinions on how we should do error type matching, would be curious to know in case the BeAssignableToTypeOf function is different than yall expected.

PR #436, opened 2022-01-11
Clone config before creating new k8s user client and introduce --race flag to ginkgo tests

kieron-dev:
#437

## What is this change about?
When testing authorization for creating apps, we noticed some tests were flaky when run in parallel. Further investigation showed the authenticated user names were being mixed up. One test would appear to be using the user from another test.

The problem is in the user client factory. It takes a pointer to rest.Config, and although it copies this on instantiation, it mutates the same pointer when producing clients. This means if clients are produced simultaneously, there is a potential race condition and their identities can be mixed up. The fix is to clone the config before producing the client.

The mitigation to stop future races is to turn the `--race` flag on in ginkgo. This PR also does that.

No

## Acceptance Steps
New test is green. I don't think you could easily reproduce this manually.

@cloudfoundry/cf-k8s

### DIFF
--- a/api/repositories/k8s_client.go
+++ b/api/repositories/k8s_client.go
@@ -27,9 +27,11 @@ func NewUnprivilegedClientFactory(config *rest.Config) UnprivilegedClientFactory
 }
 
 func (f UnprivilegedClientFactory) BuildClient(authInfo authorization.Info) (client.Client, error) {
+	config := rest.CopyConfig(f.config)
+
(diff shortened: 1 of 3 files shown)


georgethebeatle (approved):
Cool, looks good! Does this make any `Eventually` statements redundant?

kieron-dev:
> Does this make any `Eventually` statements redundant?

Nope. We didn't add any in the end.

PR #451, opened 2022-01-12
Delete namespaces in parallel in e2e tests

kieron-dev:
No

## What is this change about?
It is only necessary to wait for spaces to be deleted before deleting the org, so we can parallelise most deletions to speed up e2e tests

No

## Acceptance Steps
e2e tests are green

@cloudfoundry/eirini


kieron-dev:
This PR should be edited to compare against main if and when #450  is merged

georgethebeatle (approved):

PR #455, opened 2022-01-12
New ADR for resource name prefixes

julian-hj:
No. (we failed to copy it over from Jira) 

It records a decision we made in team forum to prefix guids with strings so that the resulting k8s objects would also be decorated.

@tcdowney PTAL

### DIFF
--- a/api/docs/architecture-decisions/0004-resource-name-prefixes.md
+++ b/api/docs/architecture-decisions/0004-resource-name-prefixes.md
@@ -1,0 +1,41 @@
+# CF Resource GUID Prefixing
+
+Date: 2022-01-12
+
+## Status
+
+Draft
+
+## Context
+
+We have received feedback from the field that naming standard k8s typed objects (e.g. namespaces, statefulsets) with raw guids makes the system difficult for a k8s operator to use.
+
+For example:
+- it's hard when looking at a list of namespaces to tell which ones are CF-created. That makes it difficult to uninstall the things you installed to a cluster if you're just kicking the tires.
+- it's hard to tell which namespaces are cf spaces versus orgs. When listing other things like secrets or roles, it is hard to tell if they are in a space or an org or are totally unrelated to CF.
+- it's hard to be sure which stateful sets are CF apps, versus possibly other workloads that also have GUID names.
+
+The proposal to address this issue is to prefix the names of these resources with strings (e.g. "cforg-") that clearly identify them as CF resources. The request is to add this prefix to the object name rather than a label or other metadata, in order to make it easily discoverable.
+
+We considered 2 possible approaches to implement this:
+1. At the time of GUID generation for spaces, orgs, and processes, we simply prepend a prefix, and then treat that prefix as an opaque part of the GUID everywhere else. This option is far less costly, but will result in GUIDs returned to the CF API client with non-standard formatting.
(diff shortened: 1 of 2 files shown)


api/docs/architecture-decisions/0004-resource-name-prefixes.md
```diff
@@ -0,0 +1,41 @@
...
+## Consequences
+
+### Pros
+* This solution is trivial to implement.
+* This decision is reversible. So long as we don't rely on the existence of the prefix, we can change or remove it at any time.
+* This decision is extensible. We can decide to apply prefixes to other object type GUIDs (e.g. CFRoute, CFApp, etc.) as we see fit.
+* Because we are not introducing a discrepancy between GUID and object name, users can still rely on them to be the same. (I.e. `kubectl get secret -n $(cf space s --guid)` will continue to work)
+
+### Cons
+* Our GUIDs will no longer conform to [RFC 4122](https://www.ietf.org/rfc/rfc4122.txt), and therfore will not be convertible into binary format.
```
tcdowney, line 36:
```suggestion
* Our GUIDs will no longer conform to [RFC 4122](https://www.ietf.org/rfc/rfc4122.txt), and therefore will not be convertible into binary format.
```

api/docs/architecture-decisions/0004-resource-name-prefixes.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Open Questions
+
+* Should we extend this decision to other types?
+* Some CF API clients may have fixed length columns in a database, or columns that are guid typed. Our new guids may not fit or may fail to parse. Does any such client exist?
```
tcdowney, line 41: tl;dr I think there are enough operator compatibility differences that existing clients that are doing this will need to be updated anyway.

Ones that I'm aware of are:
* App/Task/Service usage accounting services that store off usage events and store their GUID in the db. We don't support these events anyway right now and I imagine those types of services would require bigger changes to work.
* APM agents that watch the log stream (logs in CF for VMs are tagged with these GUIDs). That will also be significantly different (as there is no loggregator subsystem or Diego cells), so also probably not a huge concern.

api/docs/architecture-decisions/0004-resource-name-prefixes.md
```diff
@@ -0,0 +1,41 @@
...
+## Context
+
+We have received feedback from the field that naming standard k8s typed objects (e.g. namespaces, statefulsets) with raw guids makes the system difficult for a k8s operator to use.
+
+For example:
+- it's hard when looking at a list of namespaces to tell which ones are CF-created. That makes it difficult to uninstall the things you installed to a cluster if you're just kicking the tires.
+- it's hard to tell which namespaces are cf spaces versus orgs. When listing other things like secrets or roles, it is hard to tell if they are in a space or an org or are totally unrelated to CF.
+- it's hard to be sure which stateful sets are CF apps, versus possibly other workloads that also have GUID names.
+
+The proposal to address this issue is to prefix the names of these resources with strings (e.g. "cforg-") that clearly identify them as CF resources. The request is to add this prefix to the object name rather than a label or other metadata, in order to make it easily discoverable.
```
tcdowney, line 18: Minor:

I think having it be `cf-` (e.g. `cf-org-`, `cf-space-`, etc.), might be easier to parse visually (and perhaps programmatically for things like the HNC namespace allowlist).

tcdowney (commented):
Overall, LGTM.

Added a suggestion for a typo fix and some thoughts about how the prefix should look.

tcdowney (approved):
LGTM!

Issue #380, opened 2021-12-20
[Feature]: Platform Operator can delete an org


julian-hj:
### Background

**As a** Platform Operator
**I want** to delete an org
**So that** can easily clean up all of the spaces, CFApps, CFRoutes CFPackages, etc. in that org

### Acceptance Criteria

**GIVEN** A deployed cf-k8s installation, with an org, a space, and one or more pushed applications
**WHEN I** `cf delete-org <org-name>`
**THEN I** can observe that the namespace for the org along with the space subnamespace, all of the contained applications and their dependent objects have been cleaned up.

### Dev Notes

`kubectl delete subns $(cf org o --guid) -n cf` **does not** work when there are spaces within the org.

> Error from server (Forbidden): admission webhook "namespaces.hnc.x-k8s.io" denied the request: Please set allowCascadingDeletion first

`kubectl hns set $(cf org o --guid) --allowCascadingDeletion` followed by
`kubectl delete subns $(cf org o --guid) -n cf` **works** and does all the right things.
> **Note:** should we set the allowCascadingDeletion flag when we create the namespaces? Or just wait to set it right before deletion?

After discussion with Eric, we should pursue the easy implementation first (putting this logic into the API Shim) and possibly pursue creation of a CFOrg custom resource later outside the context of this story.

gcapizzi:
I have created #438 for the authorization aspects of this story.

PR #464, opened 2022-01-13
Platform Operator can delete an Org

akrishna90:

#380 

## What is this change about?

Platform Operator can delete an Org
- Sets `allowCascadingDelete` to `true` on the hierarchyConfigurations object for Orgs to be deleted
- Deleting `subnamespaceanchor` object deletes `namespace` because of OwnerRefs

No

## Acceptance Steps

Follow acceptance on #380 

@julian-hj


api/repositories/org_repository.go
```diff
@@ -17,14 +19,16 @@ import (
...
+//+kubebuilder:rbac:groups=hnc.x-k8s.io,resources=subnamespaceanchors,verbs=list;create;delete;watch
+//+kubebuilder:rbac:groups=hnc.x-k8s.io,resources=hierarchyconfigurations,verbs=get;list;watch;update
 //+kubebuilder:rbac:groups="",resources=serviceaccounts;secrets,verbs=get;list;create;delete;watch
 
 //counterfeiter:generate -o fake -fake-name CFOrgRepository . CFOrgRepository
 
 type CFOrgRepository interface {
 	CreateOrg(ctx context.Context, info authorization.Info, org CreateOrgMessage) (OrgRecord, error)
 	ListOrgs(ctx context.Context, info authorization.Info, orgNames []string) ([]OrgRecord, error)
+	DeleteOrg(context.Context, authorization.Info, DeleteOrgMessage) error
```
matt-royal, line 31: As a general rule in Go, you want to define your interfaces where they're used and not where the implementation is defined. This encourages you to keep the interface smaller, since you only add methods you use, and also makes it simpler to swap in a different implementation later.

I missed that we'd started to move in that direction here. I know you didn't introduce the pattern in this PR, but this would be a good time to move the interface or at least not consolidate on the interface in the "wrong" place.

api/repositories/org_repository.go
```diff
@@ -393,6 +401,91 @@ func (r *OrgRepo) GetSpace(ctx context.Context, info authorization.Info, spaceGU
...
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return NotFoundError{
+				Err: err,
+			}
+		}
+		return err
+	}
+
+	if r.authEnabled {
```
matt-royal, line 418: I'd restructure this. Instead of having two separate, almost-identical paths, we can branch to choose which client to use, then rejoin a single path to do the updates and deletes. The only difference I see between the two paths is the check for IsForbidden, and I don't think it's an issue to have that for both cases.

api/repositories/org_repository.go
```diff
@@ -393,6 +401,91 @@ func (r *OrgRepo) GetSpace(ctx context.Context, info authorization.Info, spaceGU
 	return spaceRecords[0], nil
 }
 
+func (r *OrgRepo) DeleteOrg(ctx context.Context, info authorization.Info, message DeleteOrgMessage) error {
+	// Fetch 'hierarchy' object using the privileged client as we want to return a 'not-found' error for
+	// non-existent orgs as opposed to saying 'permission-denied'.
```
matt-royal, line 406: I'm not sure if you got notifications for the comment I deleted, but this comment has sent me on a journey. I'd suggest just deleting the comment, since the test suite documents the behavior and the comment doesn't make sense until you've read the whole method.

api/repositories/org_repository.go
```diff
@@ -393,6 +401,91 @@ func (r *OrgRepo) GetSpace(ctx context.Context, info authorization.Info, spaceGU
...
+	hierarchyObj := v1alpha2.HierarchyConfiguration{}
+	err := r.privilegedClient.Get(ctx, types.NamespacedName{Name: "hierarchy", Namespace: message.GUID}, &hierarchyObj)
```
matt-royal, line 408: "hierarchy" should be extracted into a constant instead of a magic string.

matt-royal (changes requested):
I suggested some changes, but otherwise it looks good.

api/repositories/org_repository.go
akrishna90: This is great feedback! @matt-royal 

@julian-hj and I saw that there were 2 interfaces (one is the repository and one in the handler) for the same org-repo implementation and we ended up deleting one in the handler. Looking back we should have probably looked at other repos to make a decision on which on to delete 👍🏽

api/repositories/org_repository.go
julian-hj: The caveat I would add here is that we plan to remove one of the two almost-identical paths soon, as we don't really want to support the non RBAC use case; so we should make sure that it will be easy to see what code ought to be discarded when we go to discard it.

Having said that, I don't think that doing the refactor @matt-royal suggests would cause problems--we'd just be shrinking the conditional logic down to a single block where we set the client.

api/repositories/org_repository.go
akrishna90: Addressed in commit - https://github.com/cloudfoundry/cf-k8s-controllers/pull/464/commits/a99f88f188236f227f4a094857f8ada66c9204e7

api/repositories/org_repository.go
akrishna90: I have tried it keep it simple and easy to remove later - changes in commit https://github.com/cloudfoundry/cf-k8s-controllers/pull/464/commits/a99f88f188236f227f4a094857f8ada66c9204e7

api/repositories/org_repository.go
akrishna90: Addressed in commit - https://github.com/cloudfoundry/cf-k8s-controllers/pull/464/commits/a99f88f188236f227f4a094857f8ada66c9204e7

api/repositories/org_repository.go
akrishna90: Addressed in commit - https://github.com/cloudfoundry/cf-k8s-controllers/pull/464/commits/a99f88f188236f227f4a094857f8ada66c9204e7

api/repositories/org_repository.go
julian-hj: @matt-royal @akrishna90 I suspect that if you just move the interface into the "api" package and use it from there, you will get a cyclic dependency error, since the "api" package already imports the "repositories". You might need to move interfaces into a separate package if we don't want them in "repositories".

julian-hj: ...or apparently not, since it doesn't seem as though the repositories repo was actually using that interface. So, never mind...

julian-hj (approved):
:+1:

api/config/base/rbac/role_binding.yaml
```diff
@@ -8,5 +8,5 @@ roleRef:
   name: cf-admin-clusterrole
 subjects:
 - kind: ServiceAccount
-  name: cf-admin-serviceaccount
-  namespace: default
+  name: cf-k8s-api-cf-admin-serviceaccount
+  namespace: cf-k8s-api-system
```
julian-hj, line 12: I think maybe this is a POC change, and we didn't mean to commit it?

julian-hj (changes requested):
IIRC, the role binding change we made was something we didn't intend to commit--we just did it to test something. Does that sound right to you?

api/config/base/rbac/role_binding.yaml
akrishna90: The [service-account](https://github.com/cloudfoundry/cf-k8s-controllers/blob/issues/380-delete-org/api/config/base/rbac/service_account.yaml) gets created in `cf-k8s-api-system` and get prefixed with `cf-k8s-api-`. I deployed this on Kind to verify the same. I think the changes are needed. If not for the changes, the ClusterRoleBinding will be pointing at a non existent service-account.

akrishna90: @julian-hj You were correct. This has been addressed in commit 7ebed6d1bd42cdf09dec7b90d7e1cad28adfc590

davewalter (approved):

Issue #466, opened 2022-01-14
[Chore]: Add an admin user to the deployment produced by the `deploy-on-kind` script


georgethebeatle:
### Background

Right now if you deploy cf-on-k8s on a kind cluster (the only deployment option so far) and configure a user to have some role like `SpaceDeveloper` or `SpaceManager` you instantly lose privilege to do things like list all orgs or other operations that require cf admin access. One way to work around that is to deploy with auth disabled, but the you cannot test/accept authorization so it does not serve all purposes well.

Adding a cf admin user to the kind deployment will address these issues and will make for a more flexible usage of cf-on-k8s

### Action to take

Make sure that an admin user gets created an configured when cf-on-k8s id deployed on kind.

### Impact

It is possible to do admin operations on your cluster while also using another user with some cf role assigned. You can now switch between admin and  non-admin users.

PR #479, opened 2022-01-17
Add an admin user on kind deployment

georgethebeatle:
Use the new scripts/create-new-user.sh script to create a
user with name admin and bind this user to the cf-k8s-controllers-admin
role. This role is a clone of the cf-k8s-controllers-space-developer
to begin with and should be incrementally granted more permissions as we
add new endpoints to the api. The goal is to eventually have and admin
role that is just as powerful as needed, but not as powerful as the k8s
admin user.

#466 

## What is this change about?
Use the new scripts/create-new-user.sh script to create a
user with name admin and bind this user to the cf-k8s-controllers-admin
role. This role is a clone of the cf-k8s-controllers-space-developer
to begin with and should be incrementally granted more permissions as we
add new endpoints to the api. The goal is to eventually have and admin
role that is just as powerful as needed, but not as powerful as the k8s
admin user.

No

## Acceptance Steps
- `make test-e2e` is green 
- `cf api http://localhost`
- `cf login`, chose the `admin` option
- `cf create-org myorg` 
- `cf orgs` displays the created org 

@cloudfoundry/cf-k8s

### DIFF
--- a/api/tests/e2e/e2e_suite_test.go
+++ b/api/tests/e2e/e2e_suite_test.go
@@ -73,10 +73,7 @@ var _ = BeforeSuite(func() {
 	config, err := controllerruntime.GetConfig()
 	Expect(err).NotTo(HaveOccurred())
 
-	// fine for Kind cluster; might need work for other cluster types
-	cert := config.CertData
-	cert = append(cert, config.KeyData...)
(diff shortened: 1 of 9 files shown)


danail-branekov (approved):
lgtm

PR #481, opened 2022-01-17
Setup KUBECONFIG outside of deploy-on-kind scripts

gcapizzi:
This changes `scripts/deploy-on-kind.sh` to always use the current `$KUBECONFIG` instead of exporting its own one. This allows users to create development clusters using their default config, while `scripts/run-tests.sh` can deploy test clusters with a separate config just by exporting a different `$KUBECONFIG` _before_ calling `scripts/deploy-on-kind.sh`.


georgethebeatle (approved):
LGTM

Issue #484, opened 2022-01-18
[Chore]: Extract generic helpers for making test API calls in the API e2e test suite


georgethebeatle:
### Background

Currently some api e2e test are defining their own helpers for making test API calls to the endpoint they are testing. Here are a few examples:

- https://github.com/cloudfoundry/cf-k8s-controllers/blob/1a60080607a66ce9041ec36c740379373903dd00/api/tests/e2e/e2e_suite_test.go#L150-L172
- https://github.com/cloudfoundry/cf-k8s-controllers/blob/1a60080607a66ce9041ec36c740379373903dd00/api/tests/e2e/e2e_suite_test.go#L226-L285
- https://github.com/cloudfoundry/cf-k8s-controllers/blob/1a60080607a66ce9041ec36c740379373903dd00/api/tests/e2e/e2e_suite_test.go#L466-L506

This results in tests that are hard to read, because the actual endpoint is not seen as text in the `BeforeEach` or `It` statement, but is buried under a chain of helpers. What is more, these helpers are doing almost identical http requests, so we have code repetition as well.

### Action to take

Let's refactor our e2e tests to use some rest client library like [resty](https://github.com/go-resty/resty) or even build our own if there is no suitable one out there.

### Impact

The test are easier to read, because they are using a standard set of http request helpers ands it is visible what endpoint each test is hitting.

### Dev Notes

Resty seems like it could do the trick, please refer to this [Draft PR](https://github.com/cloudfoundry/cf-k8s-controllers/pull/494/files) for an example of what the tests could look like.

georgethebeatle:
I have drafted something like what is described above:

https://github.com/cloudfoundry/cf-k8s-controllers/pull/494

Based on how people feel about it could serve as a basis for solving this chore. If we end up deciding it's no good, the chore will be closed with it

PR #494, opened 2022-01-19
Generic cf api helper for the api e2e tests

georgethebeatle:
This is an attempt to. create a generic test helper that will make it easy and readable to make any api call in our e2e tests.
The idea is to keep it concise and flexible , while still being obvious what it does. This is a draft PR so please leave comments on whether you like it better than the old style, what works and what doesn't. I have covered the orgs and roles tests which should be enough to hit all current needs.

### DIFF
--- a/api/tests/e2e/e2e_suite_test.go
+++ b/api/tests/e2e/e2e_suite_test.go
@@ -45,6 +45,7 @@ import (
 
 var (
 	k8sClient           client.Client
+	api                 helpers.APIRequest
 	clientset           *kubernetes.Clientset
 	rootNamespace       string
 	apiServerRoot       string
@@ -94,6 +95,7 @@ var _ = BeforeSuite(func() {
 })
 
 var _ = BeforeEach(func() {
+	api = helpers.NewCFAPI(apiServerRoot)
 	tokenAuthHeader = fmt.Sprintf("Bearer %s", serviceAccountToken)
 })
 
@@ -138,9 +140,9 @@ func deleteSubnamespace(parent, name string) {
 		},
 	}
 	err := k8sClient.Delete(ctx, &anchor)
-	Expect(err).NotTo(HaveOccurred())
+	ExpectWithOffset(2, err).NotTo(HaveOccurred())
 
-	Eventually(func() bool {
+	EventuallyWithOffset(2, func() bool {
 		err := k8sClient.Get(ctx, client.ObjectKeyFromObject(&anchor), &anchor)
 
 		return errors.IsNotFound(err)
@@ -159,15 +161,12 @@ func createOrgRaw(orgName, authHeader string) *http.Response {
 }
 
 func createOrg(orgName, authHeader string) presenter.OrgResponse {
-	resp := createOrgRaw(orgName, authHeader)
-	Expect(resp).To(HaveHTTPStatus(http.StatusCreated))
-	Expect(resp).To(HaveHTTPHeaderWithValue(headers.ContentType, "application/json"))
-	defer resp.Body.Close()
-
 	org := presenter.OrgResponse{}
-	err := json.NewDecoder(resp.Body).Decode(&org)
-	Expect(err).NotTo(HaveOccurred())
-
+	api.Request(http.MethodPost, "/v3/organizations").
+		WithBody(orgPayload(orgName)).
+		DoWithAuth(tokenAuthHeader).
+		ValidateStatus(http.StatusCreated).
+		DecodeResponseBody(&org)
 	return org
 }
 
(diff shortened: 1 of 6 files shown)


api/tests/e2e/e2e_suite_test.go
```diff
@@ -138,9 +140,9 @@ func deleteSubnamespace(parent, name string) {
 		},
 	}
 	err := k8sClient.Delete(ctx, &anchor)
-	Expect(err).NotTo(HaveOccurred())
+	ExpectWithOffset(2, err).NotTo(HaveOccurred())
```
danail-branekov, line 143: I am a bit skeptical about using offsets as you cannot really know the stack depth the utility function usage. I wish failures also reported a stack trace, similar to exceptions in Java, unfortunately golang errors does not support that out of the box

api/tests/e2e/e2e_suite_test.go
```diff
@@ -159,15 +161,12 @@ func createOrgRaw(orgName, authHeader string) *http.Response {
...
-	resp := createOrgRaw(orgName, authHeader)
-	Expect(resp).To(HaveHTTPStatus(http.StatusCreated))
-	Expect(resp).To(HaveHTTPHeaderWithValue(headers.ContentType, "application/json"))
-	defer resp.Body.Close()
-
 	org := presenter.OrgResponse{}
-	err := json.NewDecoder(resp.Body).Decode(&org)
-	Expect(err).NotTo(HaveOccurred())
-
+	api.Request(http.MethodPost, "/v3/organizations").
```
danail-branekov, line 165: How about 
```
api.CreateOrgRequest()
.WithBody(orgPayload(orgName)).
....
```

We would thus not have to remember the verb and the endpoint of an operation which might be annoying

georgethebeatle: I kind of like the explicitness of it, because it makes it very obvious what the test is doing. If we define a `CreateXXXRequest` for each restful resource the helper will end up having a million methods, but we could use the constants from the handlers and have something like 
```
api.CreateOrgRequest()
api.PostRequest(apis.OrgsEndpoint).
...
```
if you thing this is a better direction...

api/tests/e2e/e2e_suite_test.go
georgethebeatle: Me too :). I had to adjust these while I was driving this, since the ginkgo failures were really cryptic and low level. Not sure what is the best way of handling this.

api/tests/e2e/e2e_suite_test.go
danail-branekov: My point was that with the explicit API, even using constants, you would have to remember whether it is a post/put/patch, whatever. If we do not have endpoints serving multiple HTTP verbs we could have something like

```
api.Request(apis.CreateOrgEndpoint).....
```

and let the `Request` method figure out whether it should be a post or put or patch (via maybe supporting an internal map between endpoint and HTTP verb)

georgethebeatle: Yep, I get that. It is true that we have to remember the http verb, but as I said I kind of like this explicitness - that you can tell that this test tests `POST "/v3/organizations"` which maps nicely to what is written in the story acceptance.

The lack of explicitness is what I do not like about our current tests. As a matter of fact our current set of functions already offer the ease of use that you are after:
```
org = createOrg(orgName, tokenAuthHeader)
``` 

but they lack explicitness and sometimes when the argument list is too long can even be confusing:
```
role := createOrgRole("organization_manager", rbacv1.UserKind, userName, org.GUID, tokenAuthHeader)
```
vs
```
api.Request(http.MethodPost, "/v3/roles").
    WithBody(userOrgRolePayload("organization_user", userName, org.GUID)).
    DoWithAuth(tokenAuthHeader).
    ValidateStatus(http.StatusCreated)
```

api/tests/e2e/e2e_suite_test.go
tcdowney: This was new to me as well.. so just so I'm understanding it right, using `ExpectWithOffset` should show the _actual test code line_ that failed instead of the LOC of this helper function? I share how y'all feel about this feeling brittle, but I don't know a better way either.

Maybe @matt-royal has Gomega thoughts as well. 🤔

matt-royal: I agree that offsets are brittle. Also, when I've experimented with them I didn't find the resulting stack traces to be any clearer.

georgethebeatle:
This is now being driven by https://github.com/cloudfoundry/cf-k8s-controllers/issues/484, so this PR is no longer needed.

PR #495, opened 2022-01-19
Implement authZ for create role

kieron-dev:
#237

## What is this change about?
Require the cf admin role to create roles.

Note that creating roles in cf equates to creating role bindings in k8s. There is a restriction that a user can only create role bindings that grant permissions that the user already holds. Therefore note that the cf admin role _must_ contain the permissions of the other cf roles.

We now require the admin role in order to be able to create other roles

## Acceptance Steps
Create a couple of extra users. Check you cannot `cf set-org-role` or `cf set-space-role` on one of these users logged in as the other.

Then try the same logged in as the admin user. These should work.

@cloudfoundry/eirini


mnitchev (approved):
LGTM

Issue #235, opened 2021-11-11
[Feature]: Developers should not be able to create spaces


gcapizzi:
### Background

**As an** admin
**I want** to be the only one able to create spaces
**So that** I can keep my space structure under control

### Acceptance Criteria

**GIVEN** I am not an `Admin`
**WHEN I** try to `POST /v3/spaces`
**THEN I** should get the following response:
```
HTTP/1.1 403 Forbidden
Content-Type: application/json
```
```json
{
  "errors": [
    {
      "code": 10003,
      "title": "CF-NotAuthorized",
      "detail": "You are not authorized to perform the requested action"
    }
  ]
}
```

**GIVEN** I am an `Admin`
**WHEN I** try to `POST /v3/spaces`
**THEN I** should get a successful response

### Dev Notes

* This could trigger the creation of an `Admin` `ClusterRole`, to be bound in the root namespace so that it propagates to all orgs and spaces.

PR #499, opened 2022-01-19
Allow cf-admins only to create spaces

danail-branekov:
https://github.com/cloudfoundry/cf-k8s-controllers/issues/235

## What is this change about?
Allow space creation to users that have the `cf-admin` role assigned.

The implementation would use the user k8s client (instead the privileged one) in the space repository to create the space subnamespace anchor. This means that the user has to be assigned to the `cf-admin` role which is the only one allowing subnamespace anchor creation.

No (or sort of - from now on regular users would not be able to create spaces :) )

## Acceptance Steps

0. Deploy on kind and `cf api` the cluster 

### Admin user
1. login as `admin`
2. `cf create-org o`
3. `cf create-space -o o s`
4. Make sure the commands above caused no errors

### Non-admin users
1. `./scripts/create-new-user.sh nonadmin`
2. Assign the `organization-user` role to `nonadmin`
```
cat <<EOF | kubectl apply -f -
apiVersion: rbac.authorization.k8s.io/v1
kind: RoleBinding
metadata:
  name: nonadmin-binding
  namespace: cf
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: ClusterRole
  name: cf-k8s-controllers-organization-user
subjects:
- apiGroup: rbac.authorization.k8s.io
  kind: User
  name: nonadmin
EOF

```
2. Login as `nonadmin`
3. `cf create-space -o o s2` should fail with `You are not authorized to perform the requested action`

@kieron-dev 
@mnitchev

### DIFF
--- a/api/apis/space_handler.go
+++ b/api/apis/space_handler.go
@@ -88,6 +88,13 @@ func (h *SpaceHandler) SpaceCreateHandler(info authorization.Info, w http.Respon
 			return
 		}
 
+		if repositories.IsForbiddenError(err) {
+			h.logger.Error(err, "not allowed to create spaces")
+			writeNotAuthorizedErrorResponse(w)
+
+			return
+		}
+
 		h.logger.Error(err, "Failed to create space", "Space Name", space.Name)
 		writeUnknownErrorResponse(w)
 		return
--- a/api/apis/space_handler_test.go
+++ b/api/apis/space_handler_test.go
@@ -138,6 +138,16 @@ var _ = Describe("Spaces", func() {
 			})
 		})
 
+		When("user is not allowed to create a space", func() {
+			BeforeEach(func() {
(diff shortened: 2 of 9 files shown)


api/apis/space_handler.go
```diff
@@ -88,6 +88,13 @@ func (h *SpaceHandler) SpaceCreateHandler(info authorization.Info, w http.Respon
 			return
 		}
 
+		if repositories.IsPermissionDeniedOrNotFoundError(err) {
```
kieron-dev, line 91: We created a ForbiddenError as part of the last story. Might be controversial, but you could use that here and in the repo.

danail-branekov: Thanks! Using the `ForbiddenError` does make sense here as it is a bit weird to get a "bla-bla-or-does-not-exist` error when you are creating new stuff.

matt-royal (approved):
Looks good to me

PR #512, opened 2022-01-21
DELETE v3/routes/{guid}

### LINKED ISSUE #491: [Feature]: Space developer can delete a Route
### Blockers/Dependencies

_No response_

### Background

**As a** space developer
**I want** to be able to delete a route
**So that** I can remove unwanted  routes

The api docs specify route deletion here: https://v3-apidocs.cloudfoundry.org/version/3.111.0/index.html#delete-a-route

Example request:
```
curl "https://api.example.org/v3/routes/[guid]" \
  -X DELETE \
```

Example response
```
HTTP/1.1 202 Accepted
Location: https://api.example.org/v3/jobs/route.delete-[guid]
```

### Acceptance Criteria

**GIVEN** A Route exists
**WHEN I** use the CLI or API to delete a route
**THEN I** see that the route no longer exists

### Dev Notes

* Ensure that you support using the user client in addition to the privileged client based on the config.
* This is a `jobs` implementation. We have previously set up a dummy jobs endpoint for other cases and should align with current implementation.

gnovv:
closes #491

## What is this change about?
Implements `DELETE v3/routes/{guid}`
- update api docs
- update space developer RBAC

no

## Acceptance Steps
see issue


georgethebeatle:
Can you add [api end-to-end tests](https://github.com/cloudfoundry/cf-k8s-controllers/tree/main/api/tests/e2e) for this endpoint, please? We are starting an effort of backfilling the e2e test since most of the endpoints are lacking these and it will help a lot if we write them along with the story. Otherwise we will be chasing a moving target and it will be a pain.

gnovv:
@georgethebeatle I am having trouble getting the e2e tests passing- I don't think I changed anything related to the ones that are failing.
It is difficult to jump in and write the new e2e tests for my endpoint when I am unfamiliar with this code and how it works.
Do you have any pointers or examples to help me get started?

georgethebeatle:
@gnovv unfortunately there are flakes, it is rue so it could be that some unrelated tests are failing, but I guess just ignore them for the scope of this story. It would be sufficient to write tests for `DELETE v3routes{guid}`, which is the endpoint that this story implements, isn't it? One way to ignore flakes would be to just focus your `It`s and run them in isolation to make sure they are passing. There are not too many e2e test at the moment to use as examples, but still there are enough to get some orientation. [This one](https://github.com/cloudfoundry/cf-k8s-controllers/blob/9fa20ec0e4d15c468d2c74282bc0ca1aeac6af5a/api/tests/e2e/roles_test.go#L23-L45) is relatively straightforward I think. On line 26 it invokes a helper that just POSTs on `/v3/roles` and in the `It` it checks that a role binding was created by the endpoint handler. Hope that helps.

Birdrock (approved):
/lgtm

gnovv:
@georgethebeatle we added an auth test to `api/apis/integration/route_test.go` instead of implementing it as an e2e test, similar to @kieron-dev 's PR: https://github.com/cloudfoundry/cf-k8s-controllers/pull/537

georgethebeatle:
@gnovv @Birdrock the tests that @kieron-dev moved in #537 were error cases, not happy paths, like the one we are discussing here. It is a good idea to move error cases to integration since they are faster and we we have more control there, because mocks can be used. While there is value in the test you've written there will also be value in an e2e that does the same checks. I realise that it might be a bit odd to start from the `DELETE` method, so updated #525 to include it as well. It would be great if US colleagues can help with backfilling e2e tests for this endpoint, since you generally have more context on it.

Issue #483, opened 2022-01-18
[Chore]: Leave authentication testing to the whoami e2e tests


georgethebeatle:
### Background

In theory authentication should be tested on each and every endpoint in the API shim. Unfortunately this will generate too many tests that are mostly repeating the same checks. Given that authentication is implemented as a middleware, there is not a big risk that it will work for one endpoint but not the other. In this line of thought it should be relatively safe to thoroughly test auth in just one endpoint, like `/whoami` and remove such tests from everywhere else.

### Action to take

In the `whoami` e2e test add the following test:
- `When` an empty auth header is added `It` returns `401 Unauthorized`
- `When` an invalid cert/token is added `It` returns `401 Unauthorized`

Remove all tests testing with invalid auth header. Some examples:
- https://github.com/cloudfoundry/cf-k8s-controllers/blob/021fbf4aae57eb82c53b7499aa5bd481eea2dae0/api/tests/e2e/orgs_test.go#L198-L203
- https://github.com/cloudfoundry/cf-k8s-controllers/blob/main/api/tests/e2e/spaces_test.go#L169-L178

In all e2e tests make sure that we are using the client cert auth header only (no need to check both mechanisms since it is covered by the `whoami` test). Here are some examples:
- https://github.com/cloudfoundry/cf-k8s-controllers/blob/1a60080607a66ce9041ec36c740379373903dd00/api/tests/e2e/orgs_test.go#L46) 
- https://github.com/cloudfoundry/cf-k8s-controllers/blob/021fbf4aae57eb82c53b7499aa5bd481eea2dae0/api/tests/e2e/spaces_test.go#L41

### Impact

API e2e test are better organized with less repetition.

PR #514, opened 2022-01-21
Move authentication tests to the whoami endpoint

mnitchev:
#483 

## What is this change about?

Since we are using a middleware for parsing the authentication header,
we should just test that in one place, instead of on every endpoint.
This should eliminate a lot of duplication.
Additionally add some missing tests to the user client factory.

No

## Acceptance Steps
Look at the tests 👀 

@kieron-dev


kieron-dev (approved):
LGTM

Issue #446, opened 2022-01-12
[Feature]: Developers should only be able to set the current droplet for apps they have access to


gcapizzi:
### Background

**As an** admin
**I want** to assign the `SpaceDeveloper` role to the users I trust
**So that** they can push apps to their spaces

### Acceptance Criteria

**GIVEN** two spaces `space1` and `space2` exist
**AND** `app1` exists in `space1`
**AND** `app2` exists in `space2`
**AND** `user` has a `RoleBinding` to the `SpaceDeveloper` `ClusterRole` in `space1`

1. **WHEN I** [set the current droplet](https://v3-apidocs.cloudfoundry.org/version/3.113.0/index.html#set-current-droplet) for `app1` as `user`
  **THEN I** see it working

1. **WHEN I** set the current droplet for `app2` as `user`
  **THEN I** get the following response:
  ```
  HTTP/1.1 404 Not Found
  Content-Type: application/json
  ```
  ```json
  {
    "errors": [
      {
        "detail": "App not found",
        "title": "CF-ResourceNotFound",
        "code": 10010
      }
    ]
  }
  ```

PR #527, opened 2022-01-21
Implement authz on set current droplet endpoint

kieron-dev:
#446 

## What is this change about?
This adds cfapp patch permissions to space-developer and admin roles and uses the 'user client' to perform the patch.

We also added get permission to space-manager as a readonly user helped in the e2e test.

No

## Acceptance Steps
See you can set current droplet on an app as admin or spacedeveloper, but not as any other role.

@cloudfoundry/eirini


mnitchev (approved):
LGTM

Issue #447, opened 2022-01-12
[Feature]: Developers should only be able to list processes for apps they have access to


gcapizzi:
### Background

**As an** admin
**I want** to assign the `SpaceDeveloper` role to the users I trust
**So that** they can push apps to their spaces

### Acceptance Criteria

**GIVEN** two spaces `space1` and `space2` exist
**AND** `app1` exists in `space1`
**AND** `app2` exists in `space2`
**AND** `user` has a `RoleBinding` to the `SpaceDeveloper` `ClusterRole` in `space1`

1. **WHEN I** [list processes](https://v3-apidocs.cloudfoundry.org/version/3.113.0/index.html#list-processes-for-app) for `app1` as `user`
  **THEN I** see it working

1. **WHEN I** list processes for `app2` as `user`
  **THEN I** get the following response:
  ```
  HTTP/1.1 404 Not Found
  Content-Type: application/json
  ```
  ```json
  {
    "errors": [
      {
        "code": 10010,
        "title": "CF-ResourceNotFound",
        "detail": "App not found"
      }
    ]
  }
  ```

kieron-dev:
I think this is another case where we get a not found when the user is not authorized. This is the response from curling the endpoint on mel-b:
```
HTTP/1.1 404 Not Found
Transfer-Encoding: chunked
Content-Type: application/json; charset=utf-8
Date: Mon, 24 Jan 2022 11:36:34 GMT
Referrer-Policy: strict-origin-when-cross-origin
Server: nginx
X-B3-Spanid: 0c4c5425f60d2118
X-B3-Traceid: 0c4c5425f60d2118
X-Content-Type-Options: nosniff
X-Download-Options: noopen
X-Frame-Options: SAMEORIGIN
X-Permitted-Cross-Domain-Policies: none
X-Runtime: 0.012748
X-Vcap-Request-Id: fe23b0b0-afb9-4a9d-65c4-4d3c465b959a::5bdf829e-976e-407d-8004-1c7b04424b5b
X-Xss-Protection: 1; mode=block

{
  "errors": [
    {
      "detail": "App not found",
      "title": "CF-ResourceNotFound",
      "code": 10010
    }
  ]
}
```

gcapizzi:
Acceptance updated!

PR #546, opened 2022-01-24
Return 404 if not authorized to list app processes

kieron-dev:
#447 

## What is this change about?
Return 404 error if user is not authorized to list app processes. This currently equates to permissions to get the app, so no change has been made to the process repo code to not use the privileged k8s client.

No

## Acceptance Steps
See story

@cloudfoundry/eirini

### DIFF
--- a/api/apis/app_handler.go
+++ b/api/apis/app_handler.go
@@ -434,8 +434,8 @@ func (h *AppHandler) getProcessesForAppHandler(authInfo authorization.Info, w ht
 	app, err := h.appRepo.GetApp(ctx, authInfo, appGUID)
 	if err != nil {
 		switch err.(type) {
-		case repositories.NotFoundError:
-			h.logger.Info("App not found", "AppGUID", appGUID)
+		case repositories.PermissionDeniedOrNotFoundError:
+			h.logger.Info("Permission denied or app not found", "AppGUID", appGUID, "err", err)
 			writeNotFoundErrorResponse(w, "App")
 			return
 		default:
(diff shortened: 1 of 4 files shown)


api/repositories/app_repository_test.go
```diff
@@ -53,8 +53,8 @@ var _ = Describe("AppRepository", func() {
 		space2 = createSpaceAnchorAndNamespace(testCtx, org.Name, prefixedGUID("space2"))
 		space3 = createSpaceAnchorAndNamespace(testCtx, org.Name, prefixedGUID("space3"))
 
-		spaceDeveloperClusterRole = createSpaceDeveloperClusterRole(testCtx)
-		spaceAuditorClusterRole = createSpaceAuditorClusterRole(testCtx)
+		spaceDeveloperClusterRole = createClusterRole(testCtx, repositories.SpaceDeveloperClusterRoleRules)
```
danail-branekov, line 56: Provided that `spaceDeveloperClusterRole` and `spaceAuditorClusterRole` are clusterwide resources, they would not be deleted automatically on namespace destruction and I guess they would leak. How about deleting them on `AfterEach`?

This probably applies to all roles created via `createClusterRole`

danail-branekov (approved):
Except for the comment of deleting the cluster role on test tear down, lgtm

api/repositories/app_repository_test.go
danail-branekov: As discussed over slack, these are env tests so clusterwide resoources are destroyed anyway. So no need to cleanup explicitly.

Issue #538, opened 2022-01-24
[Bug] Roles are not set up (i.e. propagated) immediately after `create org` or `create space` API call returns


danail-branekov:
This causes lots of flakes in the E2E tests

kieron-dev:
It does not appear to be possible to determine when propagation into a new namespace is complete in the HNC design.

When a new namespace is added to the HN hierarchy, e.g. by the [anchor reconciler creating a namespace](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/550d0c85e42b60605f8518267cd8299580447839/internal/anchor/reconciler.go#L119) corresponding to a new subnamespaceanchor, the [hierarchyconfig reconciler](//github.com/kubernetes-sigs/hierarchical-namespaces/blob/master/internal/hierarchyconfig/reconciler.go) will trigger. When this notices a change, it [enqueues](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/550d0c85e42b60605f8518267cd8299580447839/internal/hierarchyconfig/reconciler.go#L758) all objects to be propagated in the namespace to the [object reconciler](https://github.com/kubernetes-sigs/hierarchical-namespaces/blob/550d0c85e42b60605f8518267cd8299580447839/internal/objects/reconciler.go#L115-L127).

Once the hierarchy config reconciler has enqueued the required objects, the connection between those objects and the namespace creation event is lost. The object reconcilers just process their queues and at some point those enqueued from the namespace creation event will be reconciled. There is no point where we can know all the object reconcilers are done with objects from the creation and it is appropriate to update a status on the anchor, say.

So I don't think a modification to HNC is going to be possible for this issue.

kieron-dev:
If we implement code to wait for rolebindings of the current user to propagate to the new namespace on namespace creation, there are also problems.

Suppose the current user is has only the org-manager role. This [permits them space creation](https://v3-apidocs.cloudfoundry.org/version/3.113.0/index.html#create-a-space) on their org. The cluster role allowing this gives create permission on subnamespace anchors. This role should not be bound in the new namespace, as that represents a space, and we should not allow subnamespaces of a space. So we would be waiting for a role that would never be propagated.

We might think about checking if a certain admin rolebinding was propagated. But there can be many admins, and the admin user might well be changed during a create namespace operation, so there's a potential for a race where we wait for a role binding that won't be propagated.

So I'm not sure we have a solution to this at present.

[...]

PR #547, opened 2022-01-24
E2E tests wait for admin role binding in orgs/spaces

danail-branekov:
https://github.com/cloudfoundry/cf-k8s-controllers/issues/538

## What is this change about?
The org repo waits for the subnamespaces anchor status to become OK,
unfortunately this only indicates that the underlying namespace has been
created. Objects propagation is carried out asynchronously, therefore
there is a tiny amount of time that rolebindings from the parent
namespace are not available into the child namespace. During this short
interval the org/space might be unusable even for admins as they are not
yet bound to their roles in that particular namespace.

HNS currently does not provide a mechanism to signal that propagation is
complete so it is very hard for us to implement a proper fix. Therefore
until we are sure that this is an actual problem, e2e tests would just
wait for the admin role binding to appear in the namespace which should
be good enough to avoid flakes.

No

## Acceptance Steps
e2e tests are passing consistently 

@mnitchev

### DIFF
--- a/api/tests/e2e/e2e_suite_test.go
+++ b/api/tests/e2e/e2e_suite_test.go
@@ -30,6 +30,7 @@ import (
 	"github.com/google/uuid"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	gomegatypes "github.com/onsi/gomega/types"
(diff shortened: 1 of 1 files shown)


mnitchev (approved):
LGTM

Issue #551, opened 2022-01-25
[chore] fix line number with custom fail handler


danail-branekov:
With the [custom fail handler in e2e tests](https://github.com/cloudfoundry/cf-k8s-controllers/blob/66ed165d02d73576f82c63ac33315e546ed5372d/api/tests/e2e/e2e_suite_test.go#L63) we are getting ginkgo errors like

```
 Expected
     <bool>: true
 to be false
 In [It] at: /home/danailster/go/pkg/mod/github.com/onsi/gomega@v1.18.0/internal/assertion.go:58
```

where the failure line is messed up.

PR #553, opened 2022-01-25
Fix line numbers in ginkgo error output in E2E tests

danail-branekov:
https://github.com/cloudfoundry/cf-k8s-controllers/issues/551

## What is this change about?
Ginkgo's default fail handler (the one the E2EFailHandler delegates to)
uses the `callerSkip` argument to skip stack frames when printing the
error line. As we wrap the standard fail handler with a custom one, we
should increase the `callerSkip` argument with one (wrapping the default
handler adds one stack frame) so that the error message points to the
actual failure.

This fix also works well with `ExpectWithOffset`. Being on the subject,
the offsets in `createRole` have been adjusted to `2` which is the
correct value.

No

## Acceptance Steps
* all e2e tests are green
* if an e2e test is made to fail deliberately, ginkgo's error message should refer to the line of failure


kieron-dev (approved):
LGTM

Issue #443, opened 2022-01-12
[Chore]: Fix flake in e2e tests where creating an org via the API fails


kieron-dev:
### Background

We've seen several occurrences of the error here: https://github.com/cloudfoundry/cf-k8s-controllers/runs/4705805944?check_suite_focus=true.

```
Unexpected error:
      <*url.Error | 0xc0007d48a0>: {
          Op: "Post",
          URL: "http://localhost/v3/organizations",
          Err: <*errors.errorString | 0xc000054110>{s: "EOF"},
      }
      Post "http://localhost/v3/organizations": EOF
  occurred
```

This is a create org call to the API in a BeforeEach call. EOF might indicate a panic, although the logs don't show that.

### Action to take

Investigate and fix.

### Impact

Fewer flakes. Perhaps fixing a real production issue?

danail-branekov:
It is very probably that those flakes are related to https://github.com/cloudfoundry/cf-k8s-controllers/issues/538

What most probably happens in the failure referenced is that after [creating the orgs](https://github.com/cloudfoundry/cf-k8s-controllers/blob/66ed165d02d73576f82c63ac33315e546ed5372d/api/tests/e2e/spaces_test.go#L110-L112), the test immediately [tries to create spaces](https://github.com/cloudfoundry/cf-k8s-controllers/blob/66ed165d02d73576f82c63ac33315e546ed5372d/api/tests/e2e/spaces_test.go#L122-L132) in them. If the admin role bindings have not been propagated to the orgs' namespaces, creating spaces would fail with a forbidden error.

We tried making tests more resilient to such sort of failures via PR https://github.com/cloudfoundry/cf-k8s-controllers/pull/547 but we only did that for the `createOrg` and `createSpace` functions. Instead, in order to address the issue here we should move the fix into `createOrgRaw` and `createSpaceRaw`

PR #557, opened 2022-01-25
Ensure admin role when creating orgs/spaces asynchronously

danail-branekov:
https://github.com/cloudfoundry/cf-k8s-controllers/issues/443

## What is this change about?
Replace the ginkgo matcher with a utility function that can return an
error. This error is either expected not to occur (sync create
org/space), or sent over the error channel in the async counterpart.

No

## Acceptance Steps
E2E tests are green


kieron-dev (approved):
LGTM

Issue #552, opened 2022-01-25
[Bug]: Admin can create a sub-space of a space


kieron-dev:
### What happened?

Logged in as admin:
1. create org
2. create space
3. get space guid
4. `cf curl -XPOST /v3/spaces` with json specifying the new space guid as the organzation guid.

The sub-space of the space is successfully created.

```

~/workspace/cf-k8s-controllers main
❯ cf create-org org
Creating org org as admin...
OK

TIP: Use 'cf target -o "org"' to target new org

~/workspace/cf-k8s-controllers main
❯ cf target -o org
API endpoint:   http://localhost
API version:    3.111.0+cf-k8s
user:           admin
org:            org
No space targeted, use 'cf target -s SPACE'

~/workspace/cf-k8s-controllers main
❯ cf create-space space
Creating space space in org org as admin...
OK

Assigning role SpaceManager to user admin in org org / space space as admin...
OK

Assigning role SpaceDeveloper to user admin in org org / space space as admin...
OK

TIP: Use 'cf target -o "org" -s "space"' to target new space

~/workspace/cf-k8s-controllers main 8s
❯ cf curl -i /v3/spaces -XPOST -d '{"name": "subspace", "relationships": {"organization":{"data":{"guid":"59b1e651-f4a9-4568-9fe4-f8406e
e96cb9"}}}}'
HTTP/1.1 201 Created
Transfer-Encoding: chunked
Content-Type: application/json
Date: Tue, 25 Jan 2022 12:01:31 GMT
Server: envoy
Vary: Accept-Encoding
X-Envoy-Upstream-Service-Time: 3644

{"name":"subspace","guid":"e2045607-02b5-4d41-9b29-ae99cd274442","created_at":"2022-01-25T12:01:30Z","updated_at":"2022-01-25T12:01:30Z"
,"links":{"self":{"href":"http://localhost/v3/spaces/e2045607-02b5-4d41-9b29-ae99cd274442"},"organization":{"href":"http://localhost/v3/
organizations/59b1e651-f4a9-4568-9fe4-f8406ee96cb9"}},"metadata":{"labels":{},"annotations":{}},"relationships":{"organization":{"data":
{"guid":"59b1e651-f4a9-4568-9fe4-f8406ee96cb9"}}}}
```

### What you expected to happen

[...]

PR #564, opened 2022-01-26
Check org accessible before creating space

kieron-dev:
#552

## What is this change about?
Fixes bug where one can pass a space guid as the org guid when creating a space, and end up with a sub-space.

No

## Acceptance Steps
See repro in #552

@cloudfoundry/eirini


danail-branekov (approved):
lgtm

Issue #448, opened 2022-01-12
[Feature]: Developers should only be able to list routes for apps they have access to


gcapizzi:
### Background

**As an** admin
**I want** to assign the `SpaceDeveloper` role to the users I trust
**So that** they can push apps to their spaces

### Acceptance Criteria

**GIVEN** two spaces `space1` and `space2` exist
**AND** `app1` exists in `space1`
**AND** `app2` exists in `space2`
**AND** `user` has a `RoleBinding` to the `SpaceDeveloper` `ClusterRole` in `space1`

1. **WHEN I** [list routes](https://v3-apidocs.cloudfoundry.org/version/3.113.0/index.html#list-routes-for-an-app) for `app1` as `user`
  **THEN I** see it working

1. **WHEN I** list routes for `app2` as `user`
  **THEN I** get the following response:
  ```
  HTTP/1.1 404 Not Found
  Content-Type: application/json
  ```
  ```json
   {
    "errors": [
      {
        "code": 10010,
        "title": "CF-ResourceNotFound",
        "detail": "App not found"
      }
    ]
  }
  ```

PR #565, opened 2022-01-26
Correct authz responses for get app routes

kieron-dev:
#448 

## What is this change about?
Make sure get app routes fails with a 404 when user doesn't have permission to see the app.

No

## Acceptance Steps
See issue #448 

@cloudfoundry/eirini


danail-branekov (approved):
lgtm

PR #566, opened 2022-01-26
Use userClientFactory correctly in repo code

kieron-dev:
No

## What is this change about?
Correct the usage of userClientFactory in the app and buildpack repos

We shouldn't pass the authEnabled flag in, and we should trust that main has injected the correct client factory.

No

## Acceptance Steps
Green CI

@cloudfoundry/cf-k8s


danail-branekov (approved):
lgtm

PR #571, opened 2022-01-27
Do not run ginkgo recursively when invoked from vim

kieron-dev:
No

## What is this change about?
The vim test integration executes `scripts/test` if it exists. We want this to run on the current suite, not any suites in sub-directories.

To enable this, we add another env option `NON_RECUSIVE_TEST` to run-tests.sh, and set this in `script/test` before invoking `run-tests.sh`. When not set, ginkgo is executed with the `-r` flag, otherwise without it.

No

## Acceptance Steps
In vim, with the eirini team's configuration, press `<space>ts` from a test in `api/apis`, and see only the apis suite runs, not the apis/integration suite too.

@cloudfoundry/eirini

### DIFF
--- a/scripts/run-tests.sh
+++ b/scripts/run-tests.sh
@@ -32,4 +32,8 @@ if [[ -n "$GINKGO_NODES" ]]; then
   extra_args+=("--procs=${GINKGO_NODES}")
 fi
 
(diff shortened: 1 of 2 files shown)


danail-branekov (approved):
lgtm

Issue #575, opened 2022-01-27
[Feature]: App Developer can see more detailed stack info with`cf buildpacks`


gnovv:
### Blockers/Dependencies

Follow up to https://github.com/cloudfoundry/cf-k8s-controllers/issues/349

### Background

Today the output looks like:
```
 $ cf buildpacks

position   name                                      stack   enabled   locked   filename
1          tanzu-buildpacks/dotnet-core-lite         base    true      false    tanzu-buildpacks/dotnet-core-lite@1.11.0
2          tanzu-buildpacks/nodejs-lite              base    true      false    tanzu-buildpacks/nodejs-lite@1.13.3
3          tanzu-buildpacks/go-lite                  base    true      false    tanzu-buildpacks/go-lite@1.10.3
4          tanzu-buildpacks/python-lite              base    true      false    tanzu-buildpacks/python-lite@1.1.3
5          tanzu-buildpacks/java-native-image-lite   base    true      false    tanzu-buildpacks/java-native-image-lite@6.3.0
6          tanzu-buildpacks/java-lite                base    true      false    tanzu-buildpacks/java-lite@6.3.0
```

Where the stack column lacks detail: `base`

We want to take the `id` field from the default ClusterBuilder `status.stack`:
```
  stack:
    id: io.buildpacks.stacks.bionic
    runImage: gcr.io/cf-relint-greengrass/cf-k8s-controllers/kpack/beta@sha256:0f5c4717bd2b51e5131e2278dbaa5405fdc64ed65f4d88cb997777225de11818
```

### Acceptance Criteria

**GIVEN** I have cf-on-k8s and all prerequisites installed +  RBAC
**WHEN I** run `cf buildpacks`
**THEN I** see that the output of the stack column matches the `status.stack.id` of the default ClusterBuilder

tcdowney:
Closing: this was addressed by https://github.com/cloudfoundry/cf-k8s-controllers/pull/576

PR #576, opened 2022-01-27
Use Stack ID from Builder's Status when listing Buildpacks

tcdowney:
Issue: https://github.com/cloudfoundry/cf-k8s-controllers/issues/575

## What is this change about?
The Stack/ClusterStack name in a Builder's Spec is often something generic like "default" or "base". This change uses the Stack ID from the Builder's Status which typically includes more meaningful information for developers (typically something like `io.buildpacks.stacks.bionic`) that conveys details about the base OS.

No

## Acceptance Steps
1. Deploy and run `cf buildpacks`. Should see something like this:

```
11:52 $ cf buildpacks
Getting buildpacks as tdowney...

position   name                         stack                         enabled   locked   filename
1          paketo-buildpacks/java       io.buildpacks.stacks.bionic   true      false    paketo-buildpacks/java@6.7.1
2          paketo-buildpacks/go         io.buildpacks.stacks.bionic   true      false    paketo-buildpacks/go@0.13.2
3          paketo-buildpacks/nodejs     io.buildpacks.stacks.bionic   true      false    paketo-buildpacks/nodejs@0.12.0
4          paketo-buildpacks/ruby       io.buildpacks.stacks.bionic   true      false    paketo-buildpacks/ruby@0.10.0
5          paketo-buildpacks/procfile   io.buildpacks.stacks.bionic   true      false    paketo-buildpacks/procfile@5.0.2
```

@gnovv

### DIFF
--- a/api/repositories/buildpack_repository.go
+++ b/api/repositories/buildpack_repository.go
@@ -61,7 +61,7 @@ func clusterBuilderToBuildpackRecords(builder *buildv1alpha2.ClusterBuilder) []B
 		currentRecord := BuildpackRecord{
 			Name:      orderEntry.Group[0].Id,
 			Position:  i + 1,
-			Stack:     builder.Spec.Stack.Name,
+			Stack:     builder.Status.Stack.ID,
 			Version:   orderEntry.Group[0].Version,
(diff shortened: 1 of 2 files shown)


gnovv (approved):
LGTM!

PR #588, opened 2022-01-28
Integration route tests run in serial

Birdrock:
No.

## What is this change about?
There is some test pollution when running integration tests in parallel. They use the same etcd and apiserver, so they can step on each other.

No.

## Acceptance Steps
`make test-api`

@davewalter @akrishna90


Birdrock:
Does not address root cause of test flakiness. Closing for later fix.

PR #594, opened 2022-01-31
Use client.MatchingFields to find resources by name

mnitchev:
No

## What is this change about?
Use client.MatchingFields to find resource by name instead of doing it manually as it is also probably more efficient and we can eventually take advantage of controller-runtime cache.

No

## Acceptance Steps
N/A

@kieron-dev


kieron-dev (approved):
lgtm

Issue #289, opened 2021-11-25
[Chore] Use single payload validator in all handlers


mnitchev:
### Background

The [validator](https://github.com/cloudfoundry/cf-k8s-controllers/blob/e897aeab3117b021ca9b270a6cc7fcd100722152/api/apis/shared.go#L69) we use states that it is designed to be used as a singleton, because it utilizes a cache. We, however, are currently creating it each time we receive a request. This also means that we are setting up the validators each time, but [ignoring the errors](https://github.com/cloudfoundry/cf-k8s-controllers/blob/e897aeab3117b021ca9b270a6cc7fcd100722152/api/apis/shared.go#L74-L96) that this setup might generate. 

### Action to take
Setup the validator in `main.go` (failing fast in case of errors) and inject it into all handlers.

gcapizzi:
See also duplicate #143!

danail-branekov:
The PR above extracts the validator into a singleton that is injected in all handlers. 

We had an idea to drive the refactoring further into introducing some sort of a middleware that performs that payload decoding and validation so that handlers do not have to do it. The main challenge here is type compatibility - an eventual middleware would not know what payload type the next handler would need. We could maintain some mapping between endpoint URL and required type in such a middleware and populate the decoded object into the context but that introduces quite some complexity:
a) the mapping in the middleware and the routes mapping would have to be double-maintained which is quite error-prone
b) when getting the objects from the context, handlers would still have to check whether it has been set from the middleware, which is pretty much the same code as doing the validation error check themselves.

PR #596, opened 2022-01-31
Use single payload validator in all handles

danail-branekov:
https://github.com/cloudfoundry/cf-k8s-controllers/issues/289

## What is this change about?
Use a singleton instance of payload validator instead of creating a new one on every HTTP request thus making use of the singleton validator cache.

No

## Acceptance Steps
Green tests

@mnitchev


davewalter (approved):
LGTM

PR #600, opened 2022-01-31
Fix 415 - Routes with uppercase letters in the host field result in i…

### LINKED ISSUE #415: [Bug]: Routes with uppercase letters in the `host` field result in invalid `HTTPProxy` resources
### What happened?

We pushed an app with a name that contained uppercase letters (e.g. `MyApp`). By default, `cf push` will create a route using the app's name as the `host`. So it created a route like `MyApp.apps.example.com`. This failed during route reconciliation since the `HTTPProxy` was invalid.

```
failed to patch FQDN HTTPProxy    {"error": "HTTPProxy.projectcontour.io \"MyApp.vcap.me\" is invalid: [metadata.name: Invalid value: \"MyApp.vcap.me\": a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'), spec.virtualhost.fqdn: Invalid value: \"MyApp.vcap.me\": spec.virtualhost.fqdn in body should match '^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$']"}
```

### What you expected to happen

We expected the app to be routable.

### Acceptance Criteria

GIVEN I have an app
WHEN I push it with a name containing upper case letters
THEN I see the `CFRoute`/`HTTPProxy` were created successfully with the app's name downcased.

### How to reproduce it (as minimally and precisely as possible)

1. `cf push MyApp`
2. `curl -Lk MyApp.<apps-domain-here>`

---

### Anything else we need to know?

* Does this affect `CFDomain`s as well? Are there any additional validations we should add there?
* The minimal way to fix this is to just update the "default route creation" logic, but we should probably fix it more generally for those who create routes via `POST /v3/routes` as well. Should we add stricter validation to the routes or just downcase the host for users?

---

### Environment

* Revision of codebase: 4f1ff5c95cf737f72e05e1e5a3fc48cf51f3d91a
* Kubernetes version (use `kubectl version`): N/A
* Cloud provider or hardware configuration: N/A
* Others:
  * Contour version: `1.19.1`

### Dev Notes
- For now it is sufficient to modify the Route controller to lowercase all Host inputs when creating HTTPProxies

gnovv:
…nvalid HTTPProxy resources

Co-authored-by: Andrew Costa <ancosta@vmware.com>

Resolves #415

## What is this change about?

Forces lowercase Host in all created FQDN proxies from the RouteController

Yes, it is now possible for multiple CFRoutes to collide when they are different lowercase/uppercase versions of the same host. This is an accepted change since previously the upper-case routes weren't valid at all.

## Acceptance Steps

See issue

@acosta11

### DIFF
--- a/controllers/controllers/networking/cfroute_controller.go
+++ b/controllers/controllers/networking/cfroute_controller.go
@@ -19,11 +19,11 @@ package networking
 import (
 	"context"
 	"fmt"
-
-	"code.cloudfoundry.org/cf-k8s-controllers/controllers/config"
+	"strings"
 
 	networkingv1alpha1 "code.cloudfoundry.org/cf-k8s-controllers/controllers/apis/networking/v1alpha1"
 	workloadsv1alpha1 "code.cloudfoundry.org/cf-k8s-controllers/controllers/apis/workloads/v1alpha1"
+	"code.cloudfoundry.org/cf-k8s-controllers/controllers/config"
 
 	"github.com/go-logr/logr"
 	contourv1 "github.com/projectcontour/contour/apis/projectcontour/v1"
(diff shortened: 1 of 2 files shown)


acosta11 (approved):
LGTM

Issue #613, opened 2022-02-03
[Bug] Upload to the container registry is not authenticated


mnitchev:
While implementing #545 we noticed that even if the Patch that sets the ImageRef on the Package fails with 403 Forbidden, the image is already uploaded to the registry. This means tha
t unauthorized users can still **actually* upload the package even though they don't have permission to do that.

Given that the container registry does not use the Kubernetes RBAC we need to do this authorization ourselves. One way would be to test if the user has Patch permissions on the Packag
e resource before we do the image upload (simillar to how `kubectl auth can-i` works; see [here](https://kubernetes.io/docs/reference/access-authn-authz/authorization/#checking-api-ac
cess))

tcdowney:
Yeah permissions check on the ability to `PATCH`/`PUT` a `Package` before uploading is definitely required. That `kubectl auth can-i` looks cool.

PR #616, opened 2022-02-03
Authenticate package uploading

gcapizzi:

#545 

## What is this change about?

This introduces authentication to the package upload endpoint. Some notes:

* This actually leverages RBAC on package updates, and it does not actually prevent unauthorised users to upload packages. The endpoint will return an error _after_ the upload and leave the package unchanged. We have opened bug #613 to figure out how to address this issue, as access to the container registry is not subject to Kubernetes RBAC.
* This also fixes the error returned when trying to create a package for an app the user doesn't have access to: the correct error is `422 Unprocessable Entity` and not `404 Not Found`.

### DIFF
--- a/api/apis/package_handler.go
+++ b/api/apis/package_handler.go
@@ -168,7 +168,7 @@ func (h PackageHandler) packageCreateHandler(authInfo authorization.Info, w http
 		switch err.(type) {
 		case repositories.PermissionDeniedOrNotFoundError:
 			h.logger.Info("App not found", "App GUID", payload.Relationships.App.Data.GUID)
-			writeNotFoundErrorResponse(w, "App")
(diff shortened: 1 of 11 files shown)


tcdowney:
@gcapizzi why not fold https://github.com/cloudfoundry/cf-k8s-controllers/issues/613 into this story versus making a separate bug? I feel like it's not a bug (cause we know we have no auth) unless this gets merged and muddies the waters about the level of authz in place for this endpoint?

gcapizzi:
@tcdowney we just wanted to do it as a separate piece of work as
* the current technically already passes acceptance, the issue is not user visible
* the solution is not super clear, although the `SelfSubjectAccessReview` seems to be the way

I'm going to turn it into a feature though, as you're right that it doesn't really feel like a bug.

kieron-dev (approved):
lgtm

Issue #587, opened 2022-01-28
[Bug]: App Manifest should support the "none" health check type


tcdowney:
### What happened?

Currently if you set `health-check-type: none` in an App's manifest it will result in a 500 error from the API.

```
14:47 $ cf push dora
Pushing app dora to org o / space s as tdowney...
Applying manifest file /Users/tdowney/go/src/github.com/cloudfoundry/cf-acceptance-tests/assets/dora/manifest.yml...

Updating with these attributes...
  applications:
  - name: dora
    processes:
    - health-check-type: none
      memory: 2GB
      type: web
    default-route: true
Unexpected Response
Response Code: 500
Code: 0, Title: , Detail: {"errors":[{"detail":"An unknown error occurred.","title":"UnknownError","code":10001}]}
```

Looking at the logs on the API Shim you see an error like this:

```
1.643406468157222e+09	ERROR	SpaceManifestHandler	error applying the manifest	{"error": "CFProcess.workloads.cloudfoundry.org \"fe3170e6-ea94-475e-b53c-daeecb97fc9c\" is invalid: spec.healthCheck.type: Unsupported value: \"none\": supported values: \"http\", \"port\", \"process\""}
code.cloudfoundry.org/cf-k8s-controllers/api/apis.(*AuthAwareHandlerFuncWrapper).Wrap.func1
	/workspace/api/apis/auth_aware_handler.go:33
net/http.HandlerFunc.ServeHTTP
	/usr/local/go/src/net/http/server.go:2047
code.cloudfoundry.org/cf-k8s-controllers/api/apis.(*AuthenticationMiddleware).Middleware.func1
	/workspace/api/apis/authentication_middleware.go:76
net/http.HandlerFunc.ServeHTTP
	/usr/local/go/src/net/http/server.go:2047
github.com/gorilla/mux.(*Router).ServeHTTP
	/go/pkg/mod/github.com/gorilla/mux@v1.8.0/mux.go:210
net/http.serverHandler.ServeHTTP
	/usr/local/go/src/net/http/server.go:2879
net/http.(*conn).serve
	/usr/local/go/src/net/http/server.go:1930
```

### What you expected to happen

I expected the `cf push` to succeed and that the process's health check type would ultimately be set to `process` (`none` is an alias of the `process` health check type).

### Acceptance Criteria

GIVEN I have an app with a manifest that looks like
```yaml
name: dora
processes:
  - type: web
    health-check-type: none
```
WHEN I `cf push` the
[...]

PR #620, opened 2022-02-04
Improve health-check-type validation in manifest

tcdowney:
* https://github.com/cloudfoundry/cf-k8s-controllers/issues/587

## What is this change about?
Adds ability to accept the 'none' health check type (a legacy alias of the 'process' type that some older app manifests use) and adds payload validation of health check types so developers will get a 422 instead of a 500 error when they request an invalid type.

Relevant docs: https://v3-apidocs.cloudfoundry.org/version/3.114.0/index.html#space-manifest-process-level-configuration

No

## Acceptance Steps
See https://github.com/cloudfoundry/cf-k8s-controllers/issues/587. Basically just push an app with a manifest that has the `health-check-type` for a process set to `none`.

@akrishna90 (epic anchor)


akrishna90 (approved):
lgtm!

PR #632, opened 2022-02-08
598 Changed process controller to use SerivceBinding indexer

### LINKED ISSUE #598: [Chore]: Ensure ProcessController can reliably fetch all `CFServiceBindings` for an app
### Background

Currently we rely on a specific label being present in the Process Controller when looking up all of the `CFServiceBindings` for an app:

https://github.com/cloudfoundry/cf-k8s-controllers/blob/a0bbd275fbaa7694bba933892e4bdf105d44dc6c/controllers/controllers/workloads/cfprocess_controller.go#L382

There is no webhook or anything enforcing that label exists so it is brittle.

### Action to take

We can either add a mutating webhook to always set the `workloads.cloudfoundry.org/app-guid` label to match the `spec.appRef.name` property, or we can add a field indexer on `spec.appRef.name` and use that for the lookups.

I personally have a preference toward the field indexer approach since this is controller code, but there may be valid reasons to require the label for future API story lookups.

### Impact

Without making either of these changes (webhook or field indexer) users will need to know that they must add the `workloads.cloudfoundry.org/app-guid` for the reconciler to work. 

### Dev Notes

It may be easiest to just do both the webhook and the index (we have done this elsewhere)

gnovv:
Co-authored-by: Akira Wong <wakira@vmware.com>

Resolves #598

## What is this change about?

Changes label lookup for ServiceBinding to use an Indexed field instead

No

## Acceptance Steps

N/A

@matt-royal

### DIFF
--- a/controllers/controllers/shared/index.go
+++ b/controllers/controllers/shared/index.go
@@ -3,15 +3,30 @@ package shared
 import (
 	"context"
 
+	servicesv1alpha1 "code.cloudfoundry.org/cf-k8s-controllers/controllers/apis/services/v1alpha1"
+
 	networkingv1alpha1 "code.cloudfoundry.org/cf-k8s-controllers/controllers/apis/networking/v1alpha1"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
 
-const DestinationAppName = "destinationAppName"
+const (
+	DestinationAppName         = "destinationAppName"
+	ServiceBindingAppGUIDIndex = "serviceBindingAppGUID"
+)
 
 func SetupIndexWithManager(mgr manager.Manager) error {
+	if err := setupRouteIndex(mgr); err != nil {
+		return err
+	}
+	if err := setupServiceBindingIndex(mgr); err != nil {
+		return err
+	}
+	return nil
+}
+
+func setupRouteIndex(mgr manager.Manager) error {
 	// Generate indexes for CFRoute on field spec.Destination.AppRef.Name for efficient querying.
 	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &networkingv1alpha1.CFRoute{}, DestinationAppName,
 		func(rawObj client.Object) []string {
@@ -22,6 +37,18 @@ func SetupIndexWithManager(mgr manager.Manager) error {
 			}
 			return destinationAppNames
 		}); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func setupServiceBindingIndex(mgr manager.Manager) error {
+	// Generate indexes for CFServiceBinding on field serviceBinding.Spec.AppRef.Name for efficient querying.
+	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &servicesv1alpha1.CFServiceBinding{}, ServiceBindingAppGUIDIndex, func(rawObj client.Object) []string {
+		serviceBinding := rawObj.(*servicesv1alpha1.CFServiceBinding)
+		return []string{serviceBinding.Spec.AppRef.Name}
+	}); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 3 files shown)


controllers/controllers/shared/index.go
```diff
@@ -3,15 +3,30 @@ package shared
...
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
 
-const DestinationAppName = "destinationAppName"
+const (
+	DestinationAppName         = "destinationAppName"
+	ServiceBindingAppGUIDIndex = "serviceBindingAppGUID"
+)
```
matt-royal, line 17: It looks like you're introducing a new naming convention for these constants with the `Index` suffix. 2 things about this:

1. For consistency, I'd suggest renaming the existing constant `DestinationAppName` to match the same convention
2. A nice standard practice I've seen in Go with constants is to prefix them all with the same "type" info (e.g `http.StatusOK`, `http.StatusCreated`, etc). Maybe `IndexServiceBindingAppGUID` or `IndexForServiceBindingAppGUID` would be clearer and make it easier to find index constants generally?

controllers/controllers/shared/index.go
```diff
@@ -3,15 +3,30 @@ package shared
...
 
 func SetupIndexWithManager(mgr manager.Manager) error {
+	if err := setupRouteIndex(mgr); err != nil {
+		return err
+	}
+	if err := setupServiceBindingIndex(mgr); err != nil {
+		return err
+	}
+	return nil
```
matt-royal, line 26: This pattern works and is clear, so there's no need to change it. That said, an alternative would be to inline the `IndexField` call here, but move its function argument into its own dedicated function. Something like...

```go
err := mgr.GetFieldIndexer().IndexField(context.Background(), &networkingv1alpha1.CFRoute{}, DestinationAppName, RouteDestinationAppNameIndexFn)
if err != nil {
  return err
} 
```
```go
func RouteDestinationAppNameIndexFn(rawObj client.Object) []string {
			route := rawObj.(*networkingv1alpha1.CFRoute)
			var destinationAppNames []string
			for _, destination := range route.Spec.Destinations {
				destinationAppNames = append(destinationAppNames, destination.AppRef.Name)
			}
			return destinationAppNames
		});
```

controllers/controllers/shared/index.go
```diff
@@ -27,3 +42,15 @@ func SetupIndexWithManager(mgr manager.Manager) error {
 
 	return nil
 }
+
+func setupServiceBindingIndex(mgr manager.Manager) error {
+	// Generate indexes for CFServiceBinding on field serviceBinding.Spec.AppRef.Name for efficient querying.
+	if err := mgr.GetFieldIndexer().IndexField(context.Background(), &servicesv1alpha1.CFServiceBinding{}, ServiceBindingAppGUIDIndex, func(rawObj client.Object) []string {
+		serviceBinding := rawObj.(*servicesv1alpha1.CFServiceBinding)
+		return []string{serviceBinding.Spec.AppRef.Name}
+	}); err != nil {
```
matt-royal, line 51: I know you just copied this from the existing example, but this `err != nil` check is buried really deeply. It would be clearer to move the `err` assignment up one line and just have a simple `if err != nil` check on its own line.

matt-royal (approved):
I have some suggestions on how you could improve this, but none of them are blocking so I'll approve and leave it up to you.

controllers/controllers/shared/index.go
gnovv: Done!
Changed to `IndexServiceBindingAppGUID` and `IndexRouteDestinationAppName`

controllers/controllers/shared/index.go
gnovv: Good idea, didn't notice that those were actually inline functions. Your suggestion makes a ton of sense!

controllers/controllers/shared/index.go
gnovv: Done!

Issue #554, opened 2022-01-25
[Feature]: Developers should only be able to get processes associated to apps they have access to


gcapizzi:
### Background

**As an** admin
**I want** to assign the `SpaceDeveloper` role to the users I trust
**So that** they can push apps to their spaces

### Background

**As an** admin
**I want** to assign the `SpaceDeveloper` role to the users I trust
**So that** they can push apps to their spaces

### Acceptance Criteria

**GIVEN** two spaces `space1` and `space2` exist
**AND** `app1` exists in `space1`
**AND** `process1` exists for `app1`
**AND** `app2` exists in `space2`
**AND** `process2` exists for `app2`
**AND** `user` has a `RoleBinding` to the `SpaceDeveloper` `ClusterRole` in `space1`

1. **WHEN I** [get `process1`](https://v3-apidocs.cloudfoundry.org/version/3.113.0/index.html#get-a-process) as `user`
  **THEN I** see it working

1. **WHEN I** get `process2` as `user`
  **THEN I** get the following response:
  ```
  HTTP/1.1 404 Not Found
  Content-Type: application/json
  ```
  ```json
  {
    "errors": [
      {
        "code": 10010,
        "title": "CF-ResourceNotFound",
        "detail": "Process not found"
      }
    ]
  }
  ```

tcdowney:
@gcapizzi I think I heard during one of the sync calls that y'all have been folding in e2e test work into these stories? Is that true? Should we be doing some of the work for https://github.com/cloudfoundry/cf-k8s-controllers/issues/523 (at least the `GET /v3/processes/:guid` part) as part of this?

Also, just for confirmation, the two endpoints covered by this story are:
* `GET /v3/apps/:guid/processes/:type`
* `GET /v3/processes/:guid`

gcapizzi:
I [doesn't look like](https://github.com/cloudfoundry/cf-k8s-controllers/blob/main/api/apis/app_handler.go#L26-L39) we implement `GET /v3/apps/:guid/processes/:type` yet, so we only need to work on `GET /v3/processes/:guid` for now. 

Yes we want to write end-to-end tests before changing the corresponding endpoint, #523 can be left to a no-op if nothing is left to backfill by the time we pick it up!

gcapizzi:
Dragging this back to _Review in progress_, see [this comment](https://github.com/cloudfoundry/cf-k8s-controllers/pull/633#issuecomment-1036100687) for details!

PR #633, opened 2022-02-09
Add authorization check to get process

davewalter:
#554

## What is this change about?
This PR adds an authorization check to the GET /v3/processes/:guid endpoint.

No

## Acceptance Steps
See #554 

@Birdrock


api/repositories/process_repository.go
```diff
@@ -102,7 +110,27 @@ func (r *ProcessRepo) GetProcess(ctx context.Context, authInfo authorization.Inf
 	if err != nil { // untested
 		return ProcessRecord{}, err
 	}
-	return returnProcess(processList.Items)
```
danail-branekov, line 105: I would suggest renaming `returnProcess` to `ensureSingleProcess` or something similar. `returnProcess` sounds like a `return` statement to me and I would expect no code after it.

api/repositories/process_repository_test.go
```diff
@@ -4,131 +4,142 @@ import (
...
 
-		app1GUID = generateGUID()
-		app2GUID = generateGUID()
-		process1GUID = generateGUID()
-		process2GUID = generateGUID()
+		spaceDeveloperClusterRole = createClusterRole(testCtx, SpaceDeveloperClusterRoleRules)
 	})
 
 	AfterEach(func() {
 		Expect(
```
danail-branekov, line 62: I believe that for env tests cleaning up the namespaces makes no sense, as the test env gets destroyed anyway. Also, as in env tests there are no controllers, deleting the namespace does not delete the objects it contains, they remain as records in the etcd.
TL;DR I don't think this `AfterEach` is needed at all.

api/repositories/process_repository_test.go
```diff
@@ -4,131 +4,142 @@ import (
...
+			cfApp1 = initializeAppCR("test-app1", app1GUID, space1.Name)
+			Expect(k8sClient.Create(context.Background(), cfApp1)).To(Succeed())
+
+			cfApp2 = initializeAppCR("test-app2", app2GUID, space2.Name)
+			Expect(k8sClient.Create(context.Background(), cfApp2)).To(Succeed())
 
-			namespace2Name := generateGUID()
-			namespace2 = &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespace2Name}}
-			Expect(k8sClient.Create(context.Background(), namespace2)).To(Succeed())
+			cfProcess1 = initializeProcessCR(process1GUID, space1.Name, app1GUID)
```
danail-branekov, line 88: Are there usages of the initializeXXXCR functions that are not followed by `k8sClient.Create`? If not, I would suggest putting CR initialisation and its creation into a single function and internally ensures that cration succeeds - that would spare all the creates in tests.

api/tests/e2e/processes_test.go
```diff
@@ -0,0 +1,74 @@
...
+		appGUID     string
+		processGUID string
+		result      resource
+		list        resourceList
+		resp        *resty.Response
+	)
+
+	BeforeEach(func() {
+		orgGUID = createOrg(generateGUID("org"))
+		DeferCleanup(func() {
```
danail-branekov, line 25: I personally find `DeferCleanup` a bit hard to follow as I would not expect cleanup code in `BeforeEach`. I would suggest moving org deletion into a `AfterEach`

api/repositories/process_repository_test.go
```diff
@@ -4,131 +4,142 @@ import (
...
+			Expect(k8sClient.Create(context.Background(), cfProcess2)).To(Succeed())
 		})
 
-		AfterEach(func() {
-			Expect(k8sClient.Delete(context.Background(), namespace2)).To(Succeed())
+		JustBeforeEach(func() {
+			process, getErr = processRepo.GetProcess(testCtx, authInfo, process1GUID)
 		})
 
 		When("on the happy path", func() {
```
danail-branekov, line 99: Now that we have authrorized vs unauthorized case, I would get rid of the `happy path` context and put the `unauthorized` it in the "default" context (as by default one should get authorization error). And then move the "happy path" under `When("the user is a SpaceDeveloper", ....`

api/repositories/process_repository_test.go
```diff
@@ -138,87 +149,56 @@ var _ = Describe("ProcessRepo", func() {
...
+			Expect(k8sClient.Create(context.Background(), cfProcess1)).To(Succeed())
 
-			cfProcess2 = initializeProcessCR(process2GUID, namespace.Name, app1GUID)
-			Expect(k8sClient.Create(testCtx, cfProcess2)).To(Succeed())
+			cfProcess2 = initializeProcessCR(process2GUID, space1.Name, app1GUID)
+			Expect(k8sClient.Create(context.Background(), cfProcess2)).To(Succeed())
 		})
 
 		When("on the happy path", func() {
 			When("spaceGUID is not empty", func() {
```
danail-branekov, line 193: I believe that this is the default case, it should no be inside a dedicated context

api/repositories/process_repository_test.go
```diff
@@ -138,87 +149,56 @@ var _ = Describe("ProcessRepo", func() {
...
 				It("returns Process records for the AppGUID we request", func() {
-					processes, err := processRepo.ListProcesses(testCtx, authInfo, ListProcessesMessage{AppGUID: []string{app1GUID}, SpaceGUID: namespace.Name})
+					processes, err := processRepo.ListProcesses(testCtx, authInfo, ListProcessesMessage{AppGUID: []string{app1GUID}, SpaceGUID: space1.Name})
```
danail-branekov, line 195: I believe it would more Ginkgo-idiomatic to put the function under test (i.e. `List`) in `JustBeforeEach` instead of calling it in every `It`

danail-branekov:
It seems that PR https://github.com/cloudfoundry/cf-k8s-controllers/pull/637 is conflicting a lot with this one. I am afraid that there would be lots of merge conflicts :(

danail-branekov:
Hey @davewalter 
I have rebased the pr on top of main for you. It is unfair to make you do that on a cross-team collaboration story. 
Hopefully next time we pick a cross-team story for you we would be careful not to heavily modify the code the story is all about so that we do not step on each others' toes.

I think that most of my comments have been already addressed with PR #637 , with the exception of the discussion on dot importing non-ginkgo packages.

api/repositories/process_repository_test.go
davewalter: We've created a separate chore to address this across the entire set of repository tests: #647

julian-hj:
Thanks @kieron-dev! I see that @gcapizzi has already re-opened the github issue. We'll take another crack at it in a separate PR.

Issue #617, opened 2022-02-03
[Feature]: App Developer can delete user-provided service instances via `cf delete-service`


tcdowney:
### Background

**As an** App Developer
**I want** to be able to delete user-provided service instances (UPSIs)
**So that** I can get rid of the ones I no longer need

### Set up notes:

I created the service instance via the CLI like this:
```
cf cups my-upsi -p '{"key": "val"}'
```

---

### Instance Exists (Happy Path: CLI)
**GIVEN** I have an existing user-provided `CFServiceInstance` named `my-upsi`
**WHEN I** call `cf delete-service my-upsi -f`
**THEN I** see the command succeeds and the `CFServiceInstance` and corresponding `Secret` no longer exist on my cluster. CLI output should look similar to:

```
Deleting service instance my-upsi in org org / space space as admin...

Service instance my-upsi deleted.
OK
```

---

### Instance Exists (Happy Path: API)
**GIVEN** I have an existing user-provided `CFServiceInstance` named `my-upsi` with `18c7ca71-6fb7-4c0c-b028-bc5abe25613f` as its guid
**WHEN I** make the following request

```
cf curl -X DELETE /v3/service_instances/18c7ca71-6fb7-4c0c-b028-bc5abe25613f
```

**THEN I** see the the following response

```json
HTTP/1.1 204 No Content

{}
```

---

### Instance Does Not Exist / Not Visible
**GIVEN** I had already deleted the `my-upsi` service instance
**WHEN I** call `cf delete-service my-upsi -f`
**THEN I** see that the CLI command fails with output that looks like this

```
Deleting service instance my-upsi in org org / space space as admin...

Service instance my-upsi did not exist.
OK
```

In this case the API should return a `404` with a response like the following:

```json
{
  "errors": [
    {
      "code": 10010,
      "detail": "Service instance not found",
      "title": "CF-ResourceNotFound"
    }
  ]
}
```

---

### User Can View But Cannot Delete

**GIVEN** I the `my-upsi` service instance exists but I lack permission to `DELETE` it
**WHEN I** call `cf delete-service my-upsi -f`
**THEN I** see that the CLI command fails with output that looks like this

```
You are not authorized to perform the requested action
FAILED
```

In this case the API should return a `403` with a response like the following:

```json
{
  "errors": [
    {
      "code": 10003,
      "detail": "You are not authorized to perform the requested action",
      "title": "CF-NotAuthorized"
    }
  ]
}
```

[...]

PR #634, opened 2022-02-09
Add the DELETE /v3/service_instances endpoint

mnitchev:
#617 

## What is this change about?
Implements delete for service instances.

No

## Acceptance Steps
See issue #617 

@cloudfoundry/eirini

### DIFF
--- a/api/apis/service_instance_handler.go
+++ b/api/apis/service_instance_handler.go
@@ -196,12 +196,10 @@ func (h *ServiceInstanceHandler) serviceInstanceDeleteHandler(authInfo authoriza
 		if repositories.IsForbiddenError(err) {
 			h.logger.Error(err, "user not allowed to get service instance")
 			writeNotFoundErrorResponse(w, repositories.ServiceInstanceResourceType)
-
 			return
 		}
 
-		h.logger.Error(err, "failed to get service instance")
-		writeUnknownErrorResponse(w)
+		handleRepoErrors(h.logger, err, repositories.ServiceInstanceResourceType, serviceInstanceGUID, w)
 		return
 	}
 
@@ -210,15 +208,8 @@ func (h *ServiceInstanceHandler) serviceInstanceDeleteHandler(authInfo authoriza
 		SpaceGUID: serviceInstance.SpaceGUID,
 	})
 	if err != nil {
-		if repositories.IsForbiddenError(err) {
-			h.logger.Error(err, "user not allowed to delete service instance")
-			writeNotAuthorizedErrorResponse(w)
-
-			return
-		}
-
-		h.logger.Error(err, "failed to delete service instance")
-		writeUnknownErrorResponse(w)
+		h.logger.Error(err, "error when deleting service instance", "guid", serviceInstanceGUID)
+		handleRepoErrorsOnWrite(h.logger, err, repositories.ServiceInstanceResourceType, serviceInstanceGUID, w)
 		return
 	}
 
--- a/api/apis/service_instance_handler.go
+++ b/api/apis/service_instance_handler.go
@@ -24,6 +24,7 @@ import (
 const (
 	ServiceInstanceCreateEndpoint = "/v3/service_instances"
 	ServiceInstanceListEndpoint   = "/v3/service_instances"
+	ServiceInstanceDeleteEndpoint = "/v3/service_instances/{guid}"
 )
 
 //counterfeiter:generate -o fake -fake-name CFServiceInstanceRepository . CFServiceInstanceRepository
@@ -31,6 +32,7 @@ type CFServiceInstanceRepository interface {
 	CreateServiceInstance(context.Context, authorization.Info, repositories.CreateServiceInstanceMessage) (repositories.ServiceInstanceRecord, error)
 	ListServiceInstances(context.Context, authorization.Info, repositories.ListServiceInstanceMessage) ([]repositories.ServiceInstanceRecord, error)
 	GetServiceInstance(context.Context, authorization.Info, string) (repositories.ServiceInstanceRecord, error)
+	DeleteServiceInstance(context.Context, authorization.Info, repositories.DeleteServiceInstanceMessage) error
 }
 
 type ServiceInstanceHandler struct {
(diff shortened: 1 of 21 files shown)


api/repositories/service_instance_repository.go
```diff
@@ -134,6 +143,56 @@ func (r *ServiceInstanceRepo) ListServiceInstances(ctx context.Context, authInfo
...
+
+	record, err := returnServiceInstance(serviceInstanceList.Items)
+	if err != nil {
+		return ServiceInstanceRecord{}, err
+	}
+
+	err = userClient.Get(ctx, client.ObjectKey{Namespace: record.SpaceGUID, Name: record.GUID}, &servicesv1alpha1.CFServiceInstance{})
+	if k8serrors.IsForbidden(err) {
+		return ServiceInstanceRecord{}, NewForbiddenError(err)
+	}
```
tcdowney, line 166: This is an example of what we discussed a few weeks back right? Since we don't know what K8s namespace the instance lives in we have to list so we're optimizing by doing that with the privileged client since it is able to maintain a cache.

I think that's fine and good, but I think the way we're doing it here is risky since we're actually using the return value of the privileged client call and only using the userClient for authz.

I think it would be safer if we refactored the privileged client actions into a function (like `privilegedLookupNamespaceForServiceInstance` (I'm sure y'all can come up with a better name 😅 )) that only returns the namespace and then we rely on the userClient for actually fetching the returned object.

api/repositories/service_instance_repository_test.go
```diff
@@ -428,10 +422,134 @@ var _ = Describe("ServiceInstanceRepository", func() {
...
+			createServiceInstance(space2.Name, "some-other-service-instance")
+
+			getGUID = serviceInstance.Name
+		})
+
+		JustBeforeEach(func() {
+			record, getErr = serviceInstanceRepo.GetServiceInstance(testCtx, authInfo, getGUID)
+		})
+
+		It("returns a forbidden error", func() {
```
tcdowney, line 448: Nit: Even though there is no additional set up needed, I think it would improve clarity if we wrapped this in a `When("the user does not have permissions to get the service instance", ...`

api/repositories/service_instance_repository.go
tcdowney: I know there's test that verify the insufficient permissions case, but I think this will both provide additional context for what is going on (via the function name) and give us a little more safety in the future.

tcdowney (changes requested):
Overall looks good. I would like to see some changes to the privileged vs user client calls in the repo though. Thanks!

api/repositories/service_instance_repository_test.go
danail-branekov: I would strongly vote against such contexts as they do not contribute to making the test environment more specific, instead they are (ab)used for documentation purposes. If documenting user not being allowed to do that is really desired, I would prefer putting that into the `It` description, e.g. `It("returns a forbidden error (as the user is not allowed to do that)....`

api/apis/service_instance_handler.go
```diff
@@ -183,8 +186,48 @@ func (h *ServiceInstanceHandler) serviceInstanceListHandler(authInfo authorizati
...
+
+			return
+		}
+
+		h.logger.Error(err, "failed to get service instance")
+		writeUnknownErrorResponse(w)
+		return
+	}
+
+	err = h.serviceInstanceRepo.DeleteServiceInstance(ctx, authInfo, repositories.DeleteServiceInstanceMessage{
```
danail-branekov, line 208: Generally, the `Get-Delete` pattern when dealing with K8S is racy. I wonder whether we can implement the handler just call the repository's `Delete` method and handle an eventual not found error appropriately. The only thing the service instance found is used for is to get the space GUID, but on the other hand, the service instance GUID should be unique and therefore sufficient to locate and delete it.

api/repositories/service_instance_repository.go
danail-branekov: +1 for returning a record for the object being found by the unprivileged client

api/tests/e2e/service_instances_test.go
```diff
@@ -0,0 +1,74 @@
...
+	"net/http"
+
+	"github.com/go-resty/resty/v2"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gstruct"
+	rbacv1 "k8s.io/api/rbac/v1"
+)
+
+var _ = Describe("Service Instances", func() {
```
danail-branekov, line 13: As this PR introduces a `get service` endpoint, there should be an e2e test for it as well (provided that we need `Get` at all, see my previous comments)

api/repositories/service_instance_repository_test.go
tcdowney: 🤷 

Maybe it's due to my background in the Ruby community, but the most common style guides recommend this. I read  code better when I can scan up and down through the contexts to see what is being tested, but I see how this can be a personal preference.

What I'm suggesting is what is recommended by these guidelines and I generally agree:
* https://www.betterspecs.org/#contexts
* https://www.betterspecs.org/#short

That said, I also don't like to see deeply nested contexts for no reason, so it's definitely a judgement call. I don't think what I suggested _needs_ to be done in this case, but I also wouldn't want to see a hard rule against what I'm suggesting because I personally find it better. 🙂

georgethebeatle: The closes example that I can think of is [this](https://medium.com/swlh/effective-ginkgo-gomega-b6c28d476a09#cb3d) by one of Ginkgo's maintainers. I would suggest that you put a comment in the ginkgo style guide proposal, since it sounds related

tcdowney: Yeah I added a comment on the proposal yesterday 👍 

My interpretation of that blog post is that it's centered around the role of `BeforeEach` and less around using `Describe` and `Context` to provide test clarity.

Overall my thinking here though is that it's not a huge deal either way and I don't think there is much value in policing this issue across the whole codebase. I think either way has the potential to result in readable and functioning code so it's ok for something like this to differ based on who wrote the test.

I think on the proposal I agreed with the particular example that something like `When("on the happy path", ...` is not helpful and should be avoided, but that if someone adds a test and has something like `When("the user is unauthorized", ...` and does not need to provide additional setup because it's a lack of setup then that should be perfectly fine to do so. On the flipside I see that others disagree so I retract my suggestion here of asking for it to be done.

api/tests/e2e/service_instances_test.go
georgethebeatle: We are only introducing the ability to get a service instance on the repo level. There is no endpoint for getting the service instance on the handler level, so it is not possible to add an e2e test for this.

tcdowney: I don't think we'll be adding that endpoint soon (`GET /v3/service_instances/:guid`) since the CLI is always fetching these via list calls (filtered by `name` and `space_guid`) -- at least for the CLI commands we're targeting right now.

tcdowney (approved):
LGTM once conflicts are resolved. The namespaceGetting implementation might need to eventually be reconciled with the changes suggested in [this draft PR](https://github.com/cloudfoundry/cf-k8s-controllers/pull/689), but that's still a draft and gathering opinions.

api/apis/service_instance_handler.go
georgethebeatle: That is a very good point @tcdowney I will address it today. This is also a very good example of why we need to handle our errors in a more generic way to avoid overlooking corner cases like this one in future.

tcdowney (approved):
LGTM w/ latest changes and conflicts fixed.

Issue #653, opened 2022-02-11
Handle all unhandled errors in production code


georgethebeatle:
api/apis/log_cache_handler.go
api/apis/process_handler.go
api/apis/resource_matches_handler.go
api/apis/root_v3_handler.go
api/apis/space_manifest_handler.go
api/payloads/manifest.go
controllers/controllers/workloads/cfprocess_controller.go

PR #655, opened 2022-02-11
Handle unhandled errors

georgethebeatle:
#653 

## What is this change about?
Handle a couple of unhandled errors in the production code

No

## Acceptance Steps
Tests are green

@cloudfoundry/cf-k8s

### DIFF
--- a/api/payloads/manifest.go
+++ b/api/payloads/manifest.go
@@ -89,11 +89,13 @@ func (p ManifestApplicationProcess) ToProcessCreateMessage(appGUID, spaceGUID st
 
 	diskQuotaMB = uint64(1024)
 	if p.DiskQuota != nil {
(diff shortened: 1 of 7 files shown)


api/payloads/manifest.go
```diff
@@ -89,11 +89,13 @@ func (p ManifestApplicationProcess) ToProcessCreateMessage(appGUID, spaceGUID st
 
 	diskQuotaMB = uint64(1024)
 	if p.DiskQuota != nil {
+		// error ignored intentionally, since the manifest yaml is validated in handlers
```
matt-royal, line 92: I'm generally anti-comment, but I actually think this one and the other that was removed below are useful and should be kept.

matt-royal (commented):
These changes look good, aside from the context comments that were deleted. I'm going to add those back in main until we can find another way to capture that information without a comment.

Issue #535, opened 2022-01-24
[Feature] App Developer/Operator can set health check with `cf set-health-check`


akrishna90:
### Blockers/Dependencies

NA

### Background

**As a** App Developer/Operator
**I want** to run [`cf set-health-check APP_NAME`](https://cli.cloudfoundry.org/en-US/v7/set-health-check.html)
**So that** I can configure health check for my application

To support the CF CLI command, 

-  Implement `GET /v3/apps/:guid/processes/:web` endpoint - [V3 API documentation](https://v3-apidocs.cloudfoundry.org/version/3.113.0/#get-a-process)

    CF CLI output 
    ```
    REQUEST: [2022-01-23T17:59:46-08:00]
    GET /v3/apps/4589e494-56f3-4536-86db-1ae140d86fe9/processes/web HTTP/1.1
    Host: localhost
    Accept: application/json
    Authorization: [PRIVATE DATA HIDDEN]
    Content-Type: application/json
    User-Agent: cf/8.1.0+034d929d7.2022-01-03 (go1.16.6; amd64 darwin)
    [application/json Content Hidden]
    
    RESPONSE: [2022-01-23T17:59:46-08:00]
    HTTP/1.1 404 Not Found
    Content-Length: 19
    Content-Type: text/plain; charset=utf-8
    Date: Mon, 24 Jan 2022 01:59:46 GMT
    Server: envoy
    X-Content-Type-Options: nosniff
    X-Envoy-Upstream-Service-Time: 0
    404


    Error unmarshalling the following into a cloud controller error: 404 page not found
    ```

- Implement `PATCH /v3/processes/:guid` endpoint - [V3 API documentation](https://v3-apidocs.cloudfoundry.org/version/3.113.0/#update-a-process)

    CF CLI output
    ```
    REQUEST: [2022-01-23T18:10:07-08:00]
    PATCH /v3/processes/4589e494-56f3-4536-86db-1ae140d86fe9 HTTP/1.1
    Host: localhost
    Accept: application/json
    Authorization: [PRIVATE DATA HIDDEN]
    Content-Type: application/json
    User-Agent: cf/8.1.0+034d929d7.2022-01-03 (go1.16.6; amd64 darwin)
    {
      "health_check": {
        "data": {
          "invocation_timeout": 2
        },
        "type": "port"
      }
    }
    
    
    RESPONSE: [2022-01-23T18:10:07-08:00]
    HTTP/1.1 405 Method Not Allowed
[...]

PR #657, opened 2022-02-11
As a developer I want to set health check using CFSetHealthCheck

clintyoshimura:

#535 

## What is this change about?

As a developer I want to set health check using CFSetHealthCheck
- Implemented the patch process endpoint
- Implemented the get app process by type endpoint

no

## Acceptance Steps

Follow acceptance steps from issue

@akrishna90


akrishna90 (approved):
Paired with @clintyoshimura, lgtm!

PR #667, opened 2022-02-14
Unauthorized errors from k8s translates to NotAuthenticatedError

akrishna90:

No.

## What is this change about?

An UnauthorizedError for the underlying k8s comes as a result of user not being authenticated and we believe this needs to be translated as a NotAuthenticatedError.

No

## Acceptance Steps

`make tests` should pass

@davewalter

### DIFF
--- a/api/repositories/k8s_client.go
+++ b/api/repositories/k8s_client.go
@@ -48,7 +48,7 @@ func (f UnprivilegedClientFactory) BuildClient(authInfo authorization.Info) (cli
 		config.KeyData = pem.EncodeToMemory(keyBlock)
 
 	default:
-		return nil, authorization.NotAuthenticatedError{}
+		return nil, authorization.InvalidAuthError{}
 	}
 
 	// This does an API call within the controller-runtime code and is
@@ -57,7 +57,7 @@ func (f UnprivilegedClientFactory) BuildClient(authInfo authorization.Info) (cli
 	userClient, err := client.NewWithWatch(config, client.Options{})
 	if err != nil {
 		if k8serrors.IsUnauthorized(err) {
-			return nil, authorization.InvalidAuthError{}
+			return nil, authorization.NotAuthenticatedError{}
 		}
 		return nil, err
 	}


api/repositories/k8s_client.go
```diff
@@ -48,7 +48,7 @@ func (f UnprivilegedClientFactory) BuildClient(authInfo authorization.Info) (cli
 		config.KeyData = pem.EncodeToMemory(keyBlock)
 
 	default:
-		return nil, authorization.NotAuthenticatedError{}
```
kieron-dev, line 51: We're dealing with the `authorization.Info` object here. It has been created from [parsing the `Authorization` header in the middleware](https://github.com/cloudfoundry/cf-k8s-controllers/blob/d97da7815569d4e8f2c4a753bb70160b93c984c0/api/apis/authentication_middleware.go#L45). The logic there is if the user does not supply a `Bearer` or `ClientCert` in the `Authorization` header, then no accepted authentication method has been supplied, hence a `NotAuthenticatedError` is returned. But if they do supply one, but it isn't the correct format, an `InvalidAuthError` is returned. Note that these names match the user errors returned by the API.

The logic here duplicates that in the middleware somewhat. If the `authInfo` is not using either the bearer or the cert scheme, then no authentication has been passed, yet we are here because we need authentication, hence a `NotAuthenticatedError` is appropriate, and matches the API error returned to the user.

If we get past this point, but the supplied token or cert is not valid when checked by k8s, we return an `InvalidAuthError`.

So I don't recommend making the change you are suggesting here.

api/repositories/k8s_client.go
```diff
@@ -57,7 +57,7 @@ func (f UnprivilegedClientFactory) BuildClient(authInfo authorization.Info) (cli
 	userClient, err := client.NewWithWatch(config, client.Options{})
 	if err != nil {
 		if k8serrors.IsUnauthorized(err) {
-			return nil, authorization.InvalidAuthError{}
+			return nil, authorization.NotAuthenticatedError{}
```
tcdowney, line 60: Just noting from a convo we had during our sync that the Kubernetes errors are named for the HTTP status codes `401 Unauthorized` and `403 Forbidden` which we use to represent "not authenticated" and "not authorized" respectively. Interesting! 🤓 

* IsUnauthorized: https://github.com/kubernetes/apimachinery/blob/276a8a7530a30f42c82b1f0fa2261a905130bf97/pkg/api/errors/errors.go#L664
* IsForbidden: https://github.com/kubernetes/apimachinery/blob/276a8a7530a30f42c82b1f0fa2261a905130bf97/pkg/api/errors/errors.go#L678

api/repositories/k8s_client.go
georgethebeatle: I tend to agree with @kieron-dev's comment above, but that only means that the `authorization.InfoParser` used by the middleware should only ever return `NotAuthenticatedError`, which is [not currently the case](https://github.com/cloudfoundry/cf-k8s-controllers/blob/e05f2ca5c210e539249747fe497d1227a8e47674/api/authorization/info_parser.go#L35)

api/repositories/k8s_client.go
georgethebeatle: In the light of mimicking the CF API as closely as possible I think I favour this change. Here is why:
- As @tcdowney pointed out in this [thread](https://cloudfoundry.slack.com/archives/C0297673ASK/p1644946622354189) the Cloud Controller would only ever return a `403` error if the token is valid, has not yet expired and the user described by this token has insufficient permissions to perform the action.
- The `BuildClient` method is usually being called from a repo, meaning that the auth middleware was already able to parse the token (i.e. it is valid), but k8s rejected it with a `401`. Having a valid token rejected by k8s can mean two things - the token has expired, the token's issuer is not trusted by the cluster. The middleware only guarantees that the token is valid and well formed, not that it is issued by a trusted issuer. I think this case is similar to the token having expired or not having been issued by the correct UAA in which case the right error to return should be `401` or `NotAuthenticatedError`

api/repositories/k8s_client.go
georgethebeatle: Just to make myself clear I think in this specific instance the error should continue being a `NotAuthenticatedError`

api/repositories/k8s_client.go
georgethebeatle: In this case I think changing the error type to a `NotAuthenticatedError` is justified

georgethebeatle (commented):
As indicated in the inline comments I think that only one error type should be changed. I believe this way we will be closer to the behaviour of the CF API. Thanks @akrishna90 @davewalter for starting this discussion

georgethebeatle:
I feel like this PR has become stale, but it served the very important purpose of starting a conversation about problems with error  handling in the current code base. This conversation has become the basis for this [draft PR](https://github.com/cloudfoundry/cf-k8s-controllers/pull/695) exploring a better approach of error handling in general. I will therefore close this PR and invite the authors to review and comment on the linked one. If you find that there are concerns that are not addressed in the linkes PR, please comment no it, or even reopen this one if it makes more sense.

Issue #497, opened 2022-01-19
[Chore]: Backfill tests for `/v3/organizations`


georgethebeatle:
### Background

We need to backfill tests for this endpoint

### Action to take

- `When` I DELETE /v3/organizations/some-org `It` the org namespace is deleted
- `When` an nonexistent org is deleted `It` returns a NotFound error

PR #670, opened 2022-02-15
Backfill org tests

### LINKED ISSUE #615: [Bug]: Admin user lacks rights to delete orgs and spaces
### What happened?

With RBAC enabled, I cannot delete orgs or spaces through the cf API

```bash
$ cf delete-org 03
Really delete the org 03, including its spaces, apps, service instances, routes, private domains and space-scoped service brokers? [yN]: y
Deleting org 03 as admin...
You are not authorized to perform the requested action
```

### What you expected to happen

I should be able to delete a space or org that I just created.

### Acceptance Criteria

GIVEN a newly installed CF, with `authEnabled: true` and an org created by `admin` with `cf create-org o`
WHEN I delete the org as `admin` with `cf delete-org o`
THEN I see that the org is deleted and I do not get an error response.

### How to reproduce it (as minimally and precisely as possible)

Follow the acceptance steps above

### Anything else we need to know?

I suspect this is just a missing permission in the cf_admin role [here](https://github.com/cloudfoundry/cf-k8s-controllers/blob/main/controllers/config/cf_roles/cf_admin.yaml#L92-L93)

Adding `delete` and `get` fixes this issue. I am not sure yet if `get` is even necessary.

### Environment

Revision of codebase: [be5ea65](https://github.com/cloudfoundry/cf-k8s-controllers/commit/be5ea6580bb929baca7d4cae7722520e04c5d46a)
Kubernetes version (use `kubectl version`): 1.21
Cloud provider or hardware configuration: kind
Others: `authEnabled: true`

kieron-dev:
#497 

## What is this change about?
Add acceptance tests to cover org deletion and the recently fixed #615 

This also added a gomega matcher for resty response headers used in the new acceptance test.

While checking the error case, it seemed better to assert against a `cfErr` object, so the other acceptance tests have been modified to use this more concise pattern.

The delete org method in the org repo has been modified to remove the redundant authEnabled check, which is already performed in `main()`, and the repo tests adjusted to reflect the fact that only cf admin users can delete orgs. The cf org manager role has not yet been implemented, so space deletion is also restricted to cf admin users.

No

## Acceptance Steps
Green CI

@cloudfoundry/cf-k8s


davewalter (approved):
Thanks for taking the time to clean this up.

PR #672, opened 2022-02-15
Add RBAC to POST Route Add Destination

### LINKED ISSUE #585: [Feature]: Developers should only be able to add destinations to routes they have access to
### Blockers/Dependencies

_No response_

### Background

**As an** admin
**I want** to assign the `SpaceDeveloper` role to the users I trust
**So that** they can push apps to their spaces

### Acceptance Criteria

**GIVEN** two spaces `space1` and `space2` exist
**AND** `route1` exists in `space1` (e.g. as the default route for an app)
**AND** `route2` exists in `space2`
**AND** `user` has a `RoleBinding` to the `SpaceDeveloper` `ClusterRole` in `space1`
* **WHEN I** [add a destination](https://v3-apidocs.cloudfoundry.org/version/3.113.0/index.html#insert-destinations-for-a-route) to `route1` as `user`
  **THEN I** see it working
* **WHEN I** add a destination to `route2` as `user`
  **THEN I** get the following response:
```
HTTP/1.1 403 Forbidden
```
```json
{
  "errors": [
    {
      "code": 10003,
      "title": "CF-NotAuthorized",
      "detail": "You are not authorized to perform the requested action"
    }
  ]
}
```

### Dev Notes

_No response_

gnovv:

resolves #585 

## What is this change about?

#585
- Implements repo level RBAC w/ user client for GetRoute and AddDestination
- Updates SpaceDeveloper ClusterRole to add "patch" routes permission

Co-authored-by: Andrew Wittrock <awittrock@vmware.com>
Co-authored-by: Julian Hjortshoj <hjortshojj@vmware.com>

No

## Acceptance Steps

See issue.

@julian-hj @Birdrock @gcapizzi @kieron-dev


kieron-dev (dismissed):
Sorry for taking so long to look at this. We've been rather short staffed this week.

I've just pointed out a few little things to change, and you've seen this also needs some changes after the changes to the error constructors.

But overall looks good. Ping me when you update it and I'll have another look.

api/apis/integration/route_test.go
```diff
@@ -327,4 +301,118 @@ var _ = Describe("Route Handler", func() {
...
+			BeforeEach(func() {
+				createRoleBinding(ctx, userName, spaceManagerRole.Name, namespace.Name)
+			})
+
+			It("returns a forbidden error", func() {
+				Expect(rr).To(HaveHTTPStatus(http.StatusForbidden))
+			})
+		})
+
+		When("the user is a space developer", func() {
```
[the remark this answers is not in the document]
julian-hj: I am not super enthusiastic about the approach where we test the happy path only in e2e and then test only failure cases in the integration tests. I don't feel like this represents a meaningful de-duplication, because the majority of code & effort is spent in test setup, so just removing the happy path test doesn't buy much. I also feel like without the happy path test in the integration suite, we don't have assurance that we've written the test setup correctly, so even though the failure case fails, we don't know for sure that it is failing for the reason we intended.

I think we are planning some meetings/discussions about overall testing strategies. I wonder if more changes here might wait until we have agreed on an overall approach for the e2e versus integration tests?

api/apis/integration/route_test.go
```diff
@@ -327,4 +301,118 @@ var _ = Describe("Route Handler", func() {
...
+			BeforeEach(func() {
+				createRoleBinding(ctx, userName, spaceDeveloperRole.Name, namespace.Name)
+			})
+
+			It("updates the route", func() {
+				Expect(rr).To(HaveHTTPStatus(http.StatusOK))
+				Expect(rr).To(HaveHTTPBody(ContainSubstring(newDestinationAppGUID)), rr.Body.String())
+
+				// TODO: add an eventually against K8s client?
```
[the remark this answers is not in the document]
gnovv: Thank you! I filled in the test that was missing there

api/apis/integration/route_test.go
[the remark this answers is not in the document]
kieron-dev: Yep - fine with leaving the happy path in the integration test.

I'm more concerned that we don't have the e2e test. That would give us confidence that the cluster role yamls are correct, and the components behave as expected when using the wiring from main.go.

api/repositories/route_repository.go
```diff
@@ -141,6 +140,24 @@ func (f *RouteRepo) GetRoute(ctx context.Context, authInfo authorization.Info, r
...
+
+	err = userClient.Get(ctx, client.ObjectKey{Namespace: toReturn.SpaceGUID, Name: toReturn.GUID}, &networkingv1alpha1.CFRoute{})
+
+	if err != nil {
+		if apierrors.IsForbidden(err) {
+			return RouteRecord{}, NewForbiddenError(RouteResourceType, err)
+		}
+		return RouteRecord{}, fmt.Errorf("get-route user client get failed: %w", err) // untested
+	}
+
```
gcapizzi, line 160: It looks like between this story and #512 we have ended up implementing #640 too: was this a problem with the order of the stories being wrong or stories not being sufficiently decoupled? If not, we should be careful not to introduce too many changes outside the scope of the story.

api/repositories/route_repository.go
```diff
@@ -381,9 +398,17 @@ func (f *RouteRepo) AddDestinationsToRoute(ctx context.Context, authInfo authori
...
+	if err != nil {
+		return RouteRecord{}, fmt.Errorf("failed to build user client: %w", err)
+	}
+
+	err = userClient.Patch(ctx, cfRoute, client.MergeFrom(baseCFRoute))
+	if err != nil {
+		if apierrors.IsForbidden(err) {
+			return RouteRecord{}, NewForbiddenError(RouteResourceType, err)
+		}
+		return RouteRecord{}, fmt.Errorf("err in client.Patch: %w", err) // untested
```
gcapizzi, line 411: It looks like none of the repository changes are being tested in the `route_repository_test`. While we're covered by the higher level integration tests, I'd still like to see that test covering the whole functionality of the repository.

gcapizzi (changes requested):
I have added my comments inline, they're mostly about test coverage.

api/repositories/route_repository.go
julian-hj: We think this was a problem of ordering--in order to enforce access to a route before modifying it, it made most sense to add access checks to the get method in the route repository (which is also called from the destinations endpoint.) Avoiding this overlap would have required us to do unnatural things in the implementation.

gcapizzi: right, makes sense - maybe it makes sense to always prioritise `GET`s on every specific resource, it looks like it often ends up overlapping with the other endpoints

matt-royal:
@gcapizzi and @kieron-dev When writing the e2e tests for this we found a RBAC problem related to CFDomains. In order to list or get them, the user needs to have a rolebinding in the root namespace that grants those abilities. We did a minimal effort fix in the latest commit, but the UX isn't great. Operators will have to manually create role bindings for every user in the root namespace.

How do you think we should solve this? One option would be to change the Add Role endpoint so that it implicitly creates the root namespace role binding when adding the `org user` role. We could also add a new role called `cf user`. I'm sure there are other ways as well, and I'd love your input. That work is large enough that I think we should write follow up stories instead of rolling it in here.

matt-royal:
@gcapizzi and @kieron-dev When writing the e2e tests for this we found a RBAC problem related to CFDomains. In order to list or get them, the user needs to have a rolebinding in the root namespace that grants those abilities. We did a minimal effort fix in the latest commit, but the UX isn't great. Operators will have to manually create role bindings for every user in the root namespace.

How do you think we should solve this? One option would be to change the Add Role endpoint so that it implicitly creates the root namespace role binding when adding the `org user` role. We could also add a new role called `cf user`. I'm sure there are other ways as well, and I'd love your input. That work is large enough that I think we should write follow up stories instead of rolling it in here.

kieron-dev:
@matt-royal - we've added some ideas to your thread on cf-k8s-dev: https://cloudfoundry.slack.com/archives/C0297673ASK/p1645779601613699

matt-royal (approved):
This PR has changed hands a few times. I ended up addressing the changes requested in earlier reviews with a few different pairs. I asked @gnovv to give this another look over before I merge it, but he is the author so he can't "approve" it. This review is just to check the "approved" box.

matt-royal:
@kieron-dev When merging my changes with your upstream changes, I realized that we'd solved the Domain permission issue differently. My pairs and I had added a new role to the deployment whereas you'd just created a temporary one in the test e2e test where it was needed. Since we depended on the new ClusterRole in the api integration tests, I decided to remove your smaller hack and lean into using the ClusterRole. We can pull it out once we've solved this problem properly. I'm also happy to change this again once it's merged, but I've been in rebase/merge hell with this commit so for now I'm just going to merge it.

PR #676, opened 2022-02-16
Refactor repositories errors

kieron-dev:
No

## What is this change about?
Improve the NotFoundError and ForbiddenError in repositories:
- Make internals private
- Use constructors for creation throughout
- Remove duplication between NotFound and Forbidden errors
- Backfill missing resource types in code and tests

No

## Acceptance Steps
Green CI

@cloudfoundry/cf-k8s


danail-branekov (approved):

kieron-dev:
We've just copied the ResourceTypes strings here. It would be better if these were constants.

I think we should do another change where we replace all those magic strings with new constants defined in the repositories package.

Issue #603, opened 2022-01-31
[Feature]: API Client can list service bindings via `GET /v3/service_credential_bindings`


tcdowney:
### Blockers/Dependencies

* Not a strict blocker, but this story may be easier once the `POST /v3/service_credential_bindings` story is completed. https://github.com/cloudfoundry/cf-k8s-controllers/issues/458

### Background

**As a** client of the API Shim
**I want** to be able to list service bindings
**So that** I can discover which service instances are bound to which apps

The CF CLI makes the following request as part of `cf services`:
```
GET /v3/service_credential_bindings?include=app&service_instance_guids=0c024675-7e0b-475c-8332-74986b5001a2%2C896fd3fa-28ec-42ff-b846-531e2ecabdaa
```

**Note:** Supporting `include=app` will be done as part of https://github.com/cloudfoundry/cf-k8s-controllers/issues/627

### Acceptance Criteria

The following set up applies to all of these scenarios.

1. Created a new Org and Space
2. `cf push app-a` (app guid is `114edd6d-1960-46df-a1dc-0309ce8ed63b`)
3. `cf push app-b` (app guid is `fa1069fb-922a-4a19-8a31-9129a82baff1`)
4. `cf cups service-a` (service instance guid is `5be373e9-309c-4ad9-a705-290525f36bd3`)
5. `cf cups service-b` (service instance guid is `fa0e32b3-e58b-4533-8d61-59ac38b1fa51`)
6. `cf bind-service app-a service-a` (binding guid is `08f1db1a-c7b9-4de2-9afa-5a7e53075d6d`)
7. `cf bind-service app-a service-b` (binding guid is `78a190a3-e3dd-4d42-8158-89aabfbd1673`)
8. `cf bind-service app-b service-b` (binding guid is `9c4cb7fc-eb61-49d8-9c16-a949eae55ed9`)

---

### No Filters

**GIVEN** I have created the apps, UPSIs, and bindings defined above
**WHEN I** make the following request

```
cf curl /v3/service_credential_bindings
```

**THEN I** get back a `200` response that looks similar to the following

```json
{
  "pagination": {
    "total_results": 3,
    "total_pages": 1,
    "first": {
      "href": "https://api.atom-kitten.capi.land/v3/service_credential_bindings?page=1&per_page=50"
    },
    "last": {
      "href": "https://api.atom-kitten.capi.land/v3/service_credential_bindings?page=1&per_page=50"
    },
    "next": null,
[...]

PR #682, opened 2022-02-17
Implement the list service bindings endpoint

acosta11:
#603

## What is this change about?
#603 

No

## Acceptance Steps
#603

paired w/ @julian-hj @gnovv


julian-hj (approved):
🦸

api/apis/integration/service_binding_test.go
```diff
@@ -0,0 +1,280 @@
...
+				},
+			}
+			Expect(
+				k8sClient.Create(context.Background(), cfApp),
+			).To(Succeed())
+
+			_ = createServiceInstance(context.Background(), k8sClient, serviceInstanceGUID, spaceGUID, "service-instance-name", secretName)
+		})
+
+		When("type is `app`", func() {
```
danail-branekov, line 82: What other types are there? If `app` is the only option, this `When` is misleading and not needed

api/apis/integration/service_binding_test.go
```diff
@@ -0,0 +1,280 @@
...
+			It("is successful and creates a CFServiceBinding resource", func() {
+				var err error
+				req, err = http.NewRequestWithContext(ctx, "POST", serverURI("/v3/service_credential_bindings"), strings.NewReader(fmt.Sprintf(`{
```
danail-branekov, line 85: Invoking the method under test should generally go into `JustBeforeEach` instead of being inlined in an `It`

api/apis/integration/service_binding_test.go
```diff
@@ -0,0 +1,280 @@
...
+					"labels":      map[string]interface{}{},
+				}))
+				Expect(response).To(HaveKeyWithValue("links", MatchAllKeys(Keys{
+					"app":              HaveKeyWithValue("href", HaveSuffix("/v3/apps/"+appGUID)),
+					"service_instance": HaveKeyWithValue("href", HaveSuffix("/v3/service_instances/"+serviceInstanceGUID)),
+					"self":             HaveKeyWithValue("href", HaveSuffix("/v3/service_credential_bindings/"+guid)),
+					"details":          HaveKeyWithValue("href", HaveSuffix("/v3/service_credential_bindings/"+guid+"/details")),
+				})))
+
+				serviceBinding := new(servicesv1alpha1.CFServiceBinding)
```
danail-branekov, line 148: The test so far has been checking the response. From here on the test checks the k8s object. Therefore, that should be a separate `It`. Also, if creating the object is performed in a `JustBeforeEach`, that It would be just a sibling to the response checking one.

api/apis/integration/service_binding_test.go
```diff
@@ -0,0 +1,280 @@
...
+		)
+
+		var (
+			serviceInstance1GUID, serviceBinding1GUID string
+			serviceInstance2GUID, serviceBinding2GUID string
+			spaceGUID                                 string
+		)
+
+		BeforeEach(func() {
+			org := createOrgAnchorAndNamespace(ctx, rootNamespace, generateGUID())
```
danail-branekov, line 179: Creating the org, the space and the app is shared for both `Describes`. In. order to remove the code duplication the common setup should be extracted into a top-level `BeforeEach`

api/apis/integration/service_binding_test.go
```diff
@@ -0,0 +1,280 @@
...
+			_ = createServiceBinding(ctx, k8sClient, serviceBinding1GUID, spaceGUID, &serviceBindingName, serviceInstance1.Name, cfApp.Name)
+
+			serviceInstance2GUID = generateGUID()
+			serviceInstance2 := createServiceInstance(ctx, k8sClient, serviceInstance2GUID, spaceGUID, "service-instance-2-name", secretName)
+
+			serviceBinding2GUID = generateGUID()
+			_ = createServiceBinding(ctx, k8sClient, serviceBinding2GUID, spaceGUID, nil, serviceInstance2.Name, cfApp.Name)
+
+			var err error
+			req, err = http.NewRequestWithContext(ctx, "GET", serverURI("/v3/service_credential_bindings"), strings.NewReader(""))
```
danail-branekov, line 215: The request is not changed in the underlying contexts, so maybe moving its creation to the `JustBeforeEach` would maybe make the code a bit more readable.

api/apis/integration/service_binding_test.go
```diff
@@ -0,0 +1,280 @@
...
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		JustBeforeEach(func() {
+			router.ServeHTTP(rr, req)
+		})
+
+		When("the user has permission to List ServiceBindings", func() {
```
danail-branekov, line 223: There is no test for the case when the user does not have permissions to list service bindings.

api/apis/service_binding_handler.go
```diff
@@ -93,6 +96,49 @@ func (h *ServiceBindingHandler) createHandler(authInfo authorization.Info, w htt
...
+	if err := r.ParseForm(); err != nil {
+		h.logger.Error(err, "Unable to parse request query parameters")
+		writeUnprocessableEntityError(w, "unable to parse query")
+		return
+	}
+
+	listFilter := new(payloads.ServiceBindingList)
+	err := schema.NewDecoder().Decode(listFilter, r.Form)
+	if err != nil {
+		switch err.(type) {
```
danail-branekov, line 112: How about extracting the check for UnknownKeyError into a utility function so that the code below reads like

```
if isUnknownKeyError(err) {
   // return unknown key error
}
// return unknown error

File: api/apis/service_binding_handler_test.go [hunk 8]
```diff
@@ -365,4 +366,124 @@ var _ = Describe("ServiceBindingHandler", func() {
 			})
 		})
 	})
+
+	Describe("the GET /v3/service_credential_bindings endpoint", func() {
+		makeListRequest := func(queryParams ...string) {
```
danail-branekov, line 371: How about renaming `makeListRequest` to `JustBeforeEach` so that the test is more idiomatic. then the function itself would not have to be explicitly invoked in all `BeforeEach`s

api/repositories/service_binding_repository.go
```diff
@@ -142,3 +152,57 @@ func cfServiceBindingToRecord(binding v1alpha1.CFServiceBinding) ServiceBindingR
 		},
 	}
 }
+
+func (r *ServiceBindingRepo) ListServiceBindings(ctx context.Context, authInfo authorization.Info, message ListServiceBindingsMessage) ([]ServiceBindingRecord, error) {
+	nsList, err := r.namespacePermissions.GetAuthorizedSpaceNamespaces(ctx, authInfo)
+	if err != nil {
+		// untested
```
danail-branekov, line 159: Is there any reason to have this branch untested? Why do we need such sort of comments?

api/repositories/service_binding_repository.go
```diff
@@ -142,3 +152,57 @@ func cfServiceBindingToRecord(binding v1alpha1.CFServiceBinding) ServiceBindingR
...
+			continue
+		}
+		if err != nil {
+			// untested
+			return []ServiceBindingRecord{}, fmt.Errorf("failed to list service instances in namespace %s: %w", ns, err)
+		}
+		filteredServiceBindings = append(filteredServiceBindings, applyServiceBindingListFilter(serviceInstanceList.Items, message)...)
+	}
+
+	return returnServiceBindingList(filteredServiceBindings), nil
```
danail-branekov, line 183: `return returnXXX` sounds weird, maybe change the name to describe what the function actually does, e.g. `toServiceBindingRecords`

api/repositories/service_binding_repository_test.go
```diff
@@ -237,7 +243,103 @@ var _ = Describe("ServiceBindingRepo", func() {
...
+			requestMessage = repositories.ListServiceBindingsMessage{}
+		})
+
+		When("the user has access to both namespaces", func() {
+			BeforeEach(func() {
+				createRoleBinding(testCtx, userName, spaceDeveloperClusterRole.Name, space.Name)
+				createRoleBinding(testCtx, userName, spaceDeveloperClusterRole.Name, space2.Name)
+			})
+
+			Describe("no query parameters are specified", func() {
```
danail-branekov, line 282: In order to be consistent with the context below, this `Describe` should be a `When`

api/repositories/service_binding_repository_test.go
```diff
@@ -237,7 +243,103 @@ var _ = Describe("ServiceBindingRepo", func() {
...
+				It("returns a list of ServiceBindingRecords in the spaces the user has access to", func() {
+					Eventually(func() []repositories.ServiceBindingRecord {
```
danail-branekov, line 284: Does this need to be `Eventually` instead of ordinary `Expect`? Resources are created synchronously in the `BeforeEach`, so should be immediately available in the `It`

api/repositories/service_binding_repository_test.go
```diff
@@ -237,7 +243,103 @@ var _ = Describe("ServiceBindingRepo", func() {
...
+			When("query parameters are specified", func() {
+				BeforeEach(func() {
+					requestMessage = repositories.ListServiceBindingsMessage{
+						ServiceInstanceGUIDs: []string{serviceInstance2GUID},
+					}
+				})
+
+				It("eventually returns only a subset of the ServiceBindings that match the provided filter", func() {
+					Eventually(func() []repositories.ServiceBindingRecord {
+						responseServiceBindings, err := repo.ListServiceBindings(context.Background(), authInfo, requestMessage)
```
danail-branekov, line 326: `requestMessage` is a local to this invocation, so could be inlined, instead of initialising it in a `BeforeEach` above. Alternatively, the `ListServiceBindings` function could be invoked in `JustBeforeEach` so that it is not invoked in `Its`. Then, setting up the request message in `BeforeEach` above would make sense.

api/repositories/service_binding_repository_test.go
```diff
@@ -237,7 +243,103 @@ var _ = Describe("ServiceBindingRepo", func() {
...
+					}, timeCheckThreshold*time.Second).Should(ConsistOf(
+						MatchFields(IgnoreExtras, Fields{"GUID": Equal(serviceBinding2.Name)}),
+					))
+				})
+			})
+		})
+
+		Describe("the user does not have access to any namespaces", func() {
+			It("always returns an empty list and no error", func() {
+				Consistently(func() []repositories.ServiceBindingRecord {
```
danail-branekov, line 338: Same as the above comment regarding `Expect` vs `Eventually`

api/repositories/service_instance_repository_test.go
```diff
@@ -504,6 +504,7 @@ func initializeServiceInstanceCreateMessage(serviceInstanceName string, spaceGUI
 	}
 }
 
+// TODO: refactor to use createServiceInstanceCR
```
danail-branekov, line 507: Is there a a chore about this TODO? If yes, having it here does not make sense, if no, please create one.

danail-branekov (changes requested):
Please look at the comments above.

api/apis/service_binding_handler.go
acosta11: Added https://github.com/cloudfoundry/cf-k8s-controllers/issues/698

api/repositories/service_binding_repository_test.go
acosta11: I don't know that the resources are actually synchronously available, even after the create call succeeds,  though perhaps the cache gets updated if the element is absent? I feel like we've seen flakes related to directly calling `Expect`.

api/repositories/service_instance_repository_test.go
acosta11: Good call, we just did the refactor in place since the changes were only in one file.

acosta11:
Put the PR back in draft to prevent merges, but addressed all of the comments. Need to squash and do one last look-over, but we have now hit EOD and some.

api/apis/integration/service_binding_test.go
```diff
@@ -0,0 +1,295 @@
...
+			serviceInstanceGUID string
+			err                 error
+		)
+		BeforeEach(func() {
+			createRoleBinding(ctx, userName, spaceDeveloperRole.Name, spaceGUID)
+
+			serviceInstanceGUID = generateGUID()
+			_ = createServiceInstance(context.Background(), k8sClient, serviceInstanceGUID, spaceGUID, "service-instance-name", secretName)
+			req, err = http.NewRequestWithContext(ctx, "POST", serverURI("/v3/service_credential_bindings"), strings.NewReader(fmt.Sprintf(`{
+																																		 "relationships": {
```
georgethebeatle, line 93: This formatting is a bit weird. On our machine it is offset in the far right part of the screen and hard to read.

danail-branekov (approved):

georgethebeatle:
Overall looks good. We filed a draft PR #708 that suggests some further improvements. Please have a look.

PR #689, opened 2022-02-18
Refactor namespace lookup helper

Birdrock:
No.

## What is this change about?
Proposing a refactor to extract the list -> namespace functionality present in all of the authenticated get endpoints. We should agree on an approach to handle this generic situation.

Other idea: code generation in the same vein of kubebuilder/controller-gen annotations.

No.

## Acceptance Steps
This is a draft.

@cloudfoundry/cf-k8s


api/repositories/shared.go
```diff
@@ -59,3 +67,73 @@ func getLabelOrAnnotation(mapObj map[string]string, key string) string {
...
+			return "", err
+		}
+		return t.Items[0].Namespace, nil
+	case *servicesv1alpha1.CFServiceInstanceList:
+		err := CheckListLength(len(t.Items), "service-instance")
+		if err != nil {
+			return "", err
+		}
+		return t.Items[0].Namespace, nil
+	default:
```
Birdrock, line 126: This is obviously would not be a proper implementation - need to handle it in a better way.

api/repositories/shared.go
```diff
@@ -59,3 +67,73 @@ func getLabelOrAnnotation(mapObj map[string]string, key string) string {
 	}
 	return mapObj[key]
 }
+
+func GetNamespacedResourceNamespace(ctx context.Context, privilegedClient client.Client, userClientFactory UserK8sClientFactory, authInfo authorization.Info, resourceGUID string, resourceList client.ObjectList) (string, error) {
```
danail-branekov, line 71: There are several issues we find with extracting the code into a shared utility:
1. The privileged client is still part of the function signature, i.e. should be still available in the caller. This is one of the things we want to get rid of, i.e. handlers should only use unprivileged clients. We would recommend putting the code into a type that has the privileged client as a private property and inject that type into handlers.
2. The `userClientFactory` is not used, so it makes no sense to be in the function arguments. Moreover, this utility should be using the privileged client only.

api/repositories/shared.go
```diff
@@ -59,3 +67,73 @@ func getLabelOrAnnotation(mapObj map[string]string, key string) string {
...
+	err := privilegedClient.List(ctx, resourceList, client.MatchingFields{"metadata.name": resourceGUID})
+	if err != nil { // untested
+		return "", fmt.Errorf("privileged list failed: %w", err)
+	}
+
+	switch t := resourceList.(type) {
```
danail-branekov, line 77: This huge switch does not look great - it has quite a lot of duplicated code only because there is no `Items` method in the `ObjectList` interface. Maybe we could use reflection to see whether the incoming `resourceList` has an `Items` field and return error otherwise. Or, we could use some code generator as you suggested, but that might be way too complex.

api/repositories/shared.go
```diff
@@ -59,3 +67,73 @@ func getLabelOrAnnotation(mapObj map[string]string, key string) string {
...
+	case *workloadsv1alpha1.CFAppList:
+		err := CheckListLength(len(t.Items), "app")
```
danail-branekov, line 79: The resource type could probably be passed as an argument, rather than the utility figuring it out by itself (repositories are pretty much aware of the resource types they are working with)

api/repositories/shared.go
Birdrock: The reason for the switch is to avoid reflection. I treat reflection as a last resort - hence this approach to some unfortunate but maintainable code that should be relatively set and forget.

api/repositories/shared.go
Birdrock: 1. Agreed and we talked about this, but wanted to put the general idea out there for feedback.
2. Agreed - this is a result of moving things around while exploring multiple solutions.

api/repositories/shared.go
Birdrock: Further digging turns up the [kubernetes apimachinery converter](https://github.com/kubernetes/apimachinery/blob/master/pkg/runtime/converter.go). This looks like it'll help us out in this situation to get to unstructured objects. It uses reflection under the hood, so maybe this is the best way to do it. Since kubernetes provides this for us, I'd recommend we look into using the provided tooling instead of the reflect library directly.

Birdrock: Further spelunking suggests that https://pkg.go.dev/k8s.io/client-go/dynamic might help us.

gcapizzi: Yes, `dynamic` seems to do the trick! We'd need a privileged [`dynamic.Interface`](https://pkg.go.dev/k8s.io/client-go/dynamic#Interface) client (not sure if in substitution of `privilegedClient` or in addition to it), and then we could get an [`UnstructuredList`](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured#UnstructuredList) which _should_ be enough to check the length and to extract the namespace. [`Unstructured`](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured#Unstructured) objects are essentially `map[string]interface{}`s but it should work!

Birdrock:
We made a second pass at it after looking at the dynamic client. Open for comments!

api/repositories/shared.go
```diff
@@ -1,15 +1,56 @@
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+var CFAppGVR = schema.GroupVersionResource{
+	Group:    "workloads.cloudfoundry.org",
+	Version:  "v1alpha1",
+	Resource: "cfapps",
+}
+
+type NamespaceRetriever struct {
```
georgethebeatle, line 23: I would put this object along with its method in a separate file. Putting all sorts of utilites in a shared file would result in the file becoming huge and hard to orient in.

api/repositories/shared.go
```diff
@@ -1,15 +1,56 @@
...
+	client dynamic.Interface
+}
+
+func NewNamespaceRetriver(client dynamic.Interface) NamespaceRetriever {
+	return NamespaceRetriever{
+		client: client,
+	}
+}
+
+func (n NamespaceRetriever) GetNamespacedResourceNamespace(ctx context.Context, resourceGUID string, gvr schema.GroupVersionResource) (string, error) {
```
georgethebeatle, line 33: Maybe the name can be GetResourceNamespace? Sounds a bit redundant to me in its current form.

api/repositories/shared.go
```diff
@@ -1,15 +1,56 @@
...
+	if len(list.Items) == 0 {
+		return "", NewNotFoundError(gvr.Resource, nil)
+	}
+	if len(list.Items) > 1 {
+		return "", fmt.Errorf("get-%s duplicate records exist", gvr.Resource)
+	}
+
+	metadata := list.Items[0].Object["metadata"].(map[string]interface{})
+
+	return metadata["namespace"].(string), nil
```
georgethebeatle, line 51: Maybe we should check if namespace == "", which would mean the object is not namespace scoped, in which case we can return an error? I guess this is only possible because of programmer error, but still it will be good to catch it early.

georgethebeatle (commented):
Looks great! I left some minor comments inline.

Birdrock:
Please make adjustments as you see fit.

api/repositories/shared.go
Birdrock: I like this sort of error checking. Thanks for the input.

api/repositories/shared.go
Birdrock: Naming is hard. Thanks for the guidance.

Birdrock:
@georgethebeatle If you'd like to make the changes and squash the commits, do so - this is a team effort to refactor.

api/repositories/shared.go
danail-branekov: +1 for grouping similar objects into separate file. `shared.go` sound like `oh, I don't know where to put that, let me shove it in shared`

api/repositories/shared.go
danail-branekov: +1 for better name. How about `NamespaceFor(...)`?

danail-branekov:
I like the latest version of the utility, left few minor comments

kieron-dev:
I also think this is looking nice now.

Are we missing a contract test for the NamespaceRetriever though - e.g. a test for just the NamespaceRetriever on top of env test?

davewalter:
> Are we missing a contract test for the NamespaceRetriever though - e.g. a test for just the NamespaceRetriever on top of env test?

I was thinking the same thing. I'll work on adding one today.

georgethebeatle:
@Birdrock I feel like we have reached consensus on the shape of this and it's time to actually apply it to the codebase. Maybe it can be done in one go across all endpoints. After all we are looking up namespaces by guid only on Get endpoints and there are not that many of them. Feel free to execute it however you like. I guess since US started this conversation credit should go to you all.

Birdrock:
Sounds good. I'll implement across all the GET endpoints. The last bit is about the error messaging, since we're grabbing it from the GVR. I was thinking of making a map to try to preserve existing messaging, but I'll see how it works out.

davewalter:
We've applied this to all existing "get" endpoints. We also implemented auth for the `GET /v3/routes/:guid` endpoint in the process (#640), but there is a problem with the new E2E that we added as part of this work related to the RBAC permissions required to create domains in the root `cf` namespace. Hopefully we can get that resolved on Monday and get this merged.

gnovv (approved):
Pairing w/ Dave 2/28/22 LGTM!

PR #695, opened 2022-02-22
DRAFT: Experiment with error handling

georgethebeatle:
We have modified GetApp and SetCurrentDroplet in the app handler as an example

The main changes are as follows:

Errors:
- We have a new single location for errors: the `apierr` package
- Errors within `apierr` are objects that replace the `writeXXXError` methods from `apis/shared.go`.
- These errors are raised throughout the system - mainly in repositories and handlers
- These errors have all the information, i.e. http status and error details, to write out API error messages

Repositories:
- Repo code calling k8s methods should map k8s errors to `apierr` errors using `wrapK8sError` which also takes the resourceType

Handlers:
- Handler methods now have a new signature: `handleXXX(authInfo authorization.Info, r *http.Request) (status int, responseObj interface{}, err error)`
    - They do not get the response writer any more as this is now responsibility of the endpoint function wrapper (see below)
    - Return status and responseObj for successful requests
    - Return an error for error responses
        - An `apierr` error will render a custom API error
        - Any other error will be an 'UnknownError' 500
- Handlers should modify ForbiddenError to NotFoundErrors where appropriate using the `apierr.AsNotFoundIfForbidden(err)` helper
- Handlers should modify NotFoundError or ForbiddenError to UnprocessableEntityErrors where appropriate using the `apierr.AsUnprocessableIfNotFoundOrForbidden(err)` helper
    - We assume mapping NotFound and Forbidden to Unprocessable is the only general case. If not, we might think about callbacks we could pass in to match certain errors and be more flexible
- A new wrapper in the routes wiring handles presenting the successful or error responses

### DIFF
--- a/api/apis/app_handler.go
+++ b/api/apis/app_handler.go
@@ -2,6 +2,7 @@ package apis
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -239,17 +240,15 @@ func (h *AppHandler) appSetCurrentDropletHandler(authInfo authorization.Info, w 
 	dropletGUID := payload.Data.GUID
 	droplet, err := h.dropletRepo.GetDroplet__NewStyle(ctx, authInfo, dropletGUID)
 	if err != nil {
-		switch err.(type) {
-		case apierr.NotFoundError:
-			h.logger.Error(err, "Droplet not found", "dropletGUID", dropletGUID)
+		h.logger.Info("GetDroplet failed", "dropletGUID", dropletGUID, "error", err)
+
+		var apiErr apierr.ApiError
+		if errors.As(err, &apiErr) {
 			writeErrorResponse(w, presenter.ForError(apierr.NewUnprocessableEntityError(err, invalidDropletMsg)))
-		case apierr.ForbiddenError:
-			h.logger.Error(err, "Droplet not authorized for user", "dropletGUID", dropletGUID)
-			writeErrorResponse(w, presenter.ForError(apierr.NewUnprocessableEntityError(err, invalidDropletMsg)))
-		default:
-			h.logger.Error(err, "Error fetching droplet")
-			writeErrorResponse(w, presenter.ForError(err))
+			return
 		}
+
+		writeErrorResponse(w, presenter.ForError(err))
 		return
 	}
 
--- a/api/apis/app_handler.go
+++ b/api/apis/app_handler.go
@@ -2,7 +2,6 @@ package apis
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
@@ -13,6 +12,7 @@ import (
 
 	"code.cloudfoundry.org/cf-k8s-controllers/controllers/webhooks/workloads"
 
+	"code.cloudfoundry.org/cf-k8s-controllers/api/apierr"
 	"code.cloudfoundry.org/cf-k8s-controllers/api/authorization"
 	"code.cloudfoundry.org/cf-k8s-controllers/api/payloads"
 	"code.cloudfoundry.org/cf-k8s-controllers/api/presenter"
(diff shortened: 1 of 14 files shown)


api/apis/app_handler.go
```diff
@@ -219,37 +224,37 @@ func (h *AppHandler) appSetCurrentDropletHandler(authInfo authorization.Info, w
...
-		h.handleGetAppErr(err, w, appGUID)
+		writeErrorResponse(w, presenter.ForReadError(err))
 		return
 	}
 
 	dropletGUID := payload.Data.GUID
-	droplet, err := h.dropletRepo.GetDroplet(ctx, authInfo, dropletGUID)
+	droplet, err := h.dropletRepo.GetDroplet__NewStyle(ctx, authInfo, dropletGUID)
 	if err != nil {
 		switch err.(type) {
```
danail-branekov, line 242: I hope that this is quite a corner case where would have to use those terrible switches in rest handlers, right? I really hoped that I would not see such a switch again :(

julian-hj: +1

Maybe it would make the helper functions ugly, but it does feel like this is a common case that could be shifted to a helper. 

If I'm reading it right, then we got an error while looking something up in order to create or modify something else. In this case, if we get a forbidden or not found, we want to translate that into an "unprocessable entity" error. We will probably need this logic in almost every handler.

julian-hj: Yes, that looks much better.

api/apis/auth_aware_handler.go
```diff
@@ -33,3 +35,34 @@ func (wrapper *AuthAwareHandlerFuncWrapper) Wrap(delegate AuthAwareHandlerFunc)
...
+	logger logr.Logger
+}
+
+func NewAuthAwareHandlerFuncWrapper__NewStyle(logger logr.Logger) *AuthAwareHandlerFuncWrapper__NewStyle {
+	return &AuthAwareHandlerFuncWrapper__NewStyle{logger: logger}
+}
+
+func (wrapper *AuthAwareHandlerFuncWrapper__NewStyle) Wrap(delegate AuthAwareHandlerFunc__NewStyle) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		w.Header().Add(headers.ContentType, "application/json")
```
gnovv, line 51: What should we do when the header is not JSON?
We have one or two edge cases, like the jobs endpoints for example

matt-royal: One option would be to add a method to all of the JSON-encoded presenter structs, like `AsJSONEncodable()`. We could then use a type assertion in the wrapper to determine if it should be encoded as JSON, YAML (based on presence of `AsYAMLEncodable()` method), or as plain text (if it's Readable, a string or a byte slice).

Another, potentially simpler option would be to use a different wrapper method for different response types.

matt-royal (approved):
This is a great change. Akira's comment about encoding non-JSON responses needs to be addressed, but otherwise I think we should do this.

I don't love that the handler returns two unused values in the error case (e.g `return 0, nil, err`). One possible alternative would be to return a `SuccessResponse` struct that includes status and payload. This could also include the response encoding (e.g `json`, `text`), which would address Akira's concern. That may end up being more than we need in practice, though, so take that suggestion with a grain of salt.

api/apis/auth_aware_handler.go
danail-branekov: Jobs endpoints [do return json](https://github.com/cloudfoundry/cf-k8s-controllers/blob/489b41eecfcbc08151254da2fe9c6513300ed1fa/api/apis/job_handler.go#L62), the `writeResponse` [function](https://github.com/cloudfoundry/cf-k8s-controllers/blob/489b41eecfcbc08151254da2fe9c6513300ed1fa/api/apis/shared.go#L349-L360) always encodes the result to json.

danail-branekov:
> One possible alternative would be to return a SuccessResponse struct that includes status and payload. This could also include the response encoding (e.g json, text), which would address Akira's concern.

I agree that maybe using a response object instead of two "sucess" values is maybe a good thing, it also enable future extensibility. However, if there are no examples of endpoints returning non-json object, I would recommend doing the simplest thing, i.e. always encode to json. If non-json cases appear, we could easily extend the response object to also contain encoding instructions.

api/apis/auth_aware_handler.go
gnovv: My bad, I could have sworn that I worked on an endpoint that did not always return JSON, but I may be thinking of an older version of the code where we did something incorrectly.

davewalter: Some of the asynchronous delete handlers (for example, [app delete](https://github.com/cloudfoundry/cf-k8s-controllers/blob/46c9668d4c6a74f99337888d7a2057bdaf874a03/api/apis/app_handler.go#L526-L550)) don't return anything in the body. Would it matter if these responses included a `Content-Type` header?

georgethebeatle:
Implemented by #798

PR #700, opened 2022-02-23
Update process usage stats

### LINKED ISSUE #681: [Feature]: API can get process stats
### Blockers/Dependencies

Depends on [the spike branch](https://github.com/cloudfoundry/cf-k8s-controllers/tree/issues/459-wip-metrics-server-client).

### Background

**As a** API client
**I want** to view stats for given cf processes
**So that** I can observe my applications

### Happy Path
**GIVEN** cluster with cf-k8s-controllers, api shim, and metrics server installed
**WHEN I** make a request that looks like
```
curl "https://api.example.org/v3/processes/[guid]/stats" \
-X GET \
```
**THEN I** get a response that looks like 
```
HTTP/1.1 200 OK

{
  "resources": [
    {
      "type": "web",
      "index": 0,
      "state": "RUNNING",
      "usage": { "time": "2016-03-23T23:17:30.476314154Z", "cpu": 0.00038711029163348665, "mem": 19177472, "disk": 69705728 },
      "host": "10.244.16.10",
      "instance_ports": [],
      "uptime": 9042,
      "mem_quota": 268435456,
      "disk_quota": 1073741824,
      "fds_quota": 0,
      "isolation_segment": null,
      "details": null
    }
  ]
}
```

### Dev Notes

API docs: https://v3-apidocs.cloudfoundry.org/version/3.113.0/index.html#get-stats-for-a-process 

Because our installation scripts do not currently include metrics-server, we will need to update them to enable end to end testing.

Notes on particular fields:
* Instance ports meaning in this case is unclear, so ignore the block in the response for this story.
* Similarly, assume the file descriptor quota is 0 as we do not get insight into that information. (And may not have that concern on K8s)
* If uptime is unclear from k8s metrics response, make a note and we can also circle back.

Note on RBAC:
* The metrics api group is separate from corev1, so the existing process/pod RBAC is likely insufficient, but we should be able to determine the necessary RBAC on the metrics pod resource

Fallback:
* If metrics server is not available, continue to return the empty usage block
Do we also include a runtime log for debugging as well?

clintyoshimura:
[TME-1595](https://jira.eng.vmware.com/browse/TME-1595)

Co-authored-by: Clint Yoshimura <clinty@vmware.com>
Co-authored-by: Akira Wong <wakira@vmware.com>

Resolves #681

## What is this change about?

Updates process stats endpoint to add usage from metrics server

no

## Acceptance Steps

See issue

@akrishna90 @gnovv


gnovv (approved):
Pairing with @clintyoshimura LGTM!

Issue #705, opened 2022-02-23
[Chore]: Look for opportunities to speed up API e2e testing


julian-hj:
### Background

We noticed that the current e2e api tests take quite a lot of time for setup for each test--something like 6 minutes a test.

This will quickly become very painful as we add more tests to the suite.

We should look for ways to speed things up. 

### Action to take

new

### Impact

medium

### Dev Notes

Some starter ideas are:
- Side-load the buildpack image into kind (like we do with the api & controller images). Cache it locally so that we don't have to re-download it on each test run.
- Share the org between tests and only build and tear down the space
- Use a binary app for cf push. Or a golang app. Or whatever is fastest.
- CF push a single "read-only" app in the `beforeSuite` that can be used for testing GET endpoints so that we don't have to do as much test setup.

georgethebeatle:
Maybe another opportunity to further speed them up will be to deploy a long running gke cluster and add a new job that deploys stuff to that cluster when a PR is submitted. A complication here will be that PR workflows in github are unprivileged and therefore have no access to secrets, so it might be interesting how we pass the kubeconfig on.

georgethebeatle:
Another one that we already started workig on is #572 which actually aims at speeding up the product itself.

acosta11:
Today we tried pushing a cross compiled go executable as a binary app instead of the existing node app and saw a change from around 30 seconds to 25 seconds on push time. So we may pursue that if we find we're pushing apps frequently. But we only found 7 tests that use the existing `pushNodeApp` helper in their setup, so the potential time savings may not be as significant as we would like.

Similarly, we're looking at pre-pushing a single app in the  `BeforeSuite` (instead of the `BeforeAll` proposed on #717) to act as a read only resource to test api get endpoints. We're not sure whether user permissions are concern for this idea given the potential for test pollution if we just used a normal space developer role and the app was modified in some way. And similarly to the node vs binary app, there appears to only be 7 uses of the node app helper function.

This leaves trying to re-use orgs in between tests from the list of ideas.

julian-hj:
Thanks @georgethebeatle. Should we close this issue for now? I am not sure that it is still useful to have open.

georgethebeatle:
We are not actively looking into it and it sounds like you aren't either, so closing for the time being.

kieron-dev:
Adding some more thoughts that came up today:

[...]

PR #717, opened 2022-02-25
[DRAFT] Use Ginkgo Ordered tests to avoid repeated test setups

julian-hj:
- This is a WIP PR to test out the approach & allow for discussion. It isn't intended to be merged as-is.
- If we were to adopt the approach, we would likely want to move the
  spaceDev and spaceMgr client setup into the BeforeSuite and move the
  role setup into the createOrg and createSpace setup calls so that we
  don't have to do it in individual tests.
- Applying these changes reduced the overall run time for the 5
  affected tests from 3m34s to 1m12s, since we only create one org, one space, one app.
- In other test scenarios where we are creating or modifying objects, we would need to be more careful about teardown, since we're using the same space across tests.

@georgethebeatle @matt-royal @kieron-dev @gcapizzi 

[#705]

### DIFF
--- a/api/apis/service_binding_handler.go
+++ b/api/apis/service_binding_handler.go
@@ -110,7 +110,7 @@ func (h *ServiceBindingHandler) listHandler(authInfo authorization.Info, w http.
 	err := schema.NewDecoder().Decode(listFilter, r.Form)
 	if err != nil {
 		if isUnknownKeyError(err) {
-			h.logger.Info("Unknown key used in ServiceInstance filter")
+			h.logger.Info("Unknown key used in ServiceInstance query parameters")
 			writeUnknownKeyError(w, listFilter.SupportedFilterKeys())
 		} else {
 			h.logger.Error(err, "Unable to decode request query parameters")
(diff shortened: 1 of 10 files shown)


georgethebeatle:
Overall looks good, but we are a bit concerned about the `Ordered` constraint that comes with `BeforeAll`. If we assume this strategy then the `--randomize-all` flag that we pass to ginkgo will be meaningless. By not running tests in random order we might introduce undesired order dependencies due to shared state. 

Currently we are trying to write tests in relative isolation (each `It` gets a new namespace). After this change all `It`s in the `Describe` would run in the same namespace which might lead to pollution/flakes. On the upside though they will run faster.

Maybe we can experiment with this and see if it causes problems and adopt it more widely if it doesn't... Wdyt?

julian-hj:
I'm going to close this PR, since I think we don't want to take this approach unless we absolutely have to, as it will make our tests brittle.

Speeding up org and space creation in the product itself seems like a much better way to may the e2e tests less painful.

Issue #161, opened 2021-10-14
[Feature]: Platform Operator can follow manual instructions to configure a default application domain


davewalter:
### Background

**As a** Platform Operator
**I want** manual instructions to tell me how to configure a default application domain
**So that** application developers can push an application without specifying domain information and have the platform use the appropriate default

### Acceptance Criteria

**GIVEN I** am targeting a K8s cluster
**WHEN I** follow manual instructions for configuring the default application domain
**AND I** follow the instructions for deploying the API and controllers
**AND I** push an application without specifying domain information
**THEN I** can see that the route generated for the application used the configured default domain

**GIVEN I** am targeting a K8s cluster
**WHEN I** do not specify a default application domain
**AND I** follow the instructions for deploying the API and controllers
**AND I** push an application without specifying domain information
**THEN I** can see a 404 response specifying the default domain was not found

### Dev Notes

* We suspect that in the manifest apply flow, we can control the selected domain internally instead of concerning the logic with the order of list domains output to the CLI.
* For this issue, we are not going to massage the list of domains returned by the API to ensure that the default is always first so that it is used by the CLI. We will circle back.
* Likely entails a new named field on config
* ~Consider also adding a warning log on startup for the cf api server specifying the default domain was not found?~
* Current behavior on apply manifest - https://github.com/cloudfoundry/cloud_controller_ng/blob/8590c1a8b824e794aa01bb0cdc9394ec1c9efd92/app/actions/app_apply_manifest.rb#L122

Issue #699, opened 2022-02-22
[Feature]: Improve how we surface errors in the Space Manifest


tcdowney:
### Background

**UPDATE:** The repro steps below are no longer accurate since we fixed that particular issue in https://github.com/cloudfoundry/cf-k8s-controllers/pull/725. We can still trigger this through other cases, though, like requesting a route for an app that already exists (see https://github.com/cloudfoundry/cf-k8s-controllers/issues/548#issuecomment-1058427556). 

Currently any error in the [`applyManifestAction`](https://github.com/cloudfoundry/cf-k8s-controllers/blob/main/api/actions/apply_manifest.go) will cause the `SpaceManifestHandler` to return a `500` error:
https://github.com/cloudfoundry/cf-k8s-controllers/blob/39dae24761131efa657437c09d4bc1db153d9fc3/api/apis/space_manifest_handler.go#L73

For example, I attempted to push an app and requested `default-route: true`, but there was no default domain configured. This resulted in an error [here](https://github.com/cloudfoundry/cf-k8s-controllers/blob/39dae24761131efa657437c09d4bc1db153d9fc3/api/actions/apply_manifest.go#L149), but instead of an error message like "No default domain configured", I got a `500` error.

```console
cf push node
Pushing app node to org o / space s as tdowney...
Unexpected Response
Response Code: 500
Code: 0, Title: , Detail: {"errors":[{"detail":"An unknown error occurred.","title":"UnknownError","code":10001}]}
```

**As an** App Developer
**I want** to receive meaningful errors when manifest application fails
**So that** I can fix my app manifest without requiring the platform operator to look at component logs

---

### Acceptance Criteria

One of the easier cases to reproduce is the "no default domain" case, but in general we should think through some of the errors we can handle better here without explicit A/C.

### No Default Domain Found
**GIVEN** I configured the CF API shim with a default domain, but haven't applied the `CFDomain` CR 
**WHEN I** `cf push` an app with `default-route: true` in its manifest
**THEN I** see a friendlier `422` error saying something like "No default domain available"

For other types of errors even just something like "Error creating `CFProcess`  or "Error updating `CFRoute`" could be better.

danail-branekov:
It seems that the issue has been fixed in the meanwhile:
* When I push a manifest requesting default route with no CFRoutes configured, but with a default domain configured in the helm chart, I get the following error: 

```
The configured default domain "apps-127-0-0-1.nip.io" was not found
```

[...]

PR #725, opened 2022-02-28
Add DomainNotFound error handling and missing documentation

akrishna90:
#161 

## What is this change about?
Add missing edge case handling and documentation

- throws a HTTP 404 instead of a HTTP 500 when Domain is not found
- add documentation for setting default domains

No

## Acceptance Steps
See #161

### DIFF
--- a/api/apis/space_manifest_handler_test.go
+++ b/api/apis/space_manifest_handler_test.go
@@ -310,6 +310,16 @@ var _ = Describe("SpaceManifestHandler", func() {
 				Expect(payload.Applications[0].DefaultRoute).To(BeTrue())
 			})
 		})
+
+		When("applying the manifest errors with NotFoundErr", func() {
+			BeforeEach(func() {
+				applyManifestAction.Returns(repositories.NotFoundError{})
+			})
+
+			It("respond with not found error", func() {
+				expectNotFoundError("Domain not found")
+			})
+		})
 	})
 
 	Describe("POST /v3/spaces/{spaceGUID}/manifest_diff", func() {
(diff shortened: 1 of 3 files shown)


api/apis/space_manifest_handler_test.go
```diff
@@ -310,6 +310,16 @@ var _ = Describe("SpaceManifestHandler", func() {
 				Expect(payload.Applications[0].DefaultRoute).To(BeTrue())
 			})
 		})
+
+		When("applying the manifest errors with NotFoundErr", func() {
+			BeforeEach(func() {
+				applyManifestAction.Returns(repositories.NotFoundError{})
+			})
+
+			It("respond with Unknown Error", func() {
```
tcdowney, line 319: I think this should be "Not Found Error" instead.

tcdowney:
I know the original story called for a `404`, but I think in this case it might make sense to do handle this differently than other "not found" cases and return a `422` saying that there is no default domain configured or something.

Because most cases when a user is getting a `404` they're trying to view or do something with a specific resource, but in this case all they're doing is including the `default-domain: true` key in their app manifest and I think a `404` might be confusing. The `POST /v3/spaces/:guid/actions/apply_manifest` endpoint given what it's doing.

tcdowney:
In some ways this flow is easier to reason about in CF for VMs cause it does most of the work as part of an async job and its just able to set the job status to `FAILED` and build up a list of errors [on the job](https://v3-apidocs.cloudfoundry.org/version/3.115.0/index.html#get-a-job) to give the user.

akrishna90:
Thanks @tcdowney for the feedback. I will make the change to return a `422` 
cc @Birdrock (author of the story)

tcdowney (approved):
LGTM! Seems fine to special case it right now since we've got https://github.com/cloudfoundry/cf-k8s-controllers/issues/699 to improve errors here more broadly.

Issue #726, opened 2022-02-28
[Bug]:  E2E Service Binding List Test Flake


gnovv:
### What happened?

Observed flake running the e2e tests- There appears to be test pollution related to RBAC between the the the whens. We believe it may be due to the username being static between test runs and the tests running in parallel.
```
------------------------------
• [SLOW TEST] [100.732 seconds]
Processes getting process stats we wait for the metrics to be ready succeeds
/home/runner/work/cf-k8s-controllers/cf-k8s-controllers/cf-k8s-controllers/api/tests/e2e/processes_test.go:119

  Begin Captured StdOut/StdErr Output >>

    WAIT RESPONSE

[...]

PR #728, opened 2022-03-01
Update e2e tests to create new user per test

davewalter:
#726 

## What is this change about?
This PR updates the E2E test setup to create a new user for each test, rather than reusing the same user for all tests. We hope that this will address a number of flakes we have seen in CI that appear to be due to the user having permission to perform actions that the test thinks they should not be able to do.

No

## Acceptance Steps
The e2e tests pass more consistently.

@cloudfoundry/cf-k8s

### DIFF
--- a/api/tests/e2e/e2e_suite_test.go
+++ b/api/tests/e2e/e2e_suite_test.go
@@ -165,13 +165,13 @@ var _ = BeforeSuite(func() {
 
 	serviceAccountName = generateGUID("token-user")
 	serviceAccountToken = obtainServiceAccountToken(serviceAccountName)
-
-	certUserName = generateGUID("cert-user")
-	certSigningReq, certPEM = obtainClientCert(certUserName)
-	certAuthHeader = "ClientCert " + certPEM
 })
 
 var _ = BeforeEach(func() {
+	certUserName = generateGUID("cert-user")
+	certSigningReq, certPEM = obtainClientCert(certUserName)
+	certAuthHeader = "ClientCert " + certPEM
+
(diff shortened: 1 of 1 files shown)


kieron-dev:
I suspect the recently added workaround to allow access to cfdomains in the root namespace has caused the problem that this PR is addressing. In particular, this rolebinding: https://github.com/cloudfoundry/cf-k8s-controllers/blob/2e1e4ac3ab01b5f73562a0f8cdf8d94ae5398114/api/tests/e2e/route_test.go#L63. It either needs an annotation so that HNC does not propagate it to all sub-namespaces (`hncv1alpha2.AnnotationNoneSelector: "true"`), or it should be deleted after each test.

As it is at the moment, as soon as the 'Fetch an route' test from `e2e/route_test.go` is run, the certUser on that ginkgo node will have bindings on all orgs and spaces on the e2e cluster. Since code to list resources relies on the presence of a binding in the org or space, I expect some tests will start to fail that check not being able to list resources in unauthorised namespaces.

So, even though using new users for each test would fix this issue, I think it is better to leave that behaviour as is since it exposes the problem with the binding in the root namespace. When the production code is ready to create that binding, it would be useful for the e2e tests to fail if the HNC propagation was not disabled, as it needs to be.

I am about to create a PR for the route uniqueness validation which also backfills some route e2e tests. This duplicates the problematic route e2e test linked above, so I will delete that file in my PR, and I expect that will fix the flakiness.

davewalter:
Closing now that #730 is merged.

Issue #178, opened 2021-09-23
[Feature]: API Client cannot create a duplicate route


davewalter:
### Background

As a client of the API shim
I want to receive an error when I attempt to create a duplicate route based on its FQDN + context path
So that I am prevented from creating a route if it already exists

#### With context path
**GIVEN** I have an existing route (e.g. `foo.example.com/bar` )
**WHEN** I try to make another route with that same host, domain, and context path
**THEN** receive a `422` error

```json
{
  "errors": [
    {
      "code": 10008,
      "detail": "Route already exists with host 'foo' and path '/bar' for domain 'atom-kitten.capi.land'.",
      "title": "CF-UnprocessableEntity"
    }
  ]
}
```

#### Without context path
**GIVEN** I have an existing route (e.g. `foo.example.com` )
**WHEN** I try to make another route with that same host and domain
**THEN** receive a `422` error

```json
{
  "errors": [
    {
      "code": 10008,
      "detail": "Route already exists with host 'foo' for domain 'atom-kitten.capi.land'.",
      "title": "CF-UnprocessableEntity"
    }
  ]
}
```

### Dev Notes

* Start by implementing this the same way as we do for app names (with leases)
  * We recognize that the direct kubectl apply use case may not have sufficient guards against tampering, but the main API use case remains covered

tcdowney:
Output from a CF for VMs env when creating a duplicate route:
```json
REQUEST: [2022-02-14T10:33:04-07:00]
POST /v3/routes HTTP/1.1
Host: api.atom-kitten.capi.land
Accept: application/json
Authorization: [PRIVATE DATA HIDDEN]
Content-Type: application/json
User-Agent: cf/8.1.0+034d929d7.2022-01-03 (go1.16.6; amd64 darwin)
{
  "host": "foo",
  "relationships": {
    "domain": {
      "data": {
        "guid": "44085066-fdb5-452f-92aa-6d7b2a7c0766"
      }
    },
    "space": {
      "data": {
        "guid": "ed61c32e-1aa8-4817-a5dd-6059ae40a32b"
      }
    }
  }
}


RESPONSE: [2022-02-14T10:33:04-07:00]
HTTP/1.1 422 Unprocessable Entity
Content-Type: application/json; charset=utf-8
Date: Mon, 14 Feb 2022 17:33:04 GMT
Referrer-Policy: strict-origin-when-cross-origin
Server: nginx
[...]

PR #730, opened 2022-03-01
New validation webhook for route uniqueness

kieron-dev:
_**Cross team collab story - for review by AMER team**_

#178 

## What is this change about?
This adds a validating webhook to ensure the uniqueness of routes across all orgs and spaces.

Since the name registry is used now in 3 webhooks: subnamespaceanchor, app and route, we factored out the common logic in a new DuplicateValidator helper.

Route uniqueness is checked using a combination of host, path, domain guid and domain namespace within the cf root namespace. Note we are not using the actual domain name, as this should be immutable, and avoid an extra lookup.

We have introduced a config property to set the root CF namespace in the controllers config, and the various config yamls, as we need to use a name registry in the root namespace to guarantee route uniqueness throughout all orgs and spaces.

We have also moved the duplicate validator and error messages from workloads up a package level to webhooks, so networking and workloads can share the code.

There is no way to update route host/path/domain via the API, so we have an e2e test to trigger the update hook via the add destinations endpoint. This will exit the hook early as new and old route host/path/domain remain unchanged.

The code also fixes e2e flakiness by setting the non-propagation annotation on the cfdomain permissions workaround role binding. See https://github.com/cloudfoundry/cf-k8s-controllers/pull/728#issuecomment-1055346710

No

## Acceptance Steps
See #178 

@cloudfoundry/cf-k8s


api/tests/e2e/routes_test.go
```diff
@@ -0,0 +1,282 @@
...
+			},
+		})).To(Succeed())
+
+		cfDomainBindingName = generateGUID("cfdomain-binding")
+		Expect(k8sClient.Create(context.Background(), &rbacv1.RoleBinding{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      cfDomainBindingName,
+				Namespace: rootNamespace,
+				Annotations: map[string]string{
+					hncv1alpha2.AnnotationNoneSelector: "true",
```
gnovv, line 58: This is very helpful! Thank you for letting us know about the hnc propagations

gnovv (approved):
lgtm! Really like the shared NameRegistry in the webhooks and error type in the handlers

davewalter (approved):
Looks good!

Issue #387, opened 2021-12-22
[Feature]: Platform Operator can configure external protocol, domain, and port for API ingress


tcdowney:
### Blockers/Dependencies

Traditionally Cloud Controller expected that all incoming requests would be coming externally through the Gorouters on ports 80 or 443. So for [URLs that it would give back to clients](https://github.com/cloudfoundry/cloud_controller_ng/blob/5b8455c1a4e9a9f1eda712f5283ad663e24a4912/app/presenters/api_url_builder.rb#L3) only the external protocol and domain were configurable.

In Kubernetes, however, we have less control over ingress and operators have more flexibility in how they configure things. We should make the cf-k8s-controllers API shim have extra configurability to support non-standard ingress ports.

### Background

**As a** Platform Operator
**I want** to be able to configure `externalPort`, `externalFQDN`, and `externalProtocol` in the API Shim's [config](https://github.com/cloudfoundry/cf-k8s-controllers/blob/main/api/config/base/apiconfig/cf_k8s_api_config.yaml) instead of `serverURL`
**So that** I have more flexibility when it comes to how I set up my ingress routing to the API

### Acceptance Criteria

---

## Defaults

* `externalFQDN` - should be required and cannot be defaulted
* `externalProtocol` - should default to `https`
* `externalPort` - should be optional (could default to `""` and be omitted when not present)

---

### `externalPort` is empty

**GIVEN** I have left `externalPort` empty
**WHEN I** observe pagination links and `Job` redirection links (after applying a manifest or deleting an app)
**THEN I** see that there is no port included in the URL. E.g. `https://api.example.com/jobs/:guid`

### `externalPort` is present

**GIVEN** I have set `externalPort` (for this example let's set it to `9123`)
**WHEN I** observe pagination links and `Job` redirection links (after applying a manifest or deleting an app)
**THEN I** see that that the port is included in the URL. E.g. `https://api.example.com:9123/jobs/:guid`

---

### Dev Notes

* These properties should be familiar to operators who have deployed Cloud Controller through the capi-release BOSH release: https://github.com/cloudfoundry/capi-release/blob/develop/jobs/cloud_controller_ng/spec#L342-L347
* **NOT SURE IF STILL RELEVANT** We should centralize our URL building logic more. Right now it's spread across a couple of places, like here: https://github.com/cloudfoundry/cf-k8s-controllers/blob/main/api/apis/space_manifest_handler.go#L72 
[...]

PR #751, opened 2022-03-03
Update app ingress on Kind to HTTPS

davewalter:
No, but this work follows on from #387 

## What is this change about?
As part of #387, we updated the API ingress on Kind to use HTTPS. This PR updates the app ingress on Kind to also use HTTPS. It also updates the e2e tests to allow them to be run against all clusters.

No

## Acceptance Steps
Deploy and run e2e tests.


julian-hj (approved):
LGTM

api/tests/e2e/routes_test.go
```diff
@@ -243,11 +244,13 @@ var _ = Describe("Routes", func() {
...
-				appClient := resty.New()
+				appClient := resty.New().SetTLSClientConfig(&tls.Config{InsecureSkipVerify: true})
 				Eventually(func() int {
 					var err error
-					resp, err = appClient.R().Get(fmt.Sprintf("http://%s.%s", host, appFQDN))
-					Expect(err).NotTo(HaveOccurred())
+					resp, err = appClient.R().Get(fmt.Sprintf("https://%s.%s", host, appFQDN))
+					if err != nil {
+						return 0
+					}
```
georgethebeatle, line 253: Returning 0 as an invalid http status is a bit weird to me. I would do one of the following instead:
- change the function signature to return (int, error) and simply propagate the error. Ginkgo will detect if the second argument is non-nil and fail if it is so.
- change the function signature to accept a gomege.Gomega arg which allows the use of the Succeed mathcher and inline assertions as explained in [the inline docs](https://github.com/onsi/gomega/blob/c8ba5823d8c284a7f8c070cb0f3ec3faf076db04/gomega_dsl.go#L292-L301)
I would lean towards the latter in this case since it will get rid of the if and make the code easier to read

scripts/deploy-on-kind.sh
```diff
@@ -193,8 +193,31 @@ function deploy_cf_k8s_controllers() {
...
+        -nodes \
+        -subj '/CN=*.vcap.me' \
+        -extensions SAN -config <(cat /etc/ssl/openssl.cnf <(printf "[ SAN ]\nsubjectAltName='DNS:*.vcap.me'")) \
+        -days 365
+    fi
+
+    kubectl create secret tls cf-k8s-workloads-ingress-cert \
+      -n cf-k8s-controllers-system \
+      --cert=/tmp/app-tls.crt \
+      --key=/tmp/app-tls.key
```
georgethebeatle, line 220: Does it make sense to extract the secret generation to a separate script? We are already doing this for eirini: https://github.com/cloudfoundry/cf-k8s-controllers/blob/e2e-tests-run-on-non-kind-clusters/scripts/generate-eirini-certs-secret.sh

This would unclutter the script a little bit. Also given that these secret generation scripts are pretty similar maybe it is possible to somehow generalize and dry things up even further.

georgethebeatle (approved):
LGTM, but left some minor comments inline

scripts/deploy-on-kind.sh
julian-hj: Another thing to note is that `kubectl create secret` isn't idempotent, so we should either check to see if the secret already exists, or just delete any existing secret. Otherwise the deploy-on-kind.sh script will fail when we re-run it.

Birdrock: My opinion is that `kubectl create` should never be used in scripts for this very reason. `kubectl apply -f <<EOF...` is a reasonable alternative.

PR #753, opened 2022-03-03
Issues/704

### LINKED ISSUE #704: [Chore] Copy Smoke Tests from cf-for-k8s
### Background

In cf-for-k8s, we developed a small set of [smoke-tests](https://github.com/cloudfoundry/cf-for-k8s/blob/develop/tests/smoke/smoke_test.go) that exercised basic `cf push` functionality. It would be useful to have a similar set of tests for cf-k8s-controllers.

### Action to take

* Copy the `tests/smoke` directory from cf-for-k8s to cf-k8s-controllers.
* Delete the test that validates that docker apps can be pushed to the platform as we do not support them.
* Comment out the part of the test that verifies that application logs can be retrieved as we do not support log retrieval yet.

### Impact

Having a set of smoke tests will provide a fast way to validate a deployment of cf-k8s-controllers.

gnovv:

closes #704 

## What is this change about?

Migrates cf-on-k8s smoke tests

No

## Acceptance Steps

See issue

@julian-hj 
@davewalter

### DIFF
--- a/api/apis/space_handler.go
+++ b/api/apis/space_handler.go
@@ -7,7 +7,6 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
-	"time"
 
 	"github.com/go-logr/logr"
 	"github.com/google/uuid"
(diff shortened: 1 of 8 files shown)


julian-hj (approved):
😲

Issue #548, opened 2022-01-24
[Bug]: Unable to delete invalid `CFRoute` when there is an existing `HTTPProxy` for the FQDN in another namespace


tcdowney:
### What happened?

I was validating https://github.com/cloudfoundry/cf-k8s-controllers/issues/380 and had an app with the same name pushed to two different orgs. Two different `CFRoutes` got created, but the route in `SECOND_ORG`/`SECOND_SPACE` was invalid since the `HTTPProxy` for its FQDN existing in `FIRST_ORG`/`FIRST_SPACE`.

`cf delete-org SECOND_ORG` got stuck with both namespaces in the `Terminating` state. I dug into it and found it was due to a lingering invalid `CFRoute`.

The logs in the `CFRouteController` said this:

```
1.6430607545114973e+09	INFO	controllers.CFRoute	Reconciling deletion of CFRoute/64ce5246-bbdd-432c-8649-b2547a5cc457
1.6430607545115876e+09	ERROR	controllers.CFRoute	existing proxy found in namespace 458d35bd-4c9b-4fd2-bf2d-f0deb162a33b	{"error": "found existing HTTPProxy with FQDN node2.k8s.tim.migration.quest in another space"}
code.cloudfoundry.org/cf-k8s-controllers/controllers/controllers/networking.(*CFRouteReconciler).finalizeCFRoute
	/workspace/controllers/controllers/networking/cfroute_controller.go:188
code.cloudfoundry.org/cf-k8s-controllers/controllers/controllers/networking.(*CFRouteReconciler).Reconcile
	/workspace/controllers/controllers/networking/cfroute_controller.go:95
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.0/pkg/internal/controller/controller.go:114
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.0/pkg/internal/controller/controller.go:311
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.11.0/pkg/internal/controller/controller.go:266
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
[...]

PR #757, opened 2022-03-03
Fix deleting invalid CFRoutes

akrishna90:
#548 

## What is this change about?
Fix deletion of invalid CFRoute when there is an existing HTTPProxy for the FQDN in another namespace

No

## Acceptance Steps
Delete `Validation Webhooks` corresponding to `cf-k8s-controllers` and then follow acceptance steps on #548 

@tcdowney


tcdowney (approved):
Looks fine to me in general. I do kinda feel like the method signature for `getFQDNProxy` is getting a bit unwieldy, but it's private so I don't care too strongly. 

Some alternatives I've thought of were having perhaps having some intermediate helper functions with more descriptive names, but these are already being invoked in private functions like `createOrPatchFQDNProxy` so I'm not sure that actually improves things.

PR #760, opened 2022-03-04
Add logic to wait for HNC role propagation

julian-hj:
No. There is a slack thread here though: https://cloudfoundry.slack.com/archives/C0297673ASK/p1646344103600849

## What is this change about?
- This is basically a workaround for the problem that HNC reports
  readiness before it has completed role binding propagation. If HNC
  fixes that (or we stop using it) this code would no longer be
  necessary.
- After a new SubNamespaceAnchor is ready, we wait until we have rights in
  the namespace before returning OK.
- For now, we test access by listing CFApps with exponential backoff. There might be a more appropriate resource type to use, or we may find a better resource type to test with later (e.g. CFSpace)

No

## Acceptance Steps
- With a newly deployed environment
- Run the following:
  ```
   ```
- This command should work without adding any additional delays between commands.

@gnovv @gcapizzi

### DIFF
--- a/api/repositories/org_repository.go
+++ b/api/repositories/org_repository.go
@@ -6,15 +6,15 @@ import (
 	"fmt"
 	"time"
 
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/types"
-
 	"code.cloudfoundry.org/cf-k8s-controllers/api/authorization"
-	corev1 "k8s.io/api/core/v1"
+	workloadsv1alpha1 "code.cloudfoundry.org/cf-k8s-controllers/controllers/apis/workloads/v1alpha1"
 
+	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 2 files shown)


kieron-dev (commented):
I'd like to see this go in. It addresses the main problem of role propagation of the current user when creating an org or a space.

There are some cases it doesn't catch. E.g. assign someone an OrgManager role, and then list apps in an existing space of that org as the new user. That doesn't seem too important though.

Another approach might be to wrap the user client, and have a retry with exponential backoff on all methods when a Forbidden error is returned, but that feels overly broad.

gcapizzi (approved):
LGTM. As @kieron-dev mentioned, we had tried to solve this by making the client retry on any `403 Forbidden`, but this seems to be enough for now. We'll probably have to get more clever in the future.

gcapizzi:
@julian-hj please close #538 once this goes in!

Issue #702, opened 2022-02-23
[Bug] Org deletion might fail because of  `allowCascadingDeletion` flag not being set


danail-branekov:
Currently we are setting the `allowCascadingDeletion` flag on orgs during [org deletion](https://github.com/cloudfoundry/cf-k8s-controllers/blob/e20616afd0b5dea23eb968eb9896aa801ebf93bb/api/repositories/org_repository.go#L464-L474). 

There are some indications that this might not have immediate effect and HNC controller might deny org deletion due to maybe webhook cache. We have seen that when playing with e2e tests.

We should probably therefore move setting that flag when we actually create the org so that it is already set up whenever we want to delete it.

Here is a sample handler logs from a local e2e failure:

```
        /usr/local/go/src/net/http/server.go:1930
    1.6456165260597837e+09      ERROR   Org Handler     unauthorized to delete org      {"OrgGUID": "1f8fb1e7-2d14-48b4-aa6a-38ddb0cf11dd", "error": "Org forbi
dden: admission webhook \"subnamespaceanchors.hnc.x-k8s.io\" denied the request: The subnamespace 1f8fb1e7-2d14-48b4-aa6a-38ddb0cf11dd is not a leaf and doesn'
t allow cascading deletion. Please set allowCascadingDeletion flag or make it a leaf first."}
    code.cloudfoundry.org/cf-k8s-controllers/api/apis.(*AuthAwareHandlerFuncWrapper).Wrap.func1
        /workspace/api/apis/auth_aware_handler.go:33
    net/http.HandlerFunc.ServeHTTP
        /usr/local/go/src/net/http/server.go:2047
    code.cloudfoundry.org/cf-k8s-controllers/api/apis.(*AuthenticationMiddleware).Middleware.func1
        /workspace/api/apis/authentication_middleware.go:76
    net/http.HandlerFunc.ServeHTTP
        /usr/local/go/src/net/http/server.go:2047
    github.com/gorilla/mux.(*Router).ServeHTTP
        /go/pkg/mod/github.com/gorilla/mux@v1.8.0/mux.go:210
    net/http.serverHandler.ServeHTTP
        /usr/local/go/src/net/http/server.go:2879
    net/http.(*conn).serve
        /usr/local/go/src/net/http/server.go:1930
```

kieron-dev:
e2e tests show that this change succumbs to the HNC propagation latency problem:
1. We create subnamespaceanchor for an org
[...]

PR #761, opened 2022-03-04
Set cascading deletion on orgs during creation

kieron-dev:
#702 

## What is this change about?
To avoid races setting this during deletion, it makes more sense to set this property during creation of orgs.

We also make sure the property is set on the root CF namespace

No

## Acceptance Steps
Delete an org with spaces lots of times and don't see an error about not being able to delete a namespace with child namespaces.

@gcapizzi

### DIFF
--- a/api/repositories/org_repository.go
+++ b/api/repositories/org_repository.go
@@ -145,6 +145,10 @@ func (r *OrgRepo) CreateOrg(ctx context.Context, info authorization.Info, org Cr
 		return OrgRecord{}, err
 	}
 
+	if err := setCascadingDelete(ctx, userClient, org.GUID); err != nil {
+		return OrgRecord{}, err
+	}
+
 	orgRecord := OrgRecord{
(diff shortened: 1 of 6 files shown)


julian-hj:
One minor point to consider--since this change removes the logic at org deletion time, any existing orgs that were created before someone upgrades to this version will experience failures when the user tries to delete them. That was the reason we chose to set this flag at deletion time to begin with.

Having said that, I think we don't support upgrades yet, so I am not sure that it matters.

kieron-dev:
Good point about existing orgs. We could think about introducing a migration framework to cover these sort of things. We have an example from eirini that runs as a job on startup. It applies migrations to the concerned resources and annotates them with a migration version, a bit like database migration frameworks: https://github.com/cloudfoundry-incubator/eirini/blob/master/cmd/migration/main.go.

georgethebeatle (approved):

PR #762, opened 2022-03-04
Make scripts work with the default OSX openssl

davewalter:
No

## What is this change about?
We recently introduced more `openssl` commands into our scripts, but they only work on Linux, or if you have installed the `openssl@3` brew recipe on OSX. This PR introduces a check to see which version of `openssl` is available and adds alternative commands that are compatible with the default OSX version.

No

## Acceptance Steps
Make sure that the scripts work with both versions of `openssl` on OSX, and still work on Linux.

@clintyoshimura


clintyoshimura (approved):
Tested kind installation with LibreSSL 2.8.3 and works
LGTM

PR #773, opened 2022-03-07
[DRAFT] Remove sleeps from e2e and smoke tests

julian-hj:
no
## What is this change about?
- We had delays built into some tests to wait for HNC role propagation
- These delays are theoretically no longer necessary because the API shim waits for permissions to appear before sending the HTTP response

We've run this PR through CI a couple of times with no failures, so we think the change set doesn't introduce new flakiness.

no

## Acceptance Steps
e2e tests should pass

@akrishna90


akrishna90 (approved):
lgtm!

Issue #288, opened 2021-11-25
[Chore] Unify error handling in api


mnitchev:
We recently opened a [draft PR](https://github.com/cloudfoundry/cf-k8s-controllers/pull/695) addressing some pain points around handling errors on different levels of the code as well as representing errors a http responses. It looks like we have reached consensus on this topic so let's implement the pattern we have agreed on across all repos and handlers. 

- [x] remove all `deprecatedXXX` methods and types
- [x] remove all repositories errors and replace them with `apierrors`
- [x] delete all `writeXXXError` utility functions
- [x] delete all `writeXXXResponse` utility functions
- [ ] consider moving the `HandlerResponse` and `AuthAwareHandler` types into a separate package so that their users do not have access to `HandlerResponse.writeTo` function
- [x] Remove the logic handlers that translates repository errors to `apierrors` - repositories are expected to directly return `apierrors`
- [x] Introduce the concept of "read" vs "write" errors (translating `forbidden` to `not found`) that we [had in the draft PR](https://github.com/cloudfoundry/cf-k8s-controllers/pull/695/files#diff-e0ab2ce7bf34fc2a8323d0c177825388890b8e72c681b7795dc81af99776eb3fR199-R206) wherever needed

PR #781, opened 2022-03-08
Common error handling in api handlers

danail-branekov:
https://github.com/cloudfoundry/cf-k8s-controllers/issues/288

## What is this change about?
* Extend the handler functions interface to return a `HandlerResponse`
  and an error (that should usually be on the of errors from the
  `apierrors` package; generic errors are alos supported but would be
  presented as unknown errors)
* Handler functions no longer have access to the `http.ResponseWriter`.
  Instead, the interaction with the writer is carried out by the
  `AuthAwareHandler` wrapper which presents the errors returned by
  handlers in a generic manner.
* Misc improvements
  * All handlers that (except the ones that do not required
    authentication according to the CF API spec) are wrapped by
    `AuthAwareHandler` in order to make use of the common error
    handling.
  * Deprecated `writeStringResponse` utility function is deleted
  * Improvements in handlers unit tests

With this change repositories still return repository errors and
therefore handlers need to translate them to `apierrors`. Next steps
would be to make repositories return `apierrors` so that handlers would
no longer need to do that.

No

## Acceptance Steps
Green tests

@kieron-dev

### DIFF
--- a/api/Dockerfile
+++ b/api/Dockerfile
@@ -11,6 +11,7 @@ COPY controllers/apis controllers/apis/
 COPY controllers/webhooks controllers/webhooks/
 COPY api/actions api/actions/
 COPY api/apis api/apis/
+COPY api/apierrors api/apierrors
 COPY api/authorization api/authorization/
 COPY api/payloads api/payloads/
 COPY api/presenter api/presenter/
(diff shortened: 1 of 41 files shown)


julian-hj (approved):
I got pretty bored about 1/2 way through scrolling around in these changes, so I have to confess that I didn't fully inspect them, but there's a lot of sameness, and it looks OK to me. This is a huge change, so probably you will want to get it merged quickly!

PR #783, opened 2022-03-08
Clean up secrets generation in deploy-on-kind.sh

julian-hj:
no

## What is this change about?
This change uses a somewhat cleaner approach to creating secrets using `kubectl apply`
- adds a helper function to generate tls cert secrets, since we do it twice now
- uses kubectl/customize secretGenerator instead of sed/grep to inject the certificates

no

## Acceptance Steps
e2e tests should exercise this code.
re-running on top of an already deployed cluster would be another good test.

I did this solo, but any of @georgethebeatle @Birdrock @gnovv @davewalter PTAL

### DIFF
--- a/scripts/deploy-on-kind.sh
+++ b/scripts/deploy-on-kind.sh
@@ -81,6 +81,48 @@ if [[ -z "${cluster}" ]]; then
   exit 1
 fi
 
+function create_tls_secret() {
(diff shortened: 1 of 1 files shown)


akrishna90 (approved):
lgtm!

georgethebeatle (approved):
Awsome!

PR #786, opened 2022-03-09
Endpoints are not paths

gcapizzi:
No.

## What is this change about?
An endpoint is defined by a path and a method, we used endpoint constants for paths and duplicated them for operations using different methods on the same path (e.g. `GET` and `DELETE` on the same resource).
This removes the duplication and fixes the constant names.

No.


tcdowney (approved):
Nice 👍  lgtm

Issue #734, opened 2022-03-01
[Chore]: Developers should only be able to get processes by type for apps they have access to


gcapizzi:
### Background

**As an** admin
**I want** to assign the `SpaceDeveloper` role to the users I trust
**So that** they can push apps to their spaces

### Acceptance Criteria

**GIVEN** two spaces `space1` and `space2` exist
**AND** `app1` exists in `space1`
**AND** `app1` has a `web` process
**AND** `app2` exists in `space2`
**AND** `app2` has a `web` process
**AND** `user` has a `RoleBinding` to the `SpaceDeveloper` `ClusterRole` in `space1`

1. **WHEN I** [get the `web` process for `app1`](https://v3-apidocs.cloudfoundry.org/version/3.113.0/index.html#get-a-process) as `user`
  **THEN I** see it working

1. **WHEN I** get the `web` process for `app2` as `user`
  **THEN I** get the following response:
  ```
  HTTP/1.1 404 Not Found
  Content-Type: application/json
  ```
  ```json
  {
    "errors": [
      {
        "code": 10010,
        "title": "CF-ResourceNotFound",
        "detail": "App not found"
      }
    ]
  }
  ```

### Dev Notes

This story is effectively a chore, as the behaviour already matches the expected one because of the call to the app repository, but we still want to introduce a user client in `ProcessRepo.GetProcessByAppTypeAndSpace`. I won't accept this story, it's done as soon as it passes code review.

ToDo: Update the process repository `GetProcessByAppTypeAndSpace` function to use a user-client to retrieve the process. Validate that test coverage is in place.

PR #788, opened 2022-03-09
Backfill tests for /v3/apps/:guid/processes/:type

julian-hj:
**Note:** There is a slight discrepancy between the product behavior and the stated A/C in #734 

The story expects an error like `"detail": "Process not found"`
The product is currently returning `"detail": "App not found. Ensure it exists and you have access to it."`

We believe that the product behavior is more correct, and doesn't leak any information that the user shouldn't have access to.

We also noticed that the app routes endpoint returns the `"App not found"` error message when the app cannot be found, so this behavior is more consistent.

Having said that, please feel free to reject the story if there is some nuance we missed.

#734

## What is this change about?
- implementation was already correct, but tests were missing
- this commit adds handler unit and e2e tests.

No.

## Acceptance Steps
#734 (see note above)

@acosta11

### DIFF
--- a/api/apis/app_handler_test.go
+++ b/api/apis/app_handler_test.go
@@ -1846,6 +1846,17 @@ var _ = Describe("AppHandler", func() {
 			})
 		})
 
+		When("the user lacks access in the app namespace", func() {
+			BeforeEach(func() {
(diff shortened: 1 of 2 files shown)


acosta11 (approved):
LGTM

Issue #627, opened 2022-02-07
[Feature]: API Client can include app information when listing service bindings via GET /v3/service_credential_bindings


tcdowney:
### Blockers/Dependencies

* https://github.com/cloudfoundry/cf-k8s-controllers/issues/603

### Background

**As a** client of the API Shim
**I want** to be able to list service bindings
**So that** I can discover which service instances are bound to which apps **and include app information** so that I don't have to do separate calls to retrieve apps

The CF CLI makes the following request as part of `cf services`:
```
GET /v3/service_credential_bindings?include=app&service_instance_guids=0c024675-7e0b-475c-8332-74986b5001a2%2C896fd3fa-28ec-42ff-b846-531e2ecabdaa
```

### Acceptance Criteria

The following set up applies to all of these scenarios.

1. Created a new Org and Space
2. `cf push app-a` (app guid is `114edd6d-1960-46df-a1dc-0309ce8ed63b`)
3. `cf push app-b` (app guid is `fa1069fb-922a-4a19-8a31-9129a82baff1`)
4. `cf cups service-a` (service instance guid is `5be373e9-309c-4ad9-a705-290525f36bd3`)
5. `cf cups service-b` (service instance guid is `fa0e32b3-e58b-4533-8d61-59ac38b1fa51`)
6. `cf bind-service app-a service-a` (binding guid is `08f1db1a-c7b9-4de2-9afa-5a7e53075d6d`)
7. `cf bind-service app-a service-b` (binding guid is `78a190a3-e3dd-4d42-8158-89aabfbd1673`)
8. `cf bind-service app-b service-b` (binding guid is `9c4cb7fc-eb61-49d8-9c16-a949eae55ed9`)

---

### Include App

**GIVEN** I have created the apps, UPSIs, and bindings defined above
**WHEN I** make the following request to include app info

```
cf curl /v3/service_credential_bindings?include=app
```

**THEN I** get back a `200` response that looks similar to the following (includes info for both app-a and app-b)

```json
{
  "pagination": {
    "total_results": 3,
    "total_pages": 1,
    "first": {
      "href": "https://api.atom-kitten.capi.land/v3/service_credential_bindings?include=app&page=1&per_page=50"
    },
    "last": {
      "href": "https://api.atom-kitten.capi.land/v3/service_credential_bindings?include=app&page=1&per_page=50"
    },
    "next": null,
    "previous": null
  },
  "resources": [
    {
[...]

PR #790, opened 2022-03-09
Only include app info for returned bindings

acosta11:
#627

## What is this change about?
Fix a bug where the `included` block of the list service bindings call would return apps even if no service bindings matched the service instance guid filter. Also includes some diffs from running `make fmt`.

No

## Acceptance Steps
Follow A/C for #627 and query for service bindings with a service instance guid filter of nonexistent instances. Expect no info returned in the `included` block of the response.

paired w/ @julian-hj


julian-hj (approved):
🙆

PR #798, opened 2022-03-14
Return apierrors from repositories

danail-branekov:
https://github.com/cloudfoundry/cf-k8s-controllers/issues/288

## What is this change about?
Return apierrors from repositories. This frees api handlers from interpreting repositories and k8s errors into "UI"-friendly ones. Instead, they can simply return them.

The only exception of not returning the repositories errors as they come is in the cases when handlers explicitly want to mask repo errors, for example, mask `forbidden` as `not found`.

There are a couple of general rules:
* `apierrors` are generally a metadata layer over real errors. Therefore whenever an `apierror` is created, it should be provided with a non-nil cause.
* Both handlers and controllers should return `apierrors`. If they return non-`apierrors`, the errors would be presented as unknown errors in the response.
* Repositories should/must convert raw kubernetes errors to `apierrors`. In order to do that they use the `apierrors.FromK8sError` utility function.
* Handlers should return errors from repos without being interested in them. The only exception is in cases they want to mask errors, e.g. `forbidden` as `not found`. They could use `NotFoundAsUnprocessableEntry`, `ForbiddenAsNotFound` etc. utilities from the `apierror`s package.
* As `apierrors` are a metadata layer over actual errors, tests should be asserting on the error type via either `BeAssignableToTypeOf`, or via the newly introduced `matchers.WrapErrorAssignableToTypeOf`

There are several decisions within this PR that might sound a bit controversial:
* Repositories would handle errors from webhooks (e.g. when we try to create a duplicate object) and return `UnprocessableEntityError` (for example, [the route repository](https://github.com/cloudfoundry/cf-k8s-controllers/blob/9d39638ff7b3b49b08c06ecdf6ed4282c360c3ba/api/repositories/route_repository.go#L334-L342)). So far this handling has been performed by the handlers themselves (for example, the [route handler @ main](https://github.com/cloudfoundry/cf-k8s-controllers/blob/c15946522f9c0b569d65337a4d43ed0868487c2a/api/apis/route_handler.go#L207-L215)). We believe that this translation belongs to the repository, furthermore handlers should not be really interested in the errors coming from repos as along as they make sure to propagate them.
* The domain repository used to return a [generic error](https://github.com/cloudfoundry/cf-k8s-controllers/blob/c15946522f9c0b569d65337a4d43ed0868487c2a/api/repositories/domain_repository.go#L72) which used to be presented as an unknown error. With the current change, the repo returns an `apierrors.NotFoundError`, which is presented as a `domain not found error`. I guess this has been simply overlooked in the past, we should not really get unknown errors.
* There are a few `handleXXXError` utility functions deleted from handlers. They used to implement complex mapping from repository errors, but as we implemented this PR they shrunk down to simply logging an error and returning it. We therefore deleted them and inlined the logging at the places the errors occur. Although this is idiomatic, there is now a bit of duplication in logging in handlers. Maybe we should invent a better logging strategy...
* The `apierrors` constructors should have non-nil cause provided. This cause is used to present the error in the HTTP response and logs. However, for the sake of not panicking, the base `apiError` [handles nil causes](https://github.com/cloudfoundry/cf-k8s-controllers/blob/9d39638ff7b3b49b08c06ecdf6ed4282c360c3ba/api/apierrors/errors.go#L31-L33)

No

## Acceptance Steps
Green tests

@georgethebeatle


kieron-dev (approved):
I've paired on accepting this with Danail. All feedback incorporated and looks good to me.

api/apis/app_handler.go
```diff
@@ -105,7 +100,8 @@ func (h *AppHandler) appGetHandler(authInfo authorization.Info, r *http.Request)
 
 	app, err := h.appRepo.GetApp(ctx, authInfo, appGUID)
 	if err != nil {
-		return nil, h.handleGetAppErr(err, appGUID)
+		h.logger.Error(err, "Failed to fetch app from Kubernetes", "AppGUID", appGUID)
+		return nil, apierrors.ForbiddenAsNotFound(err)
```
davewalter, line 104: The name of the helper function is somewhat confusing, as it makes it look like the code is expecting a forbidden error and is translating it to a not found error. The fact that other types of errors are passed-through without modification is lost without looking at the implementation of the function.

davewalter (approved):
Looks good. The only comment that @acosta11 and I had was around the naming of the error translation functions.

julian-hj (approved):
This is excellent work--it's great to see all those switch statements in the handler layer going away.

One thing I wonder looking at these changes: could we further simplify the code by just having the repo layer return "not found" instead of "forbidden" on any fetch operation? We would lose some fidelity in the logging, but maybe we could just move the logging down into the repo layer to address that. That's just a thought though and shouldn't stop you from merging this commit.

kieron-dev:
> could we further simplify the code by just having the repo layer return "not found" instead of "forbidden" on any fetch operation?

There is a problem where the same repo method's errors are transformed differently by different handler functions. For example, in the app handler, `appSetCurrentDropletHandler()` transforms a forbidden error from `dropletRepo.GetDroplet()` to an [unprocessable entity error](https://github.com/cloudfoundry/cf-k8s-controllers/blob/048991ccfcc4c9c143da75476ab5c96af3982504/api/apis/app_handler.go#L192), whereas the next method `appGetCurrentDropletHandler()` transforms the same error from `dropletRepo.GetDroplet()` to a [not found error](https://github.com/cloudfoundry/cf-k8s-controllers/blob/048991ccfcc4c9c143da75476ab5c96af3982504/api/apis/app_handler.go#L220)

Issue #721, opened 2022-02-28
[Feature]: Every user should be bound to the `cf_user` role in the root "cf" namespace


gcapizzi:
### Background

We store [shared domains](https://docs.cloudfoundry.org/devguide/deploy-apps/routes-domains.html#shared-domains) as `CFDomain` resources in our root namespace. Since shared domains are available to users in all orgs of a CF deployment, we need a way to allow all users of the system to list and get them.

The best way to do this is to automatically assign a `cf_user` role to any user who gets assigned any role on the system, if they don't have it already. The `cf_user` role can be configured with permissions to get and list shared domains. Only admin can create shared domains.

### Acceptance Criteria

**GIVEN** `org` exists
**AND** `space` is in `org`
-  **AND** `user` has no roles yet
  **WHEN I** [assign any role](https://v3-apidocs.cloudfoundry.org/version/3.115.0/index.html#create-a-role) to `user` in `space` or `org`
    **THEN I** see the `cf_user` role has also been assigned to `user`
-  **AND** `user` already is a `cf_user`
  **WHEN I** assign any other role to `user` in `space` or `org`
    **THEN I** see it work
    **AND** `user` is still a `cf_user`

### Dev Notes

Some work in this direction [has already been done](https://cloudfoundry.slack.com/archives/C0297673ASK/p1645779601613699) to be able to write end-to-end tests for routes:
  - a [role](https://github.com/cloudfoundry/cf-k8s-controllers/blob/c15946522f9c0b569d65337a4d43ed0868487c2a/controllers/config/cf_roles/cf_root_namespace_user.yaml) has already been created and we could just rename it;
  - end-to-end tests are currently [explicitly creating the role](https://github.com/cloudfoundry/cf-k8s-controllers/blob/44fc30ded70277a49c0372df5229dcdb0ab9ba6f/api/tests/e2e/e2e_suite_test.go#L182-L201) in `BeforeSuite`, after this change that shouldn't be necessary anymore.

Don't attempt to address the cleanup of this role when other roles (such as OrgUser) are removed from a user.

Reach out to Matt if you need more context on this change.

matt-royal:
Thank you for writing this up, @gcapizzi!

PR #804, opened 2022-03-15
Every user bound to cf_user role in the root ns

Birdrock:
#721 

## What is this change about?
When roles are created (rolebindings are created), users get an additional `cf_user` rolebinding

No.

## Acceptance Steps
Run all tests as normal.

@clintyoshimura @tcdowney @gcapizzi

### DIFF
--- a/api/apis/integration/role_test.go
+++ b/api/apis/integration/role_test.go
@@ -21,11 +21,13 @@ var _ = Describe("Role", func() {
 	)
 
 	BeforeEach(func() {
+		// NOTE: This creates an arbitrary mapping that is not representative of production mapping
 		roleMappings := map[string]config.Role{
 			"space_developer":      {Name: "cf-space-developer"},
 			"organization_manager": {Name: "cf-organization-manager"},
+			"cf_user":              {Name: "cf-user"},
 		}
-		roleRepo := repositories.NewRoleRepo(k8sClient, clientFactory, nsPermissions, roleMappings)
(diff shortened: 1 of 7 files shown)


Birdrock:
@gcapizzi I'm curious about a scenario where we are creating a ServiceAccount role. We implemented this for the case where an SA would also need the `cf_user` role binding, but are unsure whether this is proper.

gcapizzi:
@Birdrock yes, I think this behaviour should be in place for service accounts as well.

api/apis/integration/role_test.go
```diff
@@ -21,11 +21,13 @@ var _ = Describe("Role", func() {
 	)
 
 	BeforeEach(func() {
+		// TODO: This creates an arbitrary mapping that is not representative of production mapping
```
tcdowney, line 24: I assume the mappings are tested for real by the e2e tests when we do things like this:
https://github.com/cloudfoundry/cf-k8s-controllers/blob/a1dd31293d3c0cd59bd3cd792f058ac693c95cd4/api/tests/e2e/apps_test.go#L43-L44

Regarding the "TODO" are we actually planning on changing this here or should we just make this a "NOTE" or something.

Birdrock: The real mappings are exercised in production code, which the e2e tests hit.

Regarding the "TODO" - I'm not sure what to put here, but wanted to highlight that the map keys needed to exist, but their values are irrelevant in the integration tests.

clintyoshimura (approved):
LGTM

PR #805, opened 2022-03-16
Fix error response for GET /v3/packages/:guid

davewalter:
#624

## What is this change about?
This PR fixes the error response for the `GET /v3/package/:guid` endpoint. It also adds test coverage for this endpoint.

No

## Acceptance Steps
See #624

@acosta11


acosta11 (approved):
LGTM

PR #808, opened 2022-03-17
RBAC logic for scale process endpoint

julian-hj:
[#625]

## What is this change about?
- this is primarily a testing change, since the scale and get methods
  for the process repository already use the user client.
- the exception is in the handler code where we were missing a 403->404
  translation.

no

## Acceptance Steps
[#625]

soloing at EOD... @gcapizzi @kieron-dev PTAL if you get a chance

### DIFF
--- a/api/actions/scale_process.go
+++ b/api/actions/scale_process.go
@@ -1,6 +1,7 @@
 package actions
 
 import (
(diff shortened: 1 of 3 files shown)


kieron-dev (approved):
Yep - all looks good. And I'm happy that the Forbidden -> NotFound helper makes the action code neat.

Issue #793, opened 2022-03-11
[Feature]: App Developer can supply their own "type" for UPSI secrets


tcdowney:
### Background

From [the spec](https://github.com/servicebinding/spec#provisioned-service):

>The Secret data SHOULD contain a type entry with a value that identifies the abstract classification of the binding. The Secret type (.type verses .data.type fields) SHOULD reflect this value as servicebinding.io/{type}, substituting {type} with the Secret data type.

Consumers of K8s ServiceBindings look at `type` (and/or `provider`) and take action based on it. For example, the [K8s Service Binding Ruby library](https://github.com/nebhale/client-ruby/blob/main/lib/bindings.rb#L82) filters bindings based on `type` and `provider` so an app or framework might search for `type: database` or `type: mysql`. Or there might exist a buildpack that looks for `type: my-apm-agent` and includes the APM agent as part of the app container based on the presence of that binding.

If all of our bindings are of `type: user-provided` then automatic behavior like this won't work.

**As an** App Developer
**I want** to be able to specify my own `type` when creating a user-provided service instance (UPSI)
**So that** my UPSI bindings can integrate with K8s Service Binding aware libraries and frameworks

### UPSI that specifies a type

**GIVEN** I have authenticated and targeted the CF API Shim
**WHEN I** `cf cups my-upsi -p '{"type": "database", "provider": "amazon", "other-stuff": "arbitrary-stuff"}'`
**THEN I** can see the type I specified reflected on the corresponding secret

```console
$ kubectl -n $(cf space s --guid) get secret a1e646bf-d3e9-4fa3-bf3c-904a39622d6c -o json | jq '.data | map_values(@base64d)'
{
  "other-stuff": "arbitrary-stuff",
  "provider": "amazon",
  "type": "database"
}
```

And the `spec.type` of the `Secret` should be: `servicebinding.io/database`

---

### UPSI that **does not** specify a type

**GIVEN** I have authenticated and targeted the CF API Shim
**WHEN I** `cf cups taco-upsi -p '{"provider": "tacobell", "bun": "think-outside"}'`
**THEN I** can see the that the `spec.type` of the binding is `servicebinding.io/user-provided` and the internal `data.type` is `"user-provided"

```console
$ kubectl -n $(cf space s --guid) get secret 7c8d8506-4ee5-48cf-b09c-d1651d456e55 -o json | jq '.data | map_values(@base64d)'
{
  "bun": "think-outside",
  "provider": "tacobell",
[...]

PR #812, opened 2022-03-17
Users can provide their 

tcdowney:
Issue: https://github.com/cloudfoundry/cf-k8s-controllers/issues/793

## What is this change about?
The "type" of a Service Binding is important and projects that integrate with it look for types with specific names (for example a buildpack might look for bindings of type "mysql" to depend on whether or not to include a mysql client library). Nothing will look for "user-provided" so it's not useful for us to clobber what the user has provided. That said, binding projection typically requires that there be a type so we do need to set _something_ in the case that the developer did not provide a type. So in that case we'll continue to set it to "user-provided".

See  https://github.com/cloudfoundry/cf-k8s-controllers/issues/793 for more details.

Nope

## Acceptance Steps
See  https://github.com/cloudfoundry/cf-k8s-controllers/issues/793

@Birdrock @clintyoshimura


julian-hj (approved):
LGTM

akrishna90 (approved):
Ah, the `type` on service instance create is not always mapped to `type` on the secret. User get to override it if they set it as part of credentials. This looks good to me!

PR #823, opened 2022-03-18
Reduce flakes in space list e2e tests

julian-hj:
no

## What is this change about?
We see a frequent errors in the space list tests that suggest either uninitialized variables in beforeEach or cross-contamination of the environment in parallel tests or both.

This PR attempts to clean those up by making the "list all spaces" test more resilient to extra spaces and by initializing a lot of local variables that could be leaking results from one test to the next in a single thread.

no

## Acceptance Steps
run e2e in the CI a bunch of times. There shouldn't be flakes in the space list tests anymore.

### DIFF
--- a/api/tests/e2e/spaces_test.go
+++ b/api/tests/e2e/spaces_test.go
@@ -114,6 +114,12 @@ var _ = Describe("Spaces", func() {
 		)
 
 		BeforeEach(func() {
+			org1GUID, org2GUID, org3GUID = "", "", ""
+			space11GUID, space12GUID, space13GUID = "", "", ""
+			space21GUID, space22GUID, space23GUID = "", "", ""
+			space31GUID, space32GUID, space33GUID = "", "", ""
+			result = resourceList{}
+
 			var orgWG sync.WaitGroup
 			orgErrChan := make(chan error, 3)
 			query = make(map[string]string)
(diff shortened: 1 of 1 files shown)


api/tests/e2e/spaces_test.go
```diff
@@ -177,19 +183,17 @@ var _ = Describe("Spaces", func() {
 
 		It("lists the spaces the user has role in", func() {
 			Expect(resp).To(HaveRestyStatusCode(http.StatusOK))
-			Expect(result.Resources).To(ConsistOf(
+			Expect(result.Resources).To(ContainElements(
```
gnovv, line 186: Hey @julian-hj I just made the same [change](https://github.com/cloudfoundry/cf-k8s-controllers/pull/813/files#diff-c05917f3c2675facfb5c54a733b878a228ef70c4068de32c11e69b5e1e066808R186), but didn't do your un-grouped Not Contains thing below

gnovv (approved):

api/tests/e2e/spaces_test.go
julian-hj: cool--I think decomposing that is necessary otherwise it would only fail if all three of the unwanted spaces were in the results.

gnovv: Oh wow, I didn't know that's how it worked!

julian-hj: I noticed that you also switched these tests over to use the same certUser as most of the other tests, which makes way more sense.

Issue #754, opened 2022-03-03
[Feature]: CFRoutes FQDN length is validated


Birdrock:
### Background

**As a** user
**I want** my route FQDN validated
**So that** have valid routes

### Acceptance Criteria

**GIVEN** I have a route with a 253 character or less FQDN
**WHEN I** `cf create-route`
**THEN I** the route created successfully

**GIVEN** I have a route with a 254 character or more FQDN
**WHEN I** `cf create-route`
**THEN I** the route fails validation and is not created

### Dev Notes

These same validations should be applied when a CFRoute is made via the k8s api

akrishna90:
CC implementation reference - https://github.com/cloudfoundry/cloud_controller_ng/blob/4a23a419862ef9b7272f411a4435e33fc7305f30/app/messages/domain_create_message.rb

Also max length of each subdomain is 63. i.e. length of hostname should also be restricted to 63 characters (currently we are not enforcing this)
https://github.com/cloudfoundry/cloud_controller_ng/blob/4a23a419862ef9b7272f411a4435e33fc7305f30/app/messages/route_create_message.rb

PR #832, opened 2022-03-21
Add FQDN validation on routes to api and webhook

akrishna90:
#754 

## What is this change about?
Validates FQDN on routes and also improves existing Hostname validation on routes

- added IsFQDN() for validating `host + "." + domain` value.
- IsHost() checks for validity of hostnames on routes
- removed the `HostNameIsInvalidError` as it is not being used by api repositories.

No.

## Acceptance Steps
Follow acceptance steps on #754


gnovv (approved):
lgtm!

I didn't know you could put const definitions inside of functions- that's cool that it's close by where it gets used

Issue #732, opened 2022-03-01
[Feature]: Developers should only be able to list packages they have access to


gcapizzi:
### Background

**As an** admin
**I want** to assign the `SpaceDeveloper` role to the users I trust
**So that** they can push apps to their spaces

### Acceptance Criteria

**GIVEN** the spaces `space1` and `space2` exist
**AND** `app1` exists in `space1`
**AND** `package1` is a package for `app1`
**AND** `app2` exists in `space2`
**AND** `package2` is a package for `app2`
**AND** `user` is a `SpaceDeveloper` in `space1`
**WHEN I** [list packages](https://v3-apidocs.cloudfoundry.org/version/3.115.0/index.html#list-packages) as `user`
**THEN I** only see `package1` in the result

PR #846, opened 2022-03-22
Add RBAC to List Packages endpoint

acosta11:
#732

## What is this change about?
Add RBAC to the List Packages endpoint

No

## Acceptance Steps
- Create a Package in a new space with the cf admin or cluster admin user on a cluster with the shim installed.
- Log in to the shim as a new user, create a package a new space, then curl `/v3/packages`
- See that only the package the new user created is returned in the list of resources

cc @gnovv

### DIFF
--- a/api/actions/scale_app_process_test.go
+++ b/api/actions/scale_app_process_test.go
@@ -4,13 +4,12 @@ import (
 	"context"
 	"errors"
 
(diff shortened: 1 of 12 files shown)


gnovv (approved):
pairing with @acosta11 LGTM!

PR #848, opened 2022-03-23
[TEST PR] Issue/300 run e2e on concourse

kieron-dev:

### DIFF
--- a/scripts/deploy-on-kind.sh
+++ b/scripts/deploy-on-kind.sh
@@ -230,6 +230,7 @@ function install_dependencies() {
       export DOCKER_SERVER="localregistry-docker-registry.default.svc.cluster.local:30050"
       export DOCKER_USERNAME="whatevs"
       export DOCKER_PASSWORD="whatevs"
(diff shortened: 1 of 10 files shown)


scripts/deploy-on-kind.sh
```diff
@@ -231,6 +231,7 @@ function install_dependencies() {
       export DOCKER_SERVER="localregistry-docker-registry.default.svc.cluster.local:30050"
       export DOCKER_USERNAME="whatevs"
       export DOCKER_PASSWORD="whatevs"
+      export KPACK_TAG="localregistry-docker-registry.default.svc.cluster.local:30050/cf-relint-greengrass/cf-k8s-controllers/kpack/beta"
```
julian-hj, line 234: We could probably use a shorter/cleaner tag here, if we're making the change to replace the whole repo tag. We definitely don't need all that cf-relint-greengrass stuff in there when we are storing things locally. We just left them in before because the local registry doesn't care, and it was the easiest `sed` command to write.

kieron-dev: Interestingly, kpack breaks if this doesn't end in 'kpack/beta' and does have at least three slash separated items.

Btw, this is just a test pr to trigger the new pipeline.

kieron-dev:
Closing this test PR

Issue #736, opened 2022-03-01
[Feature]: Developers should be able to create routes in a space only if they are `SpaceDeveloper`s


gcapizzi:
#### Background

**As an** admin
**I want** to assign the `SpaceDeveloper` role to the users I trust
**So that** they can push apps to their spaces

### Acceptance Criteria

**GIVEN** two spaces `space1` and `space2` exist
**AND** `mgr` has a `RoleBinding` to the `SpaceManager` `ClusterRole` in `space1`
**AND** `dev` has a `RoleBinding` to the `SpaceDeveloper` `ClusterRole` in `space1`
* **WHEN I** [create a route](https://v3-apidocs.cloudfoundry.org/version/3.115.0/index.html#create-a-route) in `space1` as `dev`
  **THEN I** see it working
* **WHEN I** create a route in `space1` as `mgr`
  **THEN I** get the following response:
  ```
  HTTP/1.1 403 Forbidden
  Content-Type: application/json
  ```
  ```json
  {
    "errors": [
      {
        "code": 10003,
        "title": "CF-NotAuthorized",
        "detail": "You are not authorized to perform the requested action"
      }
    ]
  }
  ```
* **WHEN I** create a route in `space2` as `dev`
  **THEN I** get the following response:
  ```
  HTTP/1.1 403 Forbidden
  Content-Type: application/json
  ```
  ```json
  {
    "errors": [
      {
        "code": 10003,
        "title": "CF-NotAuthorized",
        "detail": "You are not authorized to perform the requested action"
      }
    ]
  }
  ```

### Dev Notes
- Create route needs RBAC.
- Get domain and get space already have RBAC.
- e2e tests

julian-hj:
@gcapizzi we merged the changes for this story, but wanted to call out that the A/C in the story appears incorrect.

In the case where the user has no rights to the space, we believe that the API should return `422 Unprocessable Entity` in order to be consistent with other endpoints that refer to the space within the request message body. That was what the handler was already doing, so we left that behavior in place.

Please feel free to reject the story if we've got that wrong. Thanks!
cc @clintyoshimura

gcapizzi:
@julian-hj I've just checked this on a CF-on-VMs and you're absolutely right, I'll fix acceptance!

gcapizzi:
@julian-hj I double-checked this during acceptance and it looks like:
* if the space exists but the user is not authorised in it, you'll get `403 Forbidden`;
[...]

PR #849, opened 2022-03-23
Add RBAC to create route endpoint

julian-hj:

[#736]

## What is this change about?
- route creation returns 422 unprocessable entity if the user doesn't
  have rights to see the space
- it returns 403 Forbidden if the user has rights to view the space but
  not create routes.

No.

## Acceptance Steps
#736 
Caveat: the A/C on the issue appears to be incorrect and calls for a 404 instead of a 422 in the case where the user lacks view permissions on the space.

@clintyoshimura


clintyoshimura (approved):
LGTM

PR #851, opened 2022-03-23
Volume mount serviceBinding credentials onto workloads

### LINKED ISSUE #746: [Feature]: App Developer can see service binding credentials are volume mounted onto their workload containers
### Blockers/Dependencies

* Blocker: https://github.com/cloudfoundry/cf-k8s-controllers/issues/506

### Background

Cloud Native Buildpacks and some app frameworks (such as [Spring Cloud Bindings](https://github.com/spring-cloud/spring-cloud-bindings)) are adopting the [K8s ServiceBinding spec](https://github.com/servicebinding/spec#workload-projection) model of volume mounted secrets. We currently are providing apps access to these via the `VCAP_SERVICES` environment variable ([see this issue](https://github.com/cloudfoundry/cf-k8s-controllers/issues/462)) for backwards compatibility reasons, but want to support these newer developments in the ecosystem as well.

We don't need to implement this ourselves. Theoretically any reconciler that adopts the spec can update Pod-specable workloads to do this. Let's look into using [this Service Binding reconciler](https://github.com/vmware-labs/service-bindings) to do this and install it in our dev envs (Kind) and hack installation scripts. If we set up the `spec.workload` correctly in https://github.com/cloudfoundry/cf-k8s-controllers/issues/506 this reconciler should be able to modify the `StatefulSets` that Eirini Controller is producing†. Then we can document that if developers/operators want this behavior they can use that reconciler or install a comparable one. If that all works well then it means we can avoid having to develop a partial implementation of the projection spec in our own controllers.

**As an** App Developer
**I want** service binding credentials to be projected into my running workloads via volume mounts
**So that** I my K8s Service Binding-aware apps can find service binding credentials in the way they expect

_† Let's confirm there is no contention between this reconciler and Eirini Controller_

---

#### Documentation
**GIVEN** I have a cluster available
**WHEN I** Follow our installation documentation
**THEN I** Can install a Service Binding reconciler and see the K8s `ServiceBindings` associated with our `CFServiceBindings` get reconciled by it

---

#### Functionality
**GIVEN** I have installed cf-k8s-controllers and have included [a Service Binding reconciler](https://github.com/vmware-labs/service-bindings)
**WHEN I** bind a user-provided service instance to my app (`cf bind-service MY_APP MY_INSTANCE`)
**THEN I** see that eventually there exists a K8s `ServiceBinding` (done in https://github.com/cloudfoundry/cf-k8s-controllers/issues/506)
**AND I** see that the credentials for that binding are volume mounted into the `Pods` for my app

---

### Dev Notes

* https://github.com/vmware-labs/service-bindings
* https://github.com/servicebinding/spec#workload-projection
* We can use this Spring PetClinic app to test: https://github.com/vmware-labs/service-bindings/tree/main/samples/spring-petclinic

akrishna90:
closes #746 

## What is this change about?
Volume mount serviceBinding credentials onto workloads
- Add ClusterRole required by the ServiceBinding Controller to fetch the referenced CFServiceBinding object
- Add installation of ServiceBinding controller to `deploy-on-kind` script.
- Update README with manual installation instructions

no

## Acceptance Steps

see issue

@gnovv

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,21 @@ subnamespaces.
 kubectl hns config set-resource secrets --mode Propagate
 ```
 
+---
+## Optional: Install Service Bindings Controller
+
+Cloud Native Buildpacks and other app frameworks (such as [Spring Cloud Bindings](https://github.com/spring-cloud/spring-cloud-bindings)) are adopting the [K8s ServiceBinding spec](https://github.com/servicebinding/spec#workload-projection) model of volume mounted secrets. 
+We currently are providing apps access to these via the `VCAP_SERVICES` environment variable ([see this issue](https://github.com/cloudfoundry/cf-k8s-controllers/issues/462)) for backwards compatibility reasons.
+We would also want to support the newer developments in the ServiceBinding ecosystem as well.
+
+We are not implementing this ourselves but allowing controller that adopts the [ServiceBinding Spec](https://github.com/servicebinding/spec) that volume mounts secrets to workload containers to be used. 
+We have tested with [this Service Binding reconciler](https://github.com/vmware-labs/service-bindings). Install it using 
+
+```sh
+kubectl apply -f https://github.com/vmware-tanzu/servicebinding/releases/download/v0.7.1/service-bindings-0.7.1.yaml
+```
+
+---
 # Installation
 ## Configure cf-k8s-controllers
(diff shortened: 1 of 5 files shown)


gnovv (approved):
pairing with @akrishna90 LGTM!

PR #852, opened 2022-03-23
Adds Case-Insensitive CFApp spec.Name Validation

### LINKED ISSUE #723: [Feature]: App name uniqueness should be case-insensitive
### Blockers/Dependencies

_No response_

### Background

App names are [case insensitive](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/db/migrations/20141210194308_add_name_col_v3_apps.rb#L8) in CF for VMs. This applies to both uniqueness and API queries. So, for example, when I create an app with `cf push MyApP` and then do a `cf curl /v3/apps?names=myapp` it will return in the results.

```
11:40 $ cf curl /v3/apps?names=myapp
{
  "pagination": {
    "total_results": 1,
    "total_pages": 1,
    ...
  },
  "resources": [
    {
      "guid": "f42117d6-2192-4523-b458-e825bcfe71eb",
      "created_at": "2022-01-07T18:39:33Z",
      "updated_at": "2022-01-07T18:40:15Z",
      "name": "MyApP",
      "state": "STARTED",
      ...
      },
      "links": {
        ...
      }
    }
  ]
}
```

**As an** App Developer
**I want** Apps with the same case-insensitive name to be considered the same app
**So that** the platform functions the same as CF for VMs

#### Conflicting App Names During creation
**GIVEN** I have an App in my Space named `joe`
**WHEN I** try to create a new app named `JOE` using `cf create-app JOE`
**THEN I** get a response from the CLI like this

```
cf create-app JOE
Creating app JOE in org org / space space as admin...
App with the name 'JOE' already exists.
OK
```

And a `422` error like this back from `POST /v3/apps`

```json
{
  "errors": [
    {
      "code": 10016,
      "detail": "App with the name 'JOE' already exists.",
      "title": "CF-UniquenessError"
    }
  ]
}
```

---

### Dev Notes
* Validation is already implemented in https://github.com/cloudfoundry/cf-k8s-controllers/blob/main/controllers/webhooks/workloads/cfapp_validation.go so we likely just need to modify the lease name in that code
* Querying will be handled in https://github.com/cloudfoundry/cf-k8s-controllers/issues/417

gnovv:

closes #723
## What is this change about?

- Adds Case-Insensitive CFApp spec.Name Validation
- Updates CFApp spec.Name regex to allow upper-case characters

Co-authored-by: Ashwin Krishna <krishnaas@vmware.com>

No.
## Acceptance Steps

see issue
currently there is a bug that prevents you from creating app names with upper-case letters initially.
You must validate by creating a lower-case named app, then trying to push the upper-case version afterwards.
We created a [bug issue](https://github.com/cloudfoundry/cf-k8s-controllers/issues/853) to track this.

@akrishna90 
@acosta11

### DIFF
--- a/controllers/apis/workloads/v1alpha1/cfapp_types.go
+++ b/controllers/apis/workloads/v1alpha1/cfapp_types.go
@@ -24,7 +24,7 @@ import (
 // CFAppSpec defines the desired state of CFApp
 type CFAppSpec struct {
 	// Name defines the name of the app
-	// +kubebuilder:validation:Pattern="^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$"
+	// +kubebuilder:validation:Pattern="^[a-zA-Z0-9]([-a-zA-Z0-9]*[a-zA-Z0-9])?(\\.[a-zA-Z0-9]([-a-zA-Z0-9]*[a-zA-Z0-9])?)*$"
 	Name string `json:"name"`
 
 	// Specifies the current state of the CFApp
--- a/controllers/apis/workloads/v1alpha1/cfapp_types.go
+++ b/controllers/apis/workloads/v1alpha1/cfapp_types.go
@@ -24,7 +24,8 @@ import (
 // CFAppSpec defines the desired state of CFApp
 type CFAppSpec struct {
 	// Name defines the name of the app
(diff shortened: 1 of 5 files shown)


acosta11:
Talking to Tim, we may actually want to update our regex to not directly match the runtime app_model. On VMs, this creates an issue for the default route creation where the app name is directly passed as the hostname.

So for a consistent experience we would either want to restrict app names to a more similar character set for default route compatibility or enhance the logic for default route generation.

Issue #724, opened 2022-02-28
[Feature]: Operator can identify the namespaces and pods created by CF


julian-hj:
### Background

**As a** k8s operator
**I want** to see which namespaces and pods/statefulsets were created by my cf installation
**So that** I can easily distinguish orgs from spaces
**And** I can distinguish cf things from other things installed on my cluster

There is an ADR where we recorded our intent to simply generate prefixed guids, rather than trying to make the guids in our api conform to standard GUID format: https://github.com/cloudfoundry/cf-k8s-controllers/blob/main/docs/architecture-decisions/0012-resource-name-prefixes.md

### Acceptance Criteria

**GIVEN** A cf-on-k8s installation
**WHEN I** create a new org
**THEN I** see that the resulting k8s namespace is prefixed with `cf-org-`
**AND**
**WHEN I** create a new space
**THEN I** see that the resulting k8s namespace is prefixed with `cf-space-`
**AND**
**WHEN I** push an app and run it
**THEN I** see that the name of the resulting k8s statefulset is prefixed with `cf-proc-`
**AND I** see that the name of the resulting k8s service is prefixed with `cf-route` 

### Dev Notes

Initial spike work for this story is on this commit: https://github.com/cloudfoundry/cf-k8s-controllers/commit/1499a854f678cd3cb90228dee7f23d7e3856339f
The spike did not cover route guids.

julian-hj:
A/C: passed. The only exception is that the cf-route prefix ends up on the `httpproxy` object, and not the service. We could [fix that in the route controller](https://github.com/cloudfoundry/cf-k8s-controllers/blob/main/controllers/controllers/networking/cfroute_controller.go#L467) since we are already generating a name with a prefix, but it's not clear that that is important.

PR #854, opened 2022-03-24
Add prefix to org, space, process, and route GUIDS

clintyoshimura:
#724 

## What is this change about?
Add prefixes to orgs, spaces, processes, and routes to allow operators to identify resources created by cf

No

## Acceptance Steps
See issue

@julian-hj

### DIFF
--- a/api/apis/job_handler.go
+++ b/api/apis/job_handler.go
@@ -70,7 +70,7 @@ func (h *JobHandler) RegisterRoutes(router *mux.Router) {
 
 func parseJobGUID(jobGUID string) (string, string, bool) {
 	// Match job.type-GUID and capture the job type and GUID for later use
-	jobRegexp := regexp.MustCompile("([a-z_-]+[.][a-z_]+)-([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})")
+	jobRegexp := regexp.MustCompile("([a-z_-]+[.][a-z_]+)-(?:cf-[a-z]+-)?([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})")
 	matches := jobRegexp.FindStringSubmatch(jobGUID)
 
 	if len(matches) != 3 {
--- a/api/apis/org_handler.go
+++ b/api/apis/org_handler.go
@@ -7,7 +7,6 @@ import (
 	"net/url"
 
 	"github.com/go-logr/logr"
-	"github.com/google/uuid"
 	"github.com/gorilla/mux"
 	"github.com/gorilla/schema"
 	controllerruntime "sigs.k8s.io/controller-runtime"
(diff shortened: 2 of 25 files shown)


api/apis/job_handler.go
```diff
@@ -70,7 +70,7 @@ func (h *JobHandler) RegisterRoutes(router *mux.Router) {
 
 func parseJobGUID(jobGUID string) (string, string, bool) {
 	// Match job.type-GUID and capture the job type and GUID for later use
-	jobRegexp := regexp.MustCompile("([a-z_-]+[.][a-z_]+)-([0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})")
+	jobRegexp := regexp.MustCompile("([a-z_-]+[.][a-z_]+)-([a-z]*-?[a-z]*-?[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})")
```
kieron-dev, line 73: This regex is becoming quite complex. I wonder if the new optional part for matching `cf-xxx-` could be made clearer by using a non-capturing group. E.g. `(?:cf-[a-z]+-)?` instead of `[a-z]*-?[a-z]*-?`. Or if there is a finite (for now) list of prefixes, it could be more explicit with `(?:cf-(?:org|space|route|proc)-)?` maybe?

api/apis/org_handler.go
```diff
@@ -58,7 +59,7 @@ func (h *OrgHandler) orgCreateHandler(info authorization.Info, r *http.Request)
 	}
 
 	org := payload.ToMessage()
-	org.GUID = uuid.NewString()
+	org.GUID = OrgPrefix + uuid.NewString()
```
kieron-dev, line 62: I notice we are generating the GUID in 3 different layers: handler for orgs and spaces, repo for routes and the app reconciler for processes. Maybe this PR is a good opportunity to move the org and space GUID generation to the repo layer to make things more uniform. The process GUID will have to stay in the app reconciler.

kieron-dev (commented):
This looks generally fine to me. Just see what you think about suggestions about the hairy regex and whether we should move the GUID generation for orgs and spaces.

api/tests/e2e/service_instances_test.go
```diff
@@ -123,13 +123,15 @@ var _ = Describe("Service Instances", func() {
 			})
 
 			It("deletes the service instance", func() {
```
kieron-dev, line 125: This appears unrelated to the story. Maybe better in a separate PR?

api/apis/job_handler.go
clintyoshimura: Updated to use the `(?:cf-[a-z]+-)?` format. It seems like that would be the best solution without having to add to the list of types if we add more.

api/apis/org_handler.go
clintyoshimura: Moved the Org/space uuid generation to the repository layer.

api/tests/e2e/service_instances_test.go
clintyoshimura: This was only because i couldn't get it to run locally. It doesn't seem to be an issue when git runs the e2e test so I reverted it.

matt-royal (approved):
We paired to wrap this up and address the feedback. LGTM

Issue #755, opened 2022-03-03
[Feature]: CFRoutes validates path


Birdrock:
### Background

**As a** app developer
**I want** route paths to be validated
**So that** I can reach my application

### Acceptance Criteria

Validation Rules:
* Needs to be valid URI
* Cannot be a single `/`
* Must have a beginning `/`
* Path cannot contain a `?`
* Path cannot be more than 128 characters

**GIVEN** a route with a valid path
**WHEN I** `cf create-route`
**THEN I** see the route is created

**GIVEN** a route with an invalid path
**WHEN I** `cf create-route`
**THEN I** see the route fails validation and is not created

**GIVEN** a route with a valid path
**WHEN I** `kubectl apply -f <route file>`
**THEN I** see the route is created

**GIVEN** a route with an invalid path
**WHEN I** `kubectl apply -f <route file>`
**THEN I** see the route fails validation and is not created

PR #859, opened 2022-03-24
Validate CFRoute path

Birdrock:
#755 

## What is this change about?
Validates routes according to Cloud Controller rules.

It should not break anything currently existing. Previously valid routes may fail validation on update; this would require updating to a valid route.

## Acceptance Steps
* Run tests as normal.
* `cf create-route ... --path <some path>`
* `kubectl apply -f <some CFRoute>`

@tcdowney @acosta11 @davewalter


davewalter (approved):

Birdrock:
Finding some interesting behavior with manual testing of the API, so I've marked this as a draft.

PR #864, opened 2022-03-25
Add epic issue template

gcapizzi:
This will make it slightly easier to create new epics. I wasn't able to test the file so not sure it will actually work 🤷


kieron-dev (approved):
Approving so we can see the effect on the repo templates.

PR #868, opened 2022-03-25
Eliminate unused privileged clients

davewalter:
Not directly, but it is related to all the recent auth work we've done.

## What is this change about?
This PR updates the constructor functions for repositories that no longer use the privileged client directly to stop accepting it as a parameter.

It also updates more of the e2e test helper functions to use the *WithOffset() variants of `Expect()` and `Eventually()`.

No

## Acceptance Steps
Test pass.

@cloudfoundry/cf-k8s


gnovv (approved):
LGTM! I like the ExpectWithOffset changes too

Issue #741, opened 2022-03-02
[Feature]: Developers should only be able to apply manifests to spaces they have access to


gcapizzi:
Overlap with #699 

### Background

**As an** admin
**I want** to assign the `SpaceDeveloper` role to the users I trust
**So that** they can push apps to their spaces

### Acceptance Criteria

**GIVEN** two spaces `space1` and `space2` exist
**AND** `app1` exists in `space1`
**AND** `app2` exists in `space2`
**AND** `mgr` has a `RoleBinding` to the `SpaceManager` `ClusterRole` in `space1`
**AND** `dev` has a `RoleBinding` to the `SpaceDeveloper` `ClusterRole` in `space1`
* **WHEN I** [apply the following manifest](https://v3-apidocs.cloudfoundry.org/version/3.116.0/index.html#apply-a-manifest-to-a-space) to `space2` as `dev`:
  ```
  version: 1
  applications:
  - name: app2
    env:
      FOO: bar
  ```
  **THEN I** get the following response:
    ```
    HTTP/1.1 404 Not Found
    Content-Type: application/json
    ```
    ```json
    {
      "errors": [
        {
          "code": 10010,
          "detail": "Space not found",
          "title": "CF-ResourceNotFound"
        }
      ]
    }
    ```
* **WHEN I** apply the following manifest to `space1` as `mgr`:
  ```
  version: 1
  applications:
  - name: app1
    env:
      FOO: bar
  ```
  **THEN I** get the following response:
    ```
    HTTP/1.1 403 Forbidden
    Content-Type: application/json
    ```
    ```json
    {
      "errors": [
        {
          "code": 10003,
          "detail": "You are not authorized to perform the requested action",
          "title": "CF-NotAuthorized"
        }
      ]
    }
    ```
* **WHEN I** apply the following manifest to `space1` as `mgr`:
  ```
  version: 1
  applications:
  - name: app1
     processes:
    - type: web
      instances: 3
  ```
  **THEN I** get the following response:
    ```
    HTTP/1.1 403 Forbidden
    Content-Type: application/json
    ```
    ```json
    {
      "errors": [
        {
          "code": 10003,
          "detail": "You are not authorized to perform the requested action",
          "title": "CF-NotAuthorized"
        }
      ]
    }
    ```

### Dev Notes

This story should introduce authentication to the following repository methods:

* `AppRepo.GetAppByNameAndSpace`
[...]

PR #886, opened 2022-03-29
Fix RBAC issue for apply manifest endpoint

akrishna90:

No, but related to the changes introduced in #741  

## What is this change about?

Fix RBAC issue for apply manifest endpoint

- while invoking the `ListRoutesForApp()` function from `apply_manifest_action`, we were incorrectly passing `appGUID` as `spaceGUID` and vice versa. this PR fixes this issue

- `privilegedClient` was used in the repo method until the changes to address issue #741 was merged. We think `privilegedClient` dealt with non-existent spaces differently, causing the bug to fly under the radar. 

No

## Acceptance Steps

`cf push` should now succeed

@tcdowney @Birdrock @gnovv

### DIFF
--- a/api/actions/apply_manifest.go
+++ b/api/actions/apply_manifest.go
@@ -64,7 +64,7 @@ func (a *ApplyManifest) checkAndUpdateDefaultRoute(ctx context.Context, authInfo
 		return nil
 	}
 
-	existingRoutes, err := a.routeRepo.ListRoutesForApp(ctx, authInfo, appRecord.SpaceGUID, appRecord.GUID)
+	existingRoutes, err := a.routeRepo.ListRoutesForApp(ctx, authInfo, appRecord.GUID, appRecord.SpaceGUID)
 	if err != nil {
 		return err
 	}
--- a/api/tests/e2e/e2e_suite_test.go
+++ b/api/tests/e2e/e2e_suite_test.go
@@ -121,9 +121,10 @@ type manifestResource struct {
 }
 
 type applicationResource struct {
-	Name      string                               `yaml:"name"`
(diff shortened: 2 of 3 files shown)


api/actions/apply_manifest.go
```diff
@@ -64,7 +64,7 @@ func (a *ApplyManifest) checkAndUpdateDefaultRoute(ctx context.Context, authInfo
 		return nil
 	}
 
-	existingRoutes, err := a.routeRepo.ListRoutesForApp(ctx, authInfo, appRecord.SpaceGUID, appRecord.GUID)
+	existingRoutes, err := a.routeRepo.ListRoutesForApp(ctx, authInfo, appRecord.GUID, appRecord.SpaceGUID)
```
tcdowney, line 67: Is there any test that could have caught this?

akrishna90: We do not, both `e2e` and `integration` tests have cases that always satisfies [this condition](https://github.com/cloudfoundry/cf-k8s-controllers/blob/d789b4b222940e961f9864b05b93317dab3d0db1/api/actions/apply_manifest.go#L63) and returning back. 
I'll add a new test cases for this scenario

akrishna90: okay, I have added an e2e test that beats the condition above and tests this code path.

Birdrock (approved):
/lgtm

api/actions/apply_manifest.go
gcapizzi: The e2e probably doesn't hurt but this is the typical mistake that should have been caught by unit tests. The unit test for this action never checks arguments for mocked methods though 😞  which is something we should definitely fix.

Issue #601, opened 2022-01-31
[Chore]: Backfill unit tests for `repositories` package


matt-royal:
### Background

We've been feeling the pain of our integration-only strategy lately, and have been unable to test some common failure cases. We've already established patterns for testing the controllers with k8s client fakes, and we're confident we can do the same here.

### Action to take

Find the `// untested` comments in the `repositories` package and backfill unit tests for those cases.

### Impact

This will give us more confidence that we're properly implementing error handling, and also help us avoid regressions of untested behavior

### Dev Notes
- Start with CFServiceInstance and write chores for others

gcapizzi:
I would generally advice against mocking types we don't own, in this case the `controller-runtime` client. We'd be making assumptions about how the client (and the Kubernetes API) work, which might be wrong. It would allow us to test some particularly hard to reach error cases, but I could only find 3 instances of `// untested` in the `repositories` package, so maybe it's not that big of a deal? Overall I don't think it warrants the introduction of a new test suite.

matt-royal:
Thank you, @gcapizzi. I'll discuss this with the AMER team today. I agree that, if we add these, it only makes sense to test the `// untested` branches.

gcapizzi:
@matt-royal someone also pointed out that some `// untested` branches have been deleted because we have a [coverage report](https://codeclimate.com/github/cloudfoundry/cf-k8s-controllers) now, not so easy to spot them via that webpage though 😕  I'm sure there's a way to extract all uncovered lines and extract error checks from them. We should probably be looking at all our uncovered lines anyway, but that is not necessarily related to this issue.

julian-hj:
We created a POC to add unit level tests to the existing suite, but it didn't get a lot of support, so we abandoned it. @georgethebeatle has another approach to try by adding an interface layer on top of the k8s cliet, but for now we will close this issue.

PR #887, opened 2022-03-29
[DRAFT] Add unit level tests for App repository

julian-hj:
[#601]

## What is this change about?
We initially added units only to the App repo to establish the pattern. If folks feel this is a good way to go, we will make additional chores to fill out the other repos.

- This allows us to reach better code coverage for error cases
- We added interfaces for the namespace retriever and nsperms objects in
  order to make them fakeable.
- We added fakes for the client factory and client interfaces also.
- Minor typo and format fixes.

No

## Acceptance Steps
It's a chore.

@matt-royal @gcapizzi @georgethebeatle @kieron-dev @danail-branekov


api/authorization/namespace_permissions.go
```diff
@@ -24,29 +24,29 @@ type IdentityProvider interface {
...
-type NamespacePermissions struct {
+type namespacePermissions struct {
 	privilegedClient client.Client
 	identityProvider IdentityProvider
 	rootNamespace    string
 }
 
-func NewNamespacePermissions(privilegedClient client.Client, identityProvider IdentityProvider, rootNamespace string) *NamespacePermissions {
-	return &NamespacePermissions{
+func NewNamespacePermissions(privilegedClient client.Client, identityProvider IdentityProvider, rootNamespace string) *namespacePermissions {
```
kieron-dev, line 33: Returning a private struct is interesting. Maybe we should do more of this.

api/repositories/app_repository_test.go
```diff
@@ -28,6 +31,263 @@ const (
 	CFAppStoppedState  = "STOPPED"
 )
 
+//counterfeiter:generate -o fake -fake-name ClientWithWatch sigs.k8s.io/controller-runtime/pkg/client.WithWatch
+var _ = Describe("AppRepository Units", func() {
```
kieron-dev, line 35: I guess we'd split the integration tests out into a separate dir when this goes out of draft?

api/repositories/app_repository_test.go
```diff
@@ -28,6 +31,263 @@ const (
...
+		appRepo = NewAppRepo(&fakeNamespaceRetriever, &fakeFactory, &fakeNsPerms)
+	})
+
+	Describe("GetApp", func() {
+		var retErr error
+		JustBeforeEach(func() {
+			_, retErr = appRepo.GetApp(testCtx, authInfo, "some-app-name")
+		})
+
+		When("the client factory reports an error", func() {
```
kieron-dev, line 61: I would aim to cover as much as possible with the unit tests. So as well as covering the error paths, e.g. for GetApp, we should also:
- check that the namespaceRetriever is invoked with the correct args
- check the expected authInfo is passed to the userClientFactory
- check that the expected namespace and name are requested in the userClient.Get

We saw a case recently where a couple of string arguments were passed in the wrong order, so as well as covering the untested error paths, these tests could guarantee calling out correctly.

kieron-dev (commented):
This generally looks good. As I put in a file comment, I think we should go for full coverage at this level and not skip the happy paths and the argument checking.

I'd also like to see what happens with the associated integration test. Is there duplication that can be removed?

api/repositories/app_repository_test.go
julian-hj: We were on the fence about that. We are currently in a strange state because we've been using integration level tests as unit tests, and measuring coverage against those tests. This PR backfills some testing to get to unreachable code, but doesn't attempt to get to full coverage on the unit test side.

Ideally, we would stop using integration tests as unit tests, and write unit tests that achieve full coverage on their own, and move the env tests into a separate suite, but that would take a lot of time, so in the interim we settled on putting both in the same file to make it easier to understand how they relate to one another to achieve test coverage.

So, yeah basically, I agree that the two kinds of tests should be separated, but I am less sure that immediately is the right time frame, unless we were going to tear off the bandage and rewrite all the tests.

matt-royal: That's something we wanted feedback on. In the past we've put unit and integration tests in separate packages. This makes it harder to find all of the tests and requires you to remember the structure of each different package. Having the unit tests in the same file with the integration tests solves this problem, but at the cost of slightly higher overhead per test (due to the BeforeEach targeted at test-env tests)

api/authorization/namespace_permissions.go
matt-royal: I've been wanting to do this in the handlers as well. Having the type be private means that you can't instantiate it incorrectly outside of the constructor.

matt-royal:
@kieron-dev Thank you for the feedback on this PR. We added a discussion topic to the sync meeting tomorrow about this PR and the future work we'll spin out for other repos.

RE full unit test coverage: Personally, I would prefer to lean on the test-env integration tests for repositories where possible, and only fall back on the unit tests when we have a case that can't be exercised via integration. My rationale is that the test-env tests are easier to read and alter than the unit tests, and are a more realistic test of behavior. The downside is the occasional flake and somewhat slower tests, but for me this still balances out in the integration test's favor.

I agree with your comment about testing that the correct arguments are passed to the fakes in unit tests, and I'm going to modify this PR to add those.

matt-royal:
I pushed an update that adds the assertions you mentioned, @kieron-dev

api/authorization/namespace_permissions.go
georgethebeatle: I have always like this approach, but have avoided it for years, because it used to yield a [golint linter warning](https://github.com/golang/lint/blob/6edffad5e6160f5949cdefc81710b2706fbcd4f6/testdata/unexp-return.go#L11). I tried to find out one more time why exactly returning an unexported type is "annoying" and found [this issue](https://github.com/golang/lint/issues/210). However I am not convinced that it is evil if you return an interface implementation. So I think I agree with you, just putting these references here for additional context.

api/repositories/app_repository_test.go
georgethebeatle: Even if we do not convert integrations to units right away I still think it is good to have a separation between those two. If they are separated we can agree on some convention of what goes where. For example we might agree that we only ever put error cases in the units. But if there are no integrations and units it is hard to even communicate that and I am afraid these suites will continue growing in a diffuse manner.

api/authorization/namespace_permissions.go
danail-branekov: I think I found a valid use case for the annoyance argument. Imagine the following code

```go
package foo

type thing struct {...}

func NewThing() *thing{....}
```

You need to testdrive it, so you create a test package (`foo_test`):

```go
package foo_test
....

var myThing ???? // how do you var the thing to test so that you can use it in `BeforeEach`, etc?

```

Remember that the golang convention is that interfaces are defined wherever needed, not where they are implemented, therefore package `foo` might not have the `Thing` interface at all. 

In this particular case, imagine you had to write unit tests for the `namespacePermissions` struct. How would you do that?

matt-royal: In the extreme case, you can create the interface in the test file itself, since that's where it's used. However, in our codebase I think it would be reasonable to use the expected interface type in the test, since that's how the code will actually be used in practice. In this case that would mean `var nsPerm repositories.NamespacePermission`. If that feels awkward since the interface is declared in a different package, then we could move common/shared interfaces into a shared package that contains common interfaces used throughout the codebase.

Personally, I prefer this approach to having to worry about objects being constructed incorrectly.

api/repositories/app_repository_test.go
matt-royal: I forgot to comment here that I pushed a change to add the arg checks you mentioned. I chose to skip the assertions against the userClient, though, since the test-env tests already verify that the correct data is being fetched. The test-env tests will also survive a refactor (e.g adding an indexed field to a resource), where the units will require changes.

api/authorization/namespace_permissions.go
danail-branekov: > Personally, I prefer this approach to having to worry about objects being constructed incorrectly.

Absolutely agree! We should be using constructors to create objects and make all objects' field private to guarantee that, there is no argument about it.

What I doubt whether making structs private (i.e. make their names start with a lowercase letter) and returning pointers to those structs be annoying for testing. For example

```go
func NewThing() *Thing
```

vs

```go
func NewThing() *thing
```

> you can create the interface in the test file itself
Feels like a workaround for the sake of testing to me. What about cases when a struct implements a couple of interfaces? Would you need the "productive" interfaces defined wherever they are needed and a test interface that combines them all?

> then we could move the interfaces into a shared package that contains common interfaces used throughout the codebase.
I do not think that this is quite golang idiomatic though.

api/repositories/app_repository_test.go
georgethebeatle: I hear that some of the reasoning behind mixing the pure units with the test-env integrations might be that this way it is easier to see combined code coverage in the IDE. However not everyone uses the same IDE/editor, while we do have code climate [coverage](https://codeclimate.com/github/cloudfoundry/cf-k8s-controllers) setup that is merging coverage from all tests no matter if they are in the same package or not (with the exception of e2es). To me this is a better place to look for uncovered branches as it is central and we can create chores with links to the reports.

julian-hj:
Closing this PR, as discussed in the SIG meeting, due to an absence of proponents.

matt-royal:
This PR has been useful for discussion, but we haven't reached agreement on it so I'm tempted to close it. Any objections?

PR #890, opened 2022-03-30
Backfill missing unit tests for ApplyManifest

gcapizzi:
No, although #886 is related.

## What is this change about?
This adds unit tests for all positive cases in `ApplyManifest`. I assume these were missing as we wanted to rely on the integration tests, but #886 shows that it's easy to miss things as simple as swapping the order of two arguments.

No.


kieron-dev (approved):
Looks good. I wonder if there will be other areas where the non-error cases have been skipped in the unit tests.

PR #891, opened 2022-03-30
No longer run codeql on every PR

georgethebeatle:
Now that all checks take ~12m on gh actions it has become the bottleneck
with runtime of ~20m. Maybe runnign it on main only is a good compromise
that will improve our dev turnaround


kieron-dev (approved):
This seems sensible to me. Fast feedback in PRs is the priority. We just have to notice when tests on main fail.

PR #893, opened 2022-03-30
Refactor Validation Errors

Birdrock:
No. Done as support to #852

## What is this change about?
Refactor validation errors. `code` was used to extract canned messages after serialization/deserialization. This isn't as helpful as we'd like, so we replaced this with a `validationErrorType` and made some messaging more dynamic.

Should not contain breaking changes.

## Acceptance Steps
Run all the tests.

@gnovv @matt-royal @tcdowney @gcapizzi

### DIFF
--- a/api/apis/route_handler_test.go
+++ b/api/apis/route_handler_test.go
@@ -815,7 +815,7 @@ var _ = Describe("RouteHandler", func() {
 			})
 
 			It("returns an error", func() {
-				expectUnprocessableEntityError("Invalid Route, FQDN does not comply with RFC 1035 standards")
+				expectUnprocessableEntityError("Invalid Route, {\"validationErrorType\":\"RouteFQDNValidationError\",\"message\":\"FQDN does not comply with RFC 1035 standards\"}")
 			})
 		})
 
--- a/controllers/webhooks/networking/cfroute_validation.go
+++ b/controllers/webhooks/networking/cfroute_validation.go
@@ -11,6 +11,7 @@ import (
 	networkingv1alpha1 "code.cloudfoundry.org/cf-k8s-controllers/controllers/apis/networking/v1alpha1"
 	workloadsv1alpha1 "code.cloudfoundry.org/cf-k8s-controllers/controllers/apis/workloads/v1alpha1"
 	"code.cloudfoundry.org/cf-k8s-controllers/controllers/webhooks"
+
 	"github.com/go-logr/logr"
 	admissionv1 "k8s.io/api/admission/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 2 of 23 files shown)


gnovv (approved):
lgtm

api/apis/route_handler_test.go
```diff
@@ -815,7 +815,7 @@ var _ = Describe("RouteHandler", func() {
 			})
 
 			It("returns an error", func() {
-				expectUnprocessableEntityError("Invalid Route, FQDN does not comply with RFC 1035 standards")
+				expectUnprocessableEntityError("Invalid Route, {\"validationErrorType\":\"RouteFQDNValidationError\",\"message\":\"FQDN does not comply with RFC 1035 standards\"}")
```
matt-royal, line 818: This assertion is brittle, since the JSON field order can change. Better to use a `ContainSubstring` matcher, or somehow json parse the message and assert against the parsed result. Maybe we need a specific matcher for these kind of validation messages?

controllers/webhooks/networking/cfroute_validation.go
```diff
@@ -25,6 +26,19 @@ import (
 const (
 	RouteEntityType = "route"
 
+	RouteDecodingError                     = "RouteDecodingError"
```
matt-royal, line 29: I'd add a `Type` prefix to these constants, since they're not actually errors but the type on them. This also works well with the Message suffix you've given other constants.

controllers/webhooks/workloads/integration/cfapp_validation_integration_test.go
```diff
@@ -93,7 +94,7 @@ var _ = Describe("CFAppValidatingWebhook", func() {
 			})
 
 			It("should fail", func() {
-				Expect(createErr).To(MatchError(ContainSubstring("CFApp with the same spec.name exists")))
+				Expect(createErr).To(MatchError(ContainSubstring(fmt.Sprintf("App with the name '%s' already exists.", app1Name))))
```
matt-royal, line 97: You can remove the Sprintf call, since the matcher will do that for you. `ContainsSubstring("App with the name '%s' already exists.", app1Name)`

matt-royal (changes requested):
The only strictly necessary change is to the brittle assertion against a JSON-encoded string, but I have a few other suggestions. Otherwise this looks great. Thank you for the hard work on this!

api/apis/route_handler_test.go
matt-royal: Hmm... probably best to change it still, unless you're going to remove it almost immediately. The last thing we need is more flakiness in tests. You can use the `SatisfyAll` matcher to apply a `HavePrefix` matcher and 2 `ContainSubstring` matchers, which will avoid the problem.

Issue #444, opened 2022-01-12
[Feature]: Document the configuration of the CF admin user


gcapizzi:
In order to become the CF admin, a user needs to be bound to the `cf-k8s-controllers-admin` `ClusterRole` in the root `cf` namespace. The binding will then propagate to all orgs and spaces and effectively give the admin user permissions on all orgs and spaces.

Let's document this procedure in our README.

matt-royal:
We were low on stories, so we picked this up and submitted a PR

PR #898, opened 2022-03-31
Document creating root namespace & granting admin access

matt-royal:

#444 

## What is this change about?

See issue

No

## Acceptance Steps

See issue

@akrishna90


gcapizzi:
This is another case of tests not triggering because the PR only contains README changes 😞  how can we fix this?

PR #906, opened 2022-04-02
Remove redundant installation of Service Binding CRD

clintyoshimura:
No

## What is this change about?
The install-dependencies.sh script installs the Service Binding Controller which also includes the Service Binding CRDs. Therefore, installing the Service Binding CRDs separately is no longer necessary. The Service Binding Controller yaml must also be present in the dependencies directory so that the integration tests can use them.

No

## Acceptance Steps
All tests (unit, integration, e2e) should pass.

Test manually on kind:
1. Create an org, space, app, and service
2. Bind service to app
3. Ensure a corresponding servicebinding.io/v1beta1 object is created in Kubernetes


kieron-dev (approved):
Looks good.

Issue #905, opened 2022-04-01
[Bug]: The root-namespace-user role should not propagate to orgs and spaces


kieron-dev:
### What happened?

The code to create the root namespace user role inherits the `propagate.hnc.x-k8s.io/none` annotation from the role that caused it to be created, rather than always setting it to true. This means lots of user rolebindings are created in orgs and spaces for no reason. This also means users will have permissions in orgs in which they should have no permissions, e.g. for listing resources.

### What you expected to happen

The root namespace user role binding should always be created with the propagation annotation `none` set to `true`.

### Acceptance Criteria

GIVEN an org 'org1' containing a space 'space1'
WHEN I assign the org manager role to the user 'foo' in 'org1'
THEN I should not see a root-namespace-role rolebinding in org1 or space1 for the user 'foo'

### How to reproduce it (as minimally and precisely as possible)

Create org1 with space1 inside it.
Assign OrgManager to user foo in org1 (this has propagate set to true in role_mappings_config.yaml).
See that the root-namespace-rolebinding for user foo will exist in the cf namespace (correct), but also in the org1 namespace (incorrect) and the space1 namespace (incorrect)

### Environment

Revision of codebase:
Kubernetes version (use `kubectl version`):
Cloud provider or hardware configuration:
Others:

PR #911, opened 2022-04-04
Make sure `cf_user` bindings are never propagated

gcapizzi:
#905

## What is this change about?
This makes sure that `cf_user` bindings are always annotated with `propagate.hnc.x-k8s.io/none=true`, as we never want them to propagate.

No.

## Acceptance Steps
* Create `org1` with `space1` inside it.
* Assign `OrgManager` to user `foo` in `org1`
* See that the `root-namespace-rolebinding` for user `foo` only exists in the `cf` namespace, but *not* in the `org1` or `space1` namespaces.


kieron-dev (approved):
lgtm

PR #912, opened 2022-04-04
retry getting service account tokens

kieron-dev:
No

## What is this change about?
The new up-front user cert and service account creation scripting for e2es is causing a problem when the service account token isn't ready when we try to fetch it. This adds a retry loop so we will eventually get the token.

No

## Acceptance Steps
E2es should be greener


kieron-dev:
Concourse failure will be fixed by merging #910 and rebasing this.

georgethebeatle (approved):

Issue #303, opened 2021-11-26
[Chore]: Remove `Serial` restrictions from tests that have it


georgethebeatle:
### Background

There is at least one test that is configured to run [serially](https://github.com/cloudfoundry/cf-k8s-controllers/blob/d022c298fc1dc1197efa30393a566012965848ed/api/repositories/app_repository_test.go#L142). It will be great if this could be avoided, because parallel tests are not only faster, but also guarantee that you code (and tests) are more robust.

### Action to take

Remove the serial restrictions and fix the flaky tests

### Impact

We have faster and more robust tests

### Dev Notes

If the reason to flake is the ever changing environment due to parallel test we may try to make relative rather than absolute expectations. For example if we create an app and delete two apps, we can check that the new app is there and the two that we deleted are gone, while ignoring any other apps that we do not know about. If you think about it that is exactly what a human would do when playing with a busy multitenant system.

kieron-dev:
Flakes to address:
- [ ] https://ci.cf-k8s.cf/teams/main/pipelines/pr-e2e/jobs/e2e/builds/40
- [ ] https://github.com/cloudfoundry/cf-k8s-controllers/runs/5814624255?check_suite_focus=true (service bindings: 85)

kieron-dev:
https://github.com/cloudfoundry/cf-k8s-controllers/commit/bcba87283752fe0ca54631d5c947d3a3ab038e67 fixes the problem where one or more service account tokens aren't retrieved

kieron-dev:
It seems user cert creation fails sometimes: https://ci.cf-k8s.cf/teams/main/pipelines/pr-e2e/jobs/e2e/builds/45

I fly intercepted onto the worker and saw the `E2E_USER_PEMS` env var had only 3 certificates in it.

danail-branekov:
We should also look into at the api integration tests - why do they need to [start the test env](https://github.com/cloudfoundry/cf-k8s-controllers/blob/4e26a2af86acac0ab68664a54bf4e6afaa5ad3bc/api/tests/integration/integration_suite_test.go#L48) in `BeforeEach` rather than `BeforeSuite`?

PR #927, opened 2022-04-07
Remove the `Serial` constraint from ListApp repo test

danail-branekov:
https://github.com/cloudfoundry/cf-k8s-controllers/issues/303

## What is this change about?
Get rid of `Serial` in list app repository tests
* The app repo uses a restricted user client to list apps, therefore
  it can only return apps in namespaces the user client is permissioned
  to. Therefore it makes no sense to ensure that there are no apps in
  other spaces (as the repo does not "see" them at all)
* While being here introduced misc improvements:
  * Moved `cfapp2`, `cfApp3`, `space2`, `space3` down to contexts that
    really need them (i.e. they are not created for every test)
  * Extracted some invocations of the function under test in
    `JustBeforeEach` to delete a bunch of code repetition

No

## Acceptance Steps
Green tests

### DIFF
--- a/api/repositories/app_repository_test.go
+++ b/api/repositories/app_repository_test.go
@@ -30,11 +30,11 @@ const (
 
 var _ = Describe("AppRepository", func() {
 	var (
-		testCtx                context.Context
-		appRepo                *AppRepo
-		org                    *v1alpha2.SubnamespaceAnchor
-		space1, space2, space3 *v1alpha2.SubnamespaceAnchor
-		cfApp1, cfApp2, cfApp3 *workloadsv1alpha1.CFApp
+		testCtx context.Context
+		appRepo *AppRepo
+		org     *v1alpha2.SubnamespaceAnchor
+		space   *v1alpha2.SubnamespaceAnchor
+		cfApp   *workloadsv1alpha1.CFApp
 	)
 
 	BeforeEach(func() {
@@ -43,9 +43,9 @@ var _ = Describe("AppRepository", func() {
 		appRepo = NewAppRepo(namespaceRetriever, userClientFactory, nsPerms)
 
 		org = createOrgWithCleanup(testCtx, prefixedGUID("org"))
-		space1 = createSpaceWithCleanup(testCtx, org.Name, prefixedGUID("space1"))
-		space2 = createSpaceWithCleanup(testCtx, org.Name, prefixedGUID("space2"))
-		space3 = createSpaceWithCleanup(testCtx, org.Name, prefixedGUID("space3"))
+		space = createSpaceWithCleanup(testCtx, org.Name, prefixedGUID("space1"))
+
+		cfApp = createApp(space.Name)
 	})
 
 	Describe("GetApp", func() {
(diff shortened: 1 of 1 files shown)


api/repositories/app_repository_test.go
```diff
@@ -183,214 +178,178 @@ var _ = Describe("AppRepository", func() {
...
-						MatchFields(IgnoreExtras, Fields{"GUID": Equal(cfApp2.Name)}),
-					))
+		When("there are apps in non-cf namespaces", func() {
+			var nonCFApp *workloadsv1alpha1.CFApp
 
-					sortedByName := sort.SliceIsSorted(appList, func(i, j int) bool {
-						return appList[i].Name < appList[j].Name
-					})
+			BeforeEach(func() {
+				nonCFNamespace := prefixedGUID("non-cf")
```
georgethebeatle, line 223: The non-cf namespace needs to be cleaned up in an AfterEach

api/repositories/app_repository_test.go
```diff
@@ -135,16 +130,15 @@ var _ = Describe("AppRepository", func() {
...
 
 		BeforeEach(func() {
-			cfApp1 = createApp(space1.Name)
-			querySpaceName = space1.Name
+			querySpaceName = space.Name
 		})
 
 		AfterEach(func() {
-			Expect(k8sClient.Delete(context.Background(), cfApp1)).To(Succeed())
+			Expect(k8sClient.Delete(context.Background(), cfApp)).To(Succeed())
```
georgethebeatle, line 137: Is this AfterEach needed if we are going to delete the space anyway?

api/repositories/app_repository_test.go
```diff
@@ -66,28 +65,28 @@ var _ = Describe("AppRepository", func() {
 
 		When("authorized in the space", func() {
 			BeforeEach(func() {
-				createRoleBinding(testCtx, userName, spaceDeveloperRole.Name, space1.Name)
+				createRoleBinding(testCtx, userName, spaceDeveloperRole.Name, space.Name)
 			})
 
 			AfterEach(func() {
-				Expect(k8sClient.Delete(context.Background(), cfApp1)).To(Succeed())
+				Expect(k8sClient.Delete(context.Background(), cfApp)).To(Succeed())
```
georgethebeatle, line 72: Is this AfterEach needed if we are going to delete the space anyway?

georgethebeatle (approved):
LGTM with some minor comments

georgethebeatle:
Opps, this auto-merged. If you feel like my comments make any sense open a new PR and I will review it, should be a quickie.

api/repositories/app_repository_test.go
danail-branekov: Why? Env tests generally do not need clean up as their environment is thrown away anyway. How about getting rid of all cleanup?

api/repositories/app_repository_test.go
danail-branekov: Deleting spaces in env tests does not delete the resources they contain. However, as per my previous comment, cleanup in env tests is not really needed.

Issue #763, opened 2022-03-04
[Feature]: Operator can create a CFOrg Custom Resource


julian-hj:
### Background

**As a** k8s operator
**I want** to create CFOrgs in my root namespace
**So that** I can make orgs declaratively

The gist of this story is that we will create a new CFOrg CRD and corresponding controller, and copy the current logic for org creation from the API shim down into the reconcile loop for the controller.

The controller will wait for the HNC subnamespace anchor to be "ready" before reporting readiness.

### Acceptance Criteria

**GIVEN** I have cluster admin
**WHEN I** list CRDs 
**THEN I** see that there is a CFOrg CRD

**GIVEN** I have cf-admin role in my root namespace
**WHEN I** kubectl apply a `CFOrg` object in my root namespace as follows:
```
apiVersion: workloads.cloudfoundry.org/v1alpha1
kind: CFOrg
metadata:
  name: <org-guid>
  namespace: cf
spec:
  name: <org-name>
```
**THEN I** see that eventually the status is updated to reflect the guid of the org as follows:
```
status:
  guid: <org-guid>
```
**AND I** see that a corresponding SubnamespaceAnchor has been created in the root namespace, with its name matching the org guid, and `metadata/labels` has a label for the org name as follows:
```
metadata:
  ...
  labels:
    cloudfoundry.org/org-name: <org-name>
```
**AND I** see that the new SubnamespaceAnchor has an owner reference to the CFOrg object
**AND**
**WHEN I** list orgs from the CLI using `cf orgs`
**THEN I** see the org is listed. 

### Dev Notes

**Note** that there is currently logic in the API shim to wait after a subnamespace is "ready", until it sees that HNC has given the current user rights to the new namespace. That code probably will need to remain in the API shim, since the controller reconcile loop doesn't know who the "current user" is, and we don't have a better method for determining namespace readiness.

PR #945, opened 2022-04-08
Create CFOrg CRD/Controller

clintyoshimura:
#763

## What is this change about?
Operator can create a CFOrg Custom Resource

No

## Acceptance Steps
See Issue

@matt-royal @akrishna90 @julian-hj


controllers/controllers/workloads/cforg_controller.go
```diff
@@ -0,0 +1,172 @@
...
+
+	var anchor v1alpha2.SubnamespaceAnchor
+	err = r.Client.Get(ctx, req.NamespacedName, &anchor)
+	if err != nil {
+		if !k8serrors.IsNotFound(err) {
+			r.Log.Error(err, fmt.Sprintf("Error getting SubnamespaceAnchor for CFOrg %s/%s", req.Namespace, req.Name))
+			return ctrl.Result{}, err
+		}
+
+		anchor = v1alpha2.SubnamespaceAnchor{
```
georgethebeatle, line 89: It looks like a good idea to extract the code from this line until the end of the if to a private function, e.g. `createSubnamespaceAnchor`. This would make the code easier to read.

controllers/controllers/workloads/cforg_controller.go
```diff
@@ -0,0 +1,172 @@
...
+		}
+
+		err = r.Client.Create(ctx, &anchor)
+		if err != nil {
+			r.Log.Error(err, fmt.Sprintf("Error creating SubnamespaceAnchor for CFOrg %s/%s", req.Namespace, req.Name))
+			return ctrl.Result{}, err
+		}
+
+		setStatusConditionOnLocalCopy(&cfOrg.Status.Conditions, StatusConditionReady, metav1.ConditionUnknown, StatusConditionReady, "")
+		err = r.Client.Status().Update(ctx, cfOrg)
```
georgethebeatle, line 114: Would using `r.Client.Status().Patch()` reduce potential update conflicts?

controllers/controllers/workloads/cforg_controller.go
```diff
@@ -0,0 +1,172 @@
...
+		if err != nil {
+			r.Log.Error(err, "unable to update CFOrg status")
+			r.Log.Info(fmt.Sprintf("CFOrg status: %+v", cfOrg.Status))
+			return ctrl.Result{}, err
+		}
+	}
+
+	if anchor.Status.State != v1alpha2.Ok {
```
georgethebeatle, line 122: The logic from this line to line 130 is waiting for the org to become ready (the anchor to have ok state and the namespace to appear). We could extract it to a private function e.g. `isOrgReady() boolean`.

controllers/controllers/workloads/cforg_controller_test.go
```diff
@@ -0,0 +1,415 @@
...
+			case *v1.Namespace:
+				namespace.DeepCopyInto(obj)
+				return namespaceError
+			default:
+				panic("TestClient Get provided a weird obj")
+			}
+		}
+
+		fakeClient.CreateStub = func(ctx context.Context, obj client.Object, option ...client.CreateOption) error {
+			// cast obj to find its kind
```
georgethebeatle, line 88: comment is redundant

controllers/controllers/workloads/cforg_controller.go
```diff
@@ -0,0 +1,172 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/hierarchical-namespaces/api/v1alpha2"
+)
+
+// CFOrgReconciler reconciles a CFOrg object
+type CFOrgReconciler struct {
```
georgethebeatle, line 38: It is a good practice to have structs with private fields and initialize them with constructors. Otherwise it is too easy to forget to initialize a particular field, which can be annoying.

controllers/controllers/workloads/cforg_controller_test.go
```diff
@@ -0,0 +1,415 @@
...
+		}
+	})
+
+	JustBeforeEach(func() {
+		reconcileResult, reconcileErr = cfOrgReconciler.Reconcile(ctx, req)
+	})
+
+	When("a CFOrg is created and CFOrgReconciler Reconcile function is called", func() {
+		BeforeEach(func() {
+			cfOrg.DeletionTimestamp = nil
```
georgethebeatle, line 126: Is this necessary? DeletionTimesamp is not set in the parent BeforeEach, so it defaults to nil already

controllers/controllers/workloads/cforg_controller_test.go
```diff
@@ -0,0 +1,415 @@
...
+		})
+
+		When("on the happy path", func() {
```
georgethebeatle, line 129: This `When` does not add any information and can be omitted

controllers/controllers/workloads/cforg_controller_test.go
```diff
@@ -0,0 +1,415 @@
...
+			BeforeEach(func() {
```
georgethebeatle, line 130: This `BeforeEach` is redundant.
- namespaceError is already set to nil in the parent `BeforeEach`
- same applies to cfOrgError
- fakeClient.PatchReturns is set to nil by default in the parent `BeforeEach` as a fresh fake is created there
- same applies to fakeStatusWriter

controllers/controllers/workloads/cforg_controller_test.go
```diff
@@ -0,0 +1,415 @@
...
+				cfOrgError = nil
+				namespaceError = nil
+				fakeClient.PatchReturns(nil)
+				fakeStatusWriter.UpdateReturns(nil)
+			})
+
+			When("a subnamespace anchor does not exists", func() {
+				BeforeEach(func() {
```
danail-branekov, line 138: This is the first `BeforeEach` that really changes topmost `BeforeEach`. This means that the `happy path` and the `a subnamespace anchor does not exist` contexts could be removed thus reducing the nesting level by 2

controllers/controllers/workloads/cforg_controller_test.go
```diff
@@ -0,0 +1,415 @@
...
+					subNamespaceAnchorError = k8serrors.NewNotFound(schema.GroupResource{}, "CFOrg")
+					setSubNamespaceStatus = true
+					subNamespaceAnchor.Status.State = v1alpha2.Missing
+					createError = nil
+				})
+
+				It("creates a subnamespace anchor and returns an empty result and does not return error", func() {
```
danail-branekov, line 145: This is quite a huge `It`, its comments clearly separate logical pieces of the create flow. In order to improve readability, comments could be simply replaced by `It`s, e.g. `It("calls client.Get with correct arguments`, ....)`

controllers/controllers/workloads/cforg_controller_test.go
```diff
@@ -0,0 +1,415 @@
...
+					Expect(fakeStatusWriter.UpdateCallCount()).To(Equal(2))
+					_, updatedCFOrg, _ := fakeStatusWriter.UpdateArgsForCall(1)
+					castCFOrg, ok := updatedCFOrg.(*workloadsv1alpha1.CFOrg)
+					Expect(ok).To(BeTrue(), "Cast to workloadsv1alpha1.CFOrg failed")
+					Expect(meta.IsStatusConditionTrue(castCFOrg.Status.Conditions, StatusConditionReady)).To(BeTrue(), "Status Condition "+StatusConditionReady+" was not True as expected")
+					Expect(castCFOrg.Status.GUID).To(Equal(namespace.Name))
+				})
+			})
+
+			When("a subnamespace anchor exists", func() {
```
danail-branekov, line 201: In order to match @georgethebeatle proposal above (the one with `Describe("Create"...)` this could be `Describe("Update", ....)`.

controllers/controllers/workloads/cforg_controller_test.go
```diff
@@ -0,0 +1,415 @@
...
+					_, updatedCFOrg, _ := fakeStatusWriter.UpdateArgsForCall(0)
+					castCFOrg, ok := updatedCFOrg.(*workloadsv1alpha1.CFOrg)
+					Expect(ok).To(BeTrue(), "Cast to workloadsv1alpha1.CFOrg failed")
+					Expect(meta.IsStatusConditionTrue(castCFOrg.Status.Conditions, StatusConditionReady)).To(BeTrue(), "Status Condition "+StatusConditionReady+" was not True as expected")
+					Expect(castCFOrg.Status.GUID).To(Equal(namespace.Name))
+				})
+			})
+		})
+
+		When("on the unhappy path", func() {
```
danail-branekov, line 244: The `unhappy path` context does not bring any value and could be removed in order to reduce nesting

controllers/controllers/workloads/cforg_controller_test.go
```diff
@@ -0,0 +1,415 @@
...
+			When("fetch CFOrg returns an error", func() {
```
danail-branekov, line 245: On the cases that there are errors, it is essential to check that an error is returned and there were no unnecessary `create/update` calls. I don't think the repeating all the checks is necessary, e.g do you care that get has been invoked (this has been tested already above)?

controllers/controllers/workloads/cforg_controller_test.go
```diff
@@ -0,0 +1,415 @@
...
+			cfOrgError = nil
+			cfOrg.ObjectMeta.DeletionTimestamp = &metav1.Time{
+				Time: time.Now(),
+			}
+		})
+		It("returns an empty result and does not return error", func() {
+			Expect(reconcileResult).To(Equal(ctrl.Result{}))
+			Expect(reconcileErr).NotTo(HaveOccurred())
+
+			// validate call counts
```
danail-branekov, line 408: as there is no specific code that handles delete, I don't think that further verifications after the error check make much sense

controllers/controllers/workloads/integration/cforg_controller_integration_test.go
```diff
@@ -0,0 +1,213 @@
...
+			time.Sleep(time.Millisecond * 100)
+		}
+	}
+
+	// simulateHNC waits for the subnamespaceanchor to appear, and then
+	// creates the namespace, creates the hierarchyconfiguration in the
+	// namespace and sets the status on the subnamespaceanchor to Ok. It
+	// will stop waiting for the subnamespaceanchor to appear if anything
+	// is written to the done channel
+	simulateHNC := func(anchorNamespace string, anchorLabel map[string]string, done chan bool) {
```
danail-branekov, line 64: Running the HNC simulation function in the background is trying to fill the gap between envtests and a real cluster with HNC deployed. A "running HNC" is not needed for all controllers tests, for example it is not needed to verify the attributes of the subnamespace anchor itself.

Tests that check the controller behaviour when the namespace is there could be moved into a nested context (e.g. `When("the underlying namespace appears"....`) that creates the namespace in its `BeforeEach`. This would significantly simplify the code and there would be no need for switches like `doHNCSimulation`, or the `done` channel.

controllers/controllers/workloads/integration/cforg_controller_integration_test.go
```diff
@@ -0,0 +1,213 @@
...
+				Consistently(func() (bool, error) {
+					var latestOrg workloadsv1alpha1.CFOrg
+					err := k8sClient.Get(testCtx, types.NamespacedName{Namespace: rootNamespace.Name, Name: orgGUID}, &latestOrg)
+					return meta.IsStatusConditionPresentAndEqual(latestOrg.Status.Conditions, "Ready", metav1.ConditionUnknown), err
+				}).Should(BeTrue())
+			})
+		})
+	})
+
+	When("the CFOrg is deleted", func() {
```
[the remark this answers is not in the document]
clintyoshimura: Done

controllers/controllers/workloads/integration/cforg_controller_integration_test.go
clintyoshimura: Removed the HNC simulation and manually set the status to OK for the appropriate test cases. This greatly simplified the integration test.

controllers/controllers/workloads/cforg_controller_test.go
clintyoshimura: Done

controllers/controllers/workloads/cforg_controller_test.go
clintyoshimura: Done

controllers/controllers/workloads/cforg_controller_test.go
clintyoshimura: Done

controllers/controllers/workloads/cforg_controller_test.go
```diff
@@ -0,0 +1,415 @@
...
+					subNamespaceAnchorError = nil
+					setSubNamespaceStatus = false
+					subNamespaceAnchor.Status.State = v1alpha2.Ok
+				})
+
+				It("returns an empty result and does not return error", func() {
+					Expect(reconcileResult).To(Equal(ctrl.Result{}))
+					Expect(reconcileErr).NotTo(HaveOccurred())
+
+					// validate the inputs to Get
```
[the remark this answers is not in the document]
clintyoshimura: Done. Since we refactored to return after create, the test changed a bit.

controllers/controllers/workloads/cforg_controller_test.go
```diff
@@ -0,0 +1,415 @@
...
+								Name:       cfOrg.Name,
+								UID:        cfOrg.GetUID(),
+							},
+						},
+					}))
+					Expect(castAnchor.Status).To(Equal(v1alpha2.SubnamespaceAnchorStatus{
+						State: v1alpha2.Ok,
+					}))
+
+					// validate the inputs to Patch
```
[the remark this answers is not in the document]
clintyoshimura: Done

controllers/controllers/workloads/cforg_controller_test.go
clintyoshimura: Done

controllers/controllers/workloads/cforg_controller_test.go
clintyoshimura: Good idea. Done

controllers/controllers/workloads/cforg_controller_test.go
clintyoshimura: Removed

controllers/controllers/workloads/cforg_controller_test.go
clintyoshimura: Done

controllers/controllers/workloads/cforg_controller_test.go
clintyoshimura: Not necessary. Removed

controllers/controllers/workloads/cforg_controller_test.go
```diff
@@ -0,0 +1,415 @@
...
+				Name:      cfOrgGUID,
+			},
+		}
+	})
+
+	JustBeforeEach(func() {
+		reconcileResult, reconcileErr = cfOrgReconciler.Reconcile(ctx, req)
+	})
+
+	When("a CFOrg is created and CFOrgReconciler Reconcile function is called", func() {
```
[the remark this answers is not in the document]
clintyoshimura: Done

controllers/controllers/workloads/cforg_controller_test.go
clintyoshimura: Done

controllers/controllers/workloads/cforg_controller_test.go
```diff
@@ -0,0 +1,415 @@
...
+
+		setSubNamespaceStatus = true
+		cfOrgError = nil
+		subNamespaceAnchorError = nil
+		namespaceError = nil
+		createError = nil
+		reconcileErr = nil
+
+		fakeClient.GetStub = func(_ context.Context, _ types.NamespacedName, obj client.Object) error {
+			// cast obj to find its kind
```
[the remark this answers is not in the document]
clintyoshimura: Done

controllers/controllers/workloads/cforg_controller.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	err = setCascadingDelete(ctx, r.Client, req.Name)
+	if err != nil {
+		r.Log.Error(err, fmt.Sprintf("Error updating HierarchyConfiguration for CFOrg %s/%s", req.Namespace, req.Name))
+		return ctrl.Result{}, err
+	}
+
+	cfOrg.Status.GUID = namespace.Name
+	setStatusConditionOnLocalCopy(&cfOrg.Status.Conditions, StatusConditionReady, metav1.ConditionTrue, StatusConditionReady, "")
```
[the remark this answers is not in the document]
clintyoshimura: We tried to do this, but to make it usable for all types, we'd have to use a generic client.Object and that won't let us access the status. In addition, we also have to pass in the logger and client from the reconciler. The setStatusCOnditionOnLocalCopy is a repeat of the pattern in the other controllers since that seemed like the best compromise in making a shared function that didn't require passing in all the objects from the reconcilers.

controllers/controllers/workloads/cforg_controller.go
clintyoshimura: Good call. Done

controllers/controllers/workloads/cforg_controller.go
clintyoshimura: Do you mean conflicting with the update status called at the end of this function? Actually it will never hit that after creating the subnamespace anchor because it isn't being requeried (And so its Status.State param cannot be OK). So I'm actually going to return and requeue after creating the subnamespace anchor so its clear it should check its status on the next Reconcile loop.

controllers/controllers/workloads/cforg_controller.go
clintyoshimura: Good call. Done

controllers/controllers/workloads/cforg_controller.go
clintyoshimura: Done

controllers/controllers/workloads/integration/cforg_controller_integration_test.go
```diff
@@ -0,0 +1,137 @@
...
+			Expect(anchor.Labels).To(HaveKeyWithValue(workloads.OrgNameLabel, orgName))
+		})
+
+		It("makes the CFOrg the owner of its subnamespace anchor", func() {
+			var anchor hncv1alpha2.SubnamespaceAnchor
+			Eventually(func() error {
+				return k8sClient.Get(testCtx, types.NamespacedName{Namespace: rootNamespace.Name, Name: orgGUID}, &anchor)
+			}).Should(Succeed())
+
+			Expect(anchor.ObjectMeta.OwnerReferences).To(ConsistOf([]metav1.OwnerReference{
```
danail-branekov, line 76: FYI, Ginkgo offers an alternative way (see Ginkgo docs [here](https://onsi.github.io/gomega/#category-3-making-assertions-eminem-the-function-passed-into-codeeventuallycode))to put all the assertions in `Eventually` which from my point of view makes the code slightly more readable. But that is up to you really.

Applies to most of the `Eventually`s in this test.

controllers/controllers/workloads/integration/cforg_controller_integration_test.go
```diff
@@ -0,0 +1,137 @@
...
+					var latestOrg workloadsv1alpha1.CFOrg
+					err := k8sClient.Get(testCtx, types.NamespacedName{Namespace: rootNamespace.Name, Name: orgGUID}, &latestOrg)
+					return latestOrg.Status.GUID, err
+				}, 5*time.Second).Should(Equal(cfOrg.ObjectMeta.Name))
+			})
+		})
+
+		When("the subnamespace anchor never becomes ready", func() {
+			It("leaves the CFOrg 'Ready' condition as 'Unknown'", func() {
+				Eventually(func() (bool, error) {
```
danail-branekov, line 123: The function in `Eventually` and `Consistently` is the same, therefore could be extracted. Up to you really.

danail-branekov (approved):
Thank you, the code now looks much better! Approving with minor nitpicks, up to you whether you decide to fix them or not.

georgethebeatle (approved):

controllers/controllers/workloads/cforg_controller.go
```diff
@@ -0,0 +1,200 @@
...
+		anchor, err = r.createSubnamespaceAnchor(ctx, req, cfOrg)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+
+		// Requeue to verify subnamespaceanchor is ready
+		return ctrl.Result{RequeueAfter: 100 * time.Millisecond}, nil
+	}
+
+	namespace, isOrgReady := r.isOrgReady(ctx, anchor, cfOrg.ObjectMeta.Name)
```
gcapizzi, line 107: This method is weird as it performs two unrelated checks on two unrelated arguments. I would inline the anchor status check to make it more visible, and maybe extract a `getNamespace(context.Context, string) *corev1.Namespace`, which we could check for a `nil` return value. Alternatively we could make it `getNamespace(context.Context, string) (*corev1.Namespace, bool)` and use the typical Golang pattern, similarly to what we're doing now:
```golang
namespace, ok := r.getNamespace(ctx, cfOrg.ObjectMeta.Name)
if !ok {
    // ...
}
```

controllers/controllers/workloads/cforg_controller_test.go
```diff
@@ -0,0 +1,321 @@
+package workloads_test
```
gcapizzi, line 1: Similarly to the conversation we've had about testing repositories, are we sure we want to have both an EnvTest-based integration test and a unit test mocking the client? For repositories we have decided to stick to just integration tests, shouldn't we do the same for controllers?

controllers/controllers/workloads/integration/cforg_controller_integration_test.go
clintyoshimura: Thanks. Didn't know that. Done

controllers/controllers/workloads/cforg_controller.go
clintyoshimura: Done

controllers/controllers/workloads/cforg_controller_test.go
clintyoshimura: For controllers I believe we thought that the unit tests had value because we could test error cases that we couldn't with the integration tests. I'll add this to the team forum this week to discuss further.

Issue #928, opened 2022-04-07
[Bug]: Flaky "Create Org" behavior when the user already has cluster-wide permissions to list apps


julian-hj:
### What happened?

Create org has a race condition when the user has cluster admin rights (or other ClusterRoles conferring "list cfapp" permissions). In some cases it fails with 

```
Creating org o10 as kubernetes-admin...
Org not found. Ensure it exists and you have access to it.
```

### What you expected to happen

No flakes--it should just work.

### Acceptance Criteria

GIVEN a kind cluster 
WHEN I log in as a cluster admin that also has a role binding in the space
THEN I can consistently create new orgs without seeing errors

### How to reproduce it (as minimally and precisely as possible)

- deploy to kind
- add a RoleBinding to the cf-admin role for the cluster-admin user
- login as the cluster-admin user
- `for i in {1..30}; do cf co o$i; done`
- you should see that eventually one of the orgs fails after creation

### Anything else we need to know?

We think this is happening because the code in the API shim that waits for HNC to do the rolebinding propagation isn't actually waiting for a role binding to exist--it is just waiting until the current user has rights to list apps in the new namespace.  This works great for non-privileged users, but doesn't wait if the user is a cluster admin, since that user can do everything.

The call to get the org afterwards *does* look for a role binding, so it will fail unless there is a role binding, regardless of the actual permissions of the user.

We could fix this by using the admin client to wait for a RoleBinding to appear, or we could fix it by checking in the org listing code to see if the user has clusterroles.

### Environment

Revision of codebase:
Kubernetes version (use `kubectl version`):
Cloud provider or hardware configuration:
Others:

PR #953, opened 2022-04-11
Fix flakey behaviour with creating cforgs

akrishna90:

#928 

## What is this change about?
- HNC status "OK" doesn't seem to promise the existence of the namespace. This was causing the check that was waiting for role bindings to propagate to fail with a namespace "NotFound" error.

- Also when the user is a cluster-admin user, he/she would have permissions to list all kinds of objects, hence waiting to see if a user can list apps is not a strong check.

- We have switched to using the privileged client to list the role-bindings in the newly created namespace as a way to say that permissions have propagated

No

## Acceptance Steps

Creating `CFOrgs` should not be falkey.

@gnovv


gnovv (approved):
pairing w/ @akrishna90 lgtm!

PR #955, opened 2022-04-12
Fix flakey behaviour with creating cfspace

akrishna90:
No, but related to #928

## What is this change about?
- We ran into an issues with creating cf-spaces where the creation would fail with cf-org not found error.

- Having a check to see if the user is able to list apps after creation of orgs and spaces would not work the way we expect if the user is a k8s "cluster-admin". The cluster-admins have permissions to list any object.

- Added a better check to make sure that the role binding exists for the current user in the newly created namespace before returning success on create orgs and spaces

No

## Acceptance Steps

- `make test-e2e` should pass
- smoke tests should pass consistently

@gnovv @clintyoshimura


clintyoshimura (approved):
LGTM

PR #961, opened 2022-04-13
Fix the scripts to work on Macs

matt-royal:
 No

## What is this change about?

After recent changes to the scripts, they no longer run successfully on Mac computers due to different flags for `base64`. This PR accomplishes the same output in a way that works on both Mac and Linux.

No

## Acceptance Steps
Run the `./scripts/deploy-on-kind.sh` script locally and see that it succeeds


acosta11 (approved):
LGTM

PR #973, opened 2022-04-18
Fix e2e ServiceBinding List Flake

gnovv:

No 🤠 
## What is this change about?

tries to fix an RBAC flake due to test sequencing granting permissions to ServiceBindings before the empty list test runs.
The goal of the test is more RBAC focused rather than testing the empty list case anyway. Now it just confirms that it doesn't see ServiceBindings that we know aren't permitted by the RBAC.

no.

## Acceptance Steps

confirm boxes green.

@acosta11


acosta11 (approved):
LGTM

PR #975, opened 2022-04-19
Draft proposal: Support Interchangeable Build and Run Implementations

matt-royal:
[Readable Link](https://github.com/cloudfoundry/korifi/blob/proposal/interchangeable-build-and-run-implementations/docs/proposals/0000-support-interchangeable-build-and-run-implementations.md)

Some of the details need to be fleshed out, but I want to share this proposal so we can all make it better together.


docs/proposals/0000-support-interchangeable-build-and-run-implementations.md
```diff
@@ -0,0 +1,248 @@
...
+
+- Using `Deployment`, knative, or other alternative resources for run workloads
+- Signing container images in the build step, and requiring signed images in the run step
+- Security scanning container images during the build and/or run steps
+- Applying conventions to pod spec, such as labels, annotations, volume mounts, etc.
+
+We could accomplish this flexibility with a smaller change: splitting out the current reconcilers for `CFBuild` and
+`CFProcess` into their own repositories, so they can be replaced with alternatives. The issue with this change is that
+the existing reconcilers use information from many resources and an alternative implementation would have to duplicate
+this complex logic. For example, to create the LRPs that currently generate the run step's stateful set, the `CFProcess`
```
tcdowney, line 36: I think it's worth mentioning that duplicating this logic is not a one time cost either. Any new features we add or bugs we fix in our logic here will need to be carried forward to all implementors. I think the dedicated resource as an interface will make both our own lives and those of anyone extending this easier.

docs/proposals/0000-support-interchangeable-build-and-run-implementations.md
```diff
@@ -0,0 +1,248 @@
...
+      memory: 1024M
+    requests:
+      cpu: '0'
+      memory: 1024M
+      
+  envFrom:
+    - secretRef:
+        name: user-specified-env-vars
+        
+  env:
```
tcdowney, line 153: This looks like the K8s `EnvVar` type which is nice because it's something we want to add to the Eirini LRP anyway.

* https://github.com/cloudfoundry/korifi/issues/918
* https://github.com/cloudfoundry/korifi/issues/917

docs/proposals/0000-support-interchangeable-build-and-run-implementations.md
```diff
@@ -0,0 +1,248 @@
...
+  
+  version: '0'
+```
+
+
+
+# Migration
+[migration]: #migration
+
+This will be a breaking change, but we haven't made any guarantee about upgrades. Existing installations will need to be
```
tcdowney, line 202: I think we also want to rename the CRDs (take off the CF prefix) so it would be nice to coordinate this type of change with that.

docs/proposals/0000-support-interchangeable-build-and-run-implementations.md
```diff
@@ -0,0 +1,248 @@
...
+[what-it-is]: #what-it-is
+
+Here are the various resources and components involved:
+
+- `BuildWorkload` - a new Custom Resource Definition, with fields that contain all necessary information to execute the
+  build step for a korifi workload. This resource contains information from `CFBuild`, `CFPackage`, `CFServiceBinding`,
+  and the App's `Secret` resources. This CRD may exist in a new, separate API Group to indicate its status as an
+  external interface.
+
+- `RunWorkload` - a new Custom Resource Definition, with fields that contain all necessary information to execute the
```
tcdowney, line 54: I think @gcapizzi is also thinking through something to model CF Tasks. I'm wondering could be a separate case of this one or a dedicated `TaskWorkload` resource.

docs/proposals/0000-support-interchangeable-build-and-run-implementations.md
```diff
@@ -0,0 +1,248 @@
...
+# Prior Art
+[prior-art]: #prior-art
+
+N/A
+
+# Unresolved Questions
+[unresolved-questions]: #unresolved-questions
+
+- Should we place these new resources in a new API group, and what should its name be?
+- Are `BuildWorkload` and `RunWorkload` good names? Do we want to add the `CF` prefix that all other resources currently
```
tcdowney, line 241: In a past call we actually are thinking of removing the `CF` prefix from the other resources and adding convenience shortnames (e.g. `cfbuild`, `cfapp`, etc.) that we can use with `kubectl`. I can't think of anything better necessarily right now.

docs/proposals/0000-support-interchangeable-build-and-run-implementations.md
matt-royal: That's exactly what it is. I was thinking that we could just accept `env` and `envFrom` like the pod spec does, rather than having our own different representation for this information. In general, I'm thinking that we have something pod-spec-like, using https://github.com/knative/specs/blob/main/specs/serving/knative-api-specification-1.0.md#revisiontemplatespec as an example.

docs/proposals/0000-support-interchangeable-build-and-run-implementations.md
matt-royal: I think it would be clearer if we had a dedicated `TaskWorkload` for that, since it would be reconciled differently.

tcdowney: Yeah, and that would then cover the three types of Diego tasks (staging, LRP, and task) Cloud Controller makes today so it's unlikely we'd need even more.

docs/proposals/0000-support-interchangeable-build-and-run-implementations.md
```diff
@@ -0,0 +1,248 @@
...
+
+- build workload reconciler - a separate component from `korifi-controllers` that reconciles `BuildWorkload` resources
+  to execute the build workflow. Eventually multiple implementations of this reconciler will exist, each implementing
+  the build workflow in a different way to meet operator needs. These CRDs may exist in a new, separate API Group to
+  indicate their status as an external interface. We may even want to reclaim the `workloads.cloudfoundry.org` group for
+  this purpose and choose another for the existing resources.
+
+- run workload reconciler - similar to the build workload reconciler, but for `RunWorkload` resources.
+
+- the `CFBuild` reconciler: a component of `korifi-controllers` that creates `BuildWorkload` resources and monitors
```
tcdowney, line 69: Here's an idea -- I'm not sure where the best place to put this is, but will just drop it here.

What if we had forego some of the V3 CF behavior around having multiple concurrent builds and instead behaved more like V2 CF and only allowed a single build to be in flight (I think in V2 subsequent builds would either fail or supersede the ongoing one). This would allow us to have a single `BuildWorkload` per app which might simplify things and the `CFBuild` controller could just update this with the new package info, etc.

This would have some benefits:
1. In our current implementation there would be a single `kpack Image` per app, so we could leverage some of the caching behavior it has. We just push toy apps right now like the node app or staticfile app, but real world users with large Java apps (or node apps with a lot of dependencies) would appreciate this.
2. I think this would more naturally support the auto-rebuild functionality of kpack when a stack gets updated. One of the things we're missing in Korifi is automatic rollout of stack updates (like we have in CF for VMs when Diego cells get rolled and apps evacuate) and this would simplify our design and support of this.
3. I think it might also potentially make it easier to extend for implementing some of the alternative build systems I can think of.

kieron-dev:
This all makes a lot of sense to me.

I'll be interested in seeing how we structure the default implementations of reconcilers for the new resources. E.g. for runtime, would we be bringing eirini functionality into the project, reconciling from the RunWorkload instead of LRP. If so, would it run in a separate deployment, or be feature-flagged into the existing controllers deployment? Or is it better to keep the default implementation in a separate project for easy swappability?

georgethebeatle:
Aren't proposals meant to be google docs linked on the board in the [proposals project](https://github.com/orgs/cloudfoundry/projects/14)? I am not against using Draft PRs for that, but we need to be consistent about it I guess.

matt-royal:
Good point, @georgethebeatle . It had been a while and I'd forgotten about our existing process, and defaulted to how Cartographer does their RFCs. I just moved the contents (and a copy of the comments) to a google doc: https://docs.google.com/document/d/1SqCnM2kHdALoOHr5GDm8K4IKbz4E7CYNEo1f8ETTsMU/edit#

matt-royal:
Closing this, since I've moved the proposal over to a Google Doc to respect our existing process

PR #980, opened 2022-04-20
Switch OrgRepository to use CFOrg resources

### LINKED ISSUE #765: [Feature]: cf user can create/list/delete CFOrg resources through the CF API
### Blockers/Dependencies

#763 

### Background

**As a** cf cli user
**I want** to create orgs backed by CFOrg objects
**So that** I can live in harmony with the Future

In this story, we will replace the logic in the API Shim that manages subnamespaceanchors with logic that instead manages CFOrg custom resources.

Note that access control for CFOrgs will differ slightly from the current model. Root namespace users interacting through kubectl will be able to list all CFOrg objects in the root namespace. When listed through the API shim, we will continue to filter the org list to show the user only the orgs the user has a role in. This is for convenience rather than security. 

### Acceptance Criteria

**GIVEN** I am a cf-admin
**WHEN I** invoke `cf create-org o` with the CF cli
**THEN I** can observe that a new CFOrg CR has been created

**GIVEN** I am a cf-admin
**WHEN I** invoke `cf orgs` with the CF cli
**THEN I** see a list of all CFOrgs in the root namespace

**GIVEN** I am a space developer
**WHEN I** invoke `cf orgs` with the CF cli
**THEN I** see a list of only the orgs in which I have membership

**GIVEN** I am a cf-admin
**WHEN I** invoke `cf delete-org o` with the CF cli
**THEN I** see that the CFOrg and corresponding SubnamespaceAnchor and namespaces eventually disappear.

**WHEN** I look at the code
**THEN** I see that the Org repository code is no longer using a privileged client to create an org.

### Dev Notes

We have logic in the API to wait until the current user has access to the new namespace (with HNC rolebinding propagation). We need to leave that logic in place or lots of tests will start to flake.

We might break name uniqueness in this story--if so, we can make a different story to fix it.

matt-royal:

Closes #765

## What is this change about?

See issue

## Acceptance Steps

See issue

@gnovv 

## Things to remember

We know that we broke at least one e2e test, due to the missing uniqueness validation on CFOrg.spec.name. We either need to pend this test or block this PR on the work to add that validation.

### DIFF
--- a/api/config/base/rbac/role.yaml
+++ b/api/config/base/rbac/role.yaml
@@ -74,20 +74,10 @@ rules:
   resources:
   - cfapps
   - cfbuilds
+  - cfpackages
+  - cfprocesses
   verbs:
-  - create
-  - delete
-  - get
   - list
-  - patch
-  - update
-  - watch
-- apiGroups:
-  - workloads.cloudfoundry.org
-  resources:
-  - cfbuilds/status
-  verbs:
-  - get
 - apiGroups:
   - workloads.cloudfoundry.org
   resources:
@@ -102,10 +92,3 @@ rules:
   - cforgs/status
   verbs:
   - get
-- apiGroups:
-  - workloads.cloudfoundry.org
-  resources:
-  - cfpackages
-  - cfprocesses
-  verbs:
-  - list
--- a/api/config/base/rbac/role.yaml
+++ b/api/config/base/rbac/role.yaml
@@ -78,17 +78,3 @@ rules:
   - cfprocesses
   verbs:
   - list
-- apiGroups:
-  - workloads.cloudfoundry.org
-  resources:
-  - cforgs
-  verbs:
-  - get
-  - list
-  - watch
-- apiGroups:
-  - workloads.cloudfoundry.org
-  resources:
-  - cforgs/status
-  verbs:
-  - get
--- a/api/config/base/rbac/role.yaml
+++ b/api/config/base/rbac/role.yaml
@@ -74,6 +74,37 @@ rules:
   resources:
   - cfapps
   - cfbuilds
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
+  - workloads.cloudfoundry.org
+  resources:
+  - cfbuilds/status
+  verbs:
+  - get
+- apiGroups:
+  - workloads.cloudfoundry.org
+  resources:
+  - cforgs
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - workloads.cloudfoundry.org
+  resources:
+  - cforgs/status
+  verbs:
+  - get
+- apiGroups:
+  - workloads.cloudfoundry.org
+  resources:
   - cfpackages
   - cfprocesses
   verbs:
--- a/api/repositories/org_repository.go
+++ b/api/repositories/org_repository.go
@@ -18,9 +18,6 @@ import (
 	"sigs.k8s.io/hierarchical-namespaces/api/v1alpha2"
 )
 
-//+kubebuilder:rbac:groups=workloads.cloudfoundry.org,resources=cforgs,verbs=get;list;watch
-//+kubebuilder:rbac:groups=workloads.cloudfoundry.org,resources=cforgs/status,verbs=get
-
 //+kubebuilder:rbac:groups=hnc.x-k8s.io,resources=subnamespaceanchors,verbs=list;watch
 //+kubebuilder:rbac:groups=hnc.x-k8s.io,resources=hierarchyconfigurations,verbs=get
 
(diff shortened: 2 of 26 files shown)


gnovv (approved):
Pairing w/ @matt-royal  LGTM!

api/tests/e2e/orgs_test.go
```diff
@@ -219,25 +228,12 @@ var _ = Describe("Orgs", func() {
 				orgGUID = "nope"
 			})
 
-			It("returns a not found error", func() {
-				expectNotFoundError(resp, errResp, "Org")
-			})
-
 			AfterEach(func() {
-				orgGUID = originalGUID
+				deleteOrg(originalGUID)
```
kieron-dev, line 232: I guess we don't need this if the delete is deferred in the create?

kieron-dev (commented):
Looks fine to me. I added a couple of minor comments in the diff.

It seems the e2es are failing though, so they'll need fixing before merging this.

api/repositories/org_repository.go
```diff
@@ -125,61 +129,105 @@ func (r *OrgRepo) CreateOrg(ctx context.Context, info authorization.Info, org Cr
...
-		OrgResourceType,
-	)
+		Spec: workloads.CFOrgSpec{
+			Name: org.Name,
+		},
+	})
+
 	if err != nil {
-		if webhookError, ok := webhooks.WebhookErrorToValidationError(err); ok {
+		if webhookError, ok := webhooks.WebhookErrorToValidationError(err); ok { // untested
```
danail-branekov, line 144: I have always wondered what is the purpose of those `//untested` comments. Isn't codeclimate's report better to track such sort of things? Or, if you think that testing this branch is important but is hard, maybe filing a chore about it would speed up fixing that.

api/repositories/org_repository.go
```diff
@@ -125,61 +129,105 @@ func (r *OrgRepo) CreateOrg(ctx context.Context, info authorization.Info, org Cr
...
+			conditionReady = true
+			break
+		}
+	}
+
+	if !conditionReady {
+		return nil, fmt.Errorf("cf org did not get Condition `Ready`: 'True' within timeout period %d ms", r.timeout.Milliseconds())
+	}
+
+	// wait for the namespace to be created and user to have permissions
```
danail-branekov, line 204: Now that we have our custom orgs resource and custom reconciler, can we not update the org status to `ready` only when the underlying namespace is there and get rid of this code from the repo?

api/tests/e2e/orgs_test.go
```diff
@@ -211,6 +207,19 @@ var _ = Describe("Orgs", func() {
...
 
+		When("the org contains a space", func() {
+			BeforeEach(func() {
+				createSpace(generateGUID("some-space"), orgGUID)
+			})
+
+			It("can still delete the org", func() {
+				Expect(resp).To(SatisfyAll(
+					HaveRestyStatusCode(http.StatusAccepted),
+					HaveRestyHeaderWithValue("Location", HaveSuffix("/v3/jobs/org.delete-"+orgGUID)),
```
danail-branekov, line 218: For the sake of proper E2E test, we should expect that eventually the URL specified by the `Location` header says that the job has completed successfully

api/repositories/org_repository.go
matt-royal: I usually add these comments to make it clear that we chose not to test that line, rather than it being an oversight. If it's behavior that I'm concerned about testing, then I usually try to find a way to test it instead of leaving the comment.

api/repositories/org_repository.go
matt-royal: For now we believe we still need the namespace access check. The CFOrg shouldn't become Ready until the namespace exists, but we don't have a good way to check if HNC has propagated the rolebindings into the new namespace. When we replace HNC, the new solution should allow us to track rolebinding propagation easily. We'll be able to remove this then.

api/tests/e2e/orgs_test.go
matt-royal: Good idea. We moved this test up but didn't change it in this PR, but we can take a stab at testing for job success.

api/tests/e2e/orgs_test.go
matt-royal: Are you thinking of the `createOrgWithCleanup` helper that we use in the repository tests? I don't think there's any deferred cleanup here.

kieron-dev: I was confused by the `createOrgWithCleanup`. I must pay more attention to the filenames!

api/repositories/org_repository.go
danail-branekov: Yes, agree, my point was about moving those checks to the cforg reconciler that sets the ready flag whenever the very same conditions are met. Then the repo should only check the ready org flag, rather than checking the namespace as well. This would be beneficial to theoretical  kubectl users as well.

But I can also see your point that this is probably a temporary workaround..

api/config/base/rbac/role.yaml
```diff
@@ -203,6 +203,20 @@ rules:
...
+  resources:
+  - cforgs
+  verbs:
+  - get
+  - list
+  - watch
+- apiGroups:
+  - workloads.cloudfoundry.org
+  resources:
+  - cforgs/status
```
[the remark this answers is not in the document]
gnovv: It wasn't needed- you were right!
Thank you!

PR #985, opened 2022-04-21
Trim permissions of the api shim service account

danail-branekov:
https://github.com/cloudfoundry/korifi/issues/469

## What is this change about?
Restrict the permissions of the API service account down to minimum via removing/adjusting kubebuilder annotations to reflect what is actually needed.

We also needed to fix the role repository to remove the privileged
client. Previously it required that for a namespace lookup, but that can
be handled properly by the org repo, so we injected a dependency for
that.

No

## Acceptance Steps
Successful tests

@kieron-dev @georgethebeatle


kieron-dev (approved):
lgtm

Issue #469, opened 2022-01-14
[Feature]: Review permissions for `cf-admin-serviceaccount` and rename it accordingly


gcapizzi:
The API shim uses two Kubernetes clients:
* a user-scoped client, built from the authentication header sent by the user in the request;
* a privileged client.

Most endpoints currently use the user-scoped clients, but we still have some legitimate uses for the privileged client. We want to reduce the permissions of the service account that backs it to the minimum necessary to function. This is currently [not the case](https://github.com/cloudfoundry/cf-k8s-controllers/blob/c92e095997059946b3c35b267afef5d51ff4ed09/api/config/base/rbac/role.yaml), as the privileged client used to be everywhere and we haven't trimmed down its permissions as we removed it from the codebase.

The name of the service account and its role (`cf-admin-serviceaccount`/`clusterrole`/`clusterrolebinding`) is also misleading, as this has nothing to do with the [`admin` user](https://github.com/cloudfoundry/cf-k8s-controllers/blob/c92e095997059946b3c35b267afef5d51ff4ed09/controllers/config/cf_roles/cf_admin.yaml). Something like `cf-serviceaccount`/`clusterrole`/`clusterrolebinding` makes more sense.

kieron-dev:
We've implemented the minimal permissions for the service account.

It remains to rename it to something more appropriate

georgethebeatle:
Whoops, I missed that and cut the draft release @kieron-dev . I guess the service account name is not overly important, no big deal if stays the same, given that the permissions are settled. We could also re-start the version counter and cut a new one, since this one is only an unpublished draft.

gcapizzi:
I think it'd be ideal to have the rename in the draft release, but only if it's not too much work! We'd also have to delete the existing tag and release before we kick off the pipeline right? I'm afraid that we might forget something and end up in an inconsistent state, but if you think it's doable I'd do it.

PR #988, opened 2022-04-25
Improve service account / role naming

kieron-dev:
#469 

## What is this change about?
The name `cf-admin` should be reserved for the CloudFoundry Admin user [role]. So let's not use that in the API service account name, the cluster role it uses and the cluster role binding. The name `system` should be clear after kustomize has added its `korifi-api-` prefix.

No

## Acceptance Steps
`make test-e2e`


julian-hj (approved):
👏

PR #1003, opened 2022-04-29
Suppress acsii art from openssl

julian-hj:

No.

## What is this change about?
- openssl creates an ascii representation of a key every time we
  generate one, which results in a big glob of useless dots and plusses
  in our script output.
- this change redirects that output to /dev/null

No.

## Acceptance Steps
- run ./scripts/deploy-on-kind.sh or run the e2e tests
- observe that the output is 15% less annoying

Anyone? Bueller? Bueller?


clintyoshimura (approved):
LGTM

Issue #656, opened 2022-02-11
[Chore]: Introduce Correlation-IDs and use in logs


kieron-dev:
### Background

When debugging an issue, it can be hard to find the relevant information in the api logs. If e2e tests were running, for instance, we might see lots of failure lines, but not be able to work out if any are connected to the issue we're investigating.

It would be better if all log lines for a particular request contained a unique id. A common pattern for this is to use a correlation id. A request provides an id in the header field "X-Correlation-ID", and each logger should take care to set this value so it appears in its output.

We should also pass the id in a correlation id header when making calls to other systems, which helps with distributed logging and metrics.

We should integrate this approach into our e2e tests so that the fail handler is aware of the correlation id and so can filter the log lines shown to those relevant to the failing test.

#### Implement a new correlation ID middleware
This should extract the correlation ID, if any, from the request, and store it in the request context so that it is available to all code receiving the context.

It might be useful to generate a correlation ID if none is passed in. This means we can still tie together log lines from a particular request. In this case it might be useful to set "X-Correlation-ID" in the response headers.

#### Use the correlation id in all loggers
Each method or function performing logging and receiving a context should set the `correlation-id` value on the logger.

#### Teach the e2e test suite about correlation id
Each REST call from the suite should generate a correlation id and pass it in the `X-Correlation-ID` header. The fail handler should filter the logs it prints using the correlation id from the failing test.

### Impact

_Immediate_: API logs are more useful.
_Longer term_: We can contribute to distributed logging, profiling and metrics.

tcdowney:
cc @gnovv who is interested in this area (OpenTelemetry, distributed tracing, etc.)

Also this might cross the line into feature work. Feels like it's similar in spirit to `X-Vcap-Request-Id` and a key part of the platform operator debugging experience. cc @emalm

PR #1024, opened 2022-05-04
Introduce correlation IDs to requests and logs

kieron-dev:
#656 

## What is this change about?
If X-Correlation-ID is passed in the request headers, this will be used as an ID for the life of the request. If not passed, one will be generated. The value is returned in the response headers so a generated value can be parsed by the calling application if required.

All logs will include the id in the `correlation-id` value.

Removing the development: true flag to the logger, as it's harder to read with all the multi-line stacktraces in the way, and harded to link to the correlation ID.

No

## Acceptance Steps
API logs should all have correlation-id set in them.


api/actions/read_app_logs.go
```diff
@@ -35,21 +36,23 @@ func (a *ReadAppLogs) Invoke(ctx context.Context, authInfo authorization.Info, a
 		defaultLogLimit = 100
 	)
 
+	logger := apis.AddCorrelationIDToLogger(ctx, a.logger)
```
georgethebeatle, line 39: All loggers have to be initialized in the new way now, which is repetitive and can be forgotten. Could we create the logger in the auth-aware handler wrapper and pass it to the handler explicitly as an argument?

Maybe it would make sense to propagate the logger with the correlation id from the handler to all places that do logging, e.g. the pod repository, the retrying k8s client, etc. What do you think?

georgethebeatle: We could also merge this as is and have a chore to do what we suggested above. Let's discuss at conv street tomorrow.

danail-branekov (approved):

PR #1061, opened 2022-05-09
Fix unsafe references to local for loop variables

julian-hj:

No.

## What is this change about?
- for loop uses a local copy when looping. This shouldn't be
  dereferenced, as the for loop will overwrite it and leave an
  unreliable pointer behind.
- also, add underscore receivers for error returns we don't care about
  in order to make golint/gosec happy.

No.

@akrishna90


akrishna90 (approved):
paired with @julian-hj , lgtm!

georgethebeatle (approved):

PR #1071, opened 2022-05-11
Include spec.displayName as an additionalPrinterColumn

tcdowney:
`kubectl` will now print the display name and age of the following resources:
  - CFOrg
  - CFSpace
  - CFApp
  - CFServiceInstance
  - CFServieBinding

Issue: https://github.com/cloudfoundry/korifi/issues/1069

## What is this change about?
Improve the UX of kubectl users by printing the display name of a resource when its available.

No

## Acceptance Steps
Create an org, space, push an app, create a user-provided service instance, and then bind the instance. This will create all the affected custom resources on the cluster. Then view them all with `kubectl`.

These were the specific CF commands I ran:

1. `cf create-org o && sleep 1 && cf create-space s -o o && sleep 1 && cf target -o o -s s`
2. `cf push node`
3. `cf cups my-upsi`
4. `cf bind-service node my-upsi --binding-name my-binding` (if you don't specify a binding-name then the display name for that resource is blank)

Attached is a screenshot of how it looks in my env.

Screen Shot 2022-05-11 at 11 47 08 AM

I soloed on this, just tagging a couple folks from the team to review: @matt-royal @clintyoshimura

### DIFF
--- a/controllers/apis/services/v1alpha1/cfservicebinding_types.go
+++ b/controllers/apis/services/v1alpha1/cfservicebinding_types.go
@@ -48,6 +48,8 @@ type CFServiceBindingStatus struct {
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
+//+kubebuilder:printcolumn:name="Display Name",type=string,JSONPath=`.spec.displayName`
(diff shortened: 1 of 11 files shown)


Birdrock (approved):
/lgtm

PR #1073, opened 2022-05-11
ProcessController Eirini LRP calculates CPUWeight

### LINKED ISSUE #586: [Feature]: CPU Weight for Apps should be something higher than 0
### Blockers/Dependencies

_No response_

### Background

Currently we're setting the Eirini LRP `CPUWeight` property to 0 [here](https://github.com/cloudfoundry/cf-k8s-controllers/blob/b9e77468987cb7f9843c0dc0f36c0c8d9503baeb/controllers/controllers/workloads/cfprocess_controller.go#L256). This is eventually converted to a Kubernetes CPU request by [eirini-controller](https://github.com/cloudfoundry-incubator/eirini-controller/blob/60b8ffd2555ec4fdf6299bcf216b8b82565c7f32/k8s/stset/lrp_to_statefulset.go#L259) when it creates the App's `StatefulSet`.

**As an** App Developer
**I want** my Apps to have some amount of guaranteed CPU
**So that** they do not get completely starved for CPU resources

### Acceptance Criteria

**GIVEN** I have `cf push`ed an App
**WHEN I** inspect the `Pods` that are actually running the App
**THEN I** reasonable CPU Requests (and Limits?) set

---

### Dev Notes

* CF for VMs scales CPU based on the amount of memory that has been allocated to a Process: https://github.com/cloudfoundry/cloud_controller_ng/blob/111c846ceba16771bff030e583449977a593a344/lib/cloud_controller/diego/task_cpu_weight_calculator.rb

gnovv:

closes #586
## What is this change about?

Updates Process Controller to calculate CPUWeight based on Memory.
Uses [this calculation](https://github.com/cloudfoundry/cloud_controller_ng/blob/111c846ceba16771bff030e583449977a593a344/lib/cloud_controller/diego/task_cpu_weight_calculator.rb)

No.

## Acceptance Steps

See issue.

@akrishna90


gnovv:
@kieron-dev @gcapizzi can you please review this PR if you get a chance, right now the max CPU allocation the StatefulSet can end up with is `100m` which feels a little low.

kieron-dev (approved):
Like Giuseppe says [here](https://cloudfoundry.slack.com/archives/C0297673ASK/p1652349791394499?thread_ts=1652295528.620179&cid=C0297673ASK), that value only ends up in the CPU request, not the limit. The limit enforces a strict limit on CPU usage (setting `cpu.cfs_quota_us` in the cgroup), whereas the request is used for scheduling and relative CPU priority (setting `cpu.shares` in the cgroup).

So if the value is kept low, apps are not restricted for scheduling by their requested CPU values, and instead are limited by memory, which is the traditional CF approach. And since we don't set the CPU limit, the apps compete with each for CPU relative to the value set in the CPU request.

So I think we're good. Until k8s changes how it applies these values to cgroups.

gcapizzi:
This could be a problem when running non-CF apps on the same cluster though - and not an easy one to solve, given CF doesn't support CPU requests 😕

Issue #406, opened 2022-01-06
[Feature]: App Developer can use `-m` flag to set explicit memory limit for the app during `cf push`


akrishna90:
### Blockers/Dependencies

NA

### Background

While we have increased the default app memory allocation to 1G to accommodate defaults for Java apps better, we do still want to make the quantity easily configurable per app.

**As an** App Developer
**I want** `cf push my-app-name -m 1024MB`
**So that** to be able to configure memory limits for my App on the manifest

The CLI generates its own manifest payload from the cf push arguments/options and from its own internal defaults if it doesn't have an explicit manifest file

```
applications:
- name: my-app
  memory: 1024MB
```

### Acceptance Criteria

**GIVEN** I have an app
**WHEN I** `cf push` the app with `-m` specified
**THEN I**  see my workloads are started with the specified memory limits. 

### Dev Notes
- Current support for configuring app memory 
    - setting defaults in configuration file [Korifi_api_config.yaml](https://github.com/cloudfoundry/korifi/blob/main/api/config/base/apiconfig/korifi_api_config.yaml). 
     - setting them per process under processes array in the manifest
    ```
    ---
    version: 1
    applications:
    - name: app1
    processes:
      - type: web
        command: start-web.sh
        disk_quota: 512M
        instances: 3
        memory: 500M
      - type: worker
        command: start-worker.sh
        disk_quota: 1G
        instances: 2
        memory: 256M
    ```

akrishna90:
Capturing CF CLI behaviour with `-m` flag.
---
1. No memory specified

manifest.yaml
```
applications:
- name: node2
```

CF CLI - `cf push node2 -m 512M`

 updated manifest.yaml from logs
```
Updating with these attributes...
  applications:
  - name: node2
    memory: 512M
    default-route: true
```

---
2. When manifest file has `processes - web - memory` specified 

manifest.yaml
```
applications:
- name: node2
  processes:
  - type: web
    memory: 1024M
```
CF CLI - `cf push node2 -m 512M`

updated manifest.yaml from logs
```
  applications:
  - name: node2
    processes:
    - type: web
      memory: 512M
    default-route: true
```

--- 
3. When manifest file has `app - memory` and `processes - web -memory` specified 

manifest.yaml
```
applications:
- name: node2
  memory: 1024M
  processes:
  - type: web
    memory: 1024M
```

CF CLI - `cf push node2 -m 512M`

updated manifest.yaml
```
  applications:
 - name: node2
[...]

PR #1079, opened 2022-05-12
Add support for "app memory" manifest field

davewalter:
#406

## What is this change about?
This PR adds support for the memory field in an application manifest, which will create a "web" process if one does not already exist with the specific memory request.

No

## Acceptance Steps
See #406

@akrishna90


akrishna90 (approved):
paired with @davewalter, lgtm!

PR #1105, opened 2022-05-18
Fix kpack-builder Kustomization

julian-hj:
No

## What is this change about?
- the Kustomize command we were using in the makefile didn't target the
  image we want to replace when we build the controller.

no?

## Acceptance Steps
run deploy-on-kind.sh

@clintyoshimura


clintyoshimura (approved):
LGTM

Issue #1029, opened 2022-05-05
[Feature]: builds created through the shim generate BuildWorkload resources


matt-royal:
### Blockers/Dependencies

[#1002]

### Background

**As a** developer who wants to extend korifi
**I want** builds created in the API shim to be managed via BuildWorkload resources
**So that** I can use them as an extension point to implement different build workflows

### Acceptance Criteria

**GIVEN** a korifi installation on a k8s cluster
**WHEN I** create the first build for an App through the API shim (e.g. via `cf push`)
**THEN I** should see that a BuildWorkload resource is created
**AND** eventually the build information returned from the shim should show the `STAGED` state
**AND** I should be able to run the freshly built image without issues

---

**WHEN I** create a second build for the App through the API shim (e.g. via `cf push` with updated source)
**THEN I** should see that the existing BuildWorkload resource is updated
**AND** initially the build information returned from the shim should show the `STAGING` state
**AND** eventually it should show `STAGED` instead

---

**WHEN I** create a third build for the App using a package that will fail to build
**THEN I** should see that the existing BuildWorkload resource is updated
**AND** eventually the build information returned from the shim should show the `FAILED` state for the build

### Dev Notes

* **Though this work is written from the perspective of the API, the work is actually in the controllers**. We currently create kpack Images in the `CFBuild` controller, and that will now create `BuildWorkload` resources instead and watch them for status updates.
* **As part of this work, remove the existing code that directly creates kpack Image resources**. Remove the now-unnecessary RBAC rules
* Ensure that the `BuildWorkload` has an `ownerRef` for the `CFApp`
* Note that we're reusing the same `BuildWorkload` resource for every build of the same app. Each app has at most 1 `BuildWorkload` resource
* A later story will ensure that only 1 build runs at a time. For this story, assume that 2 builds for the same app will never run concurrently.

matt-royal:
Tasks:
- [] CFBuild reconciler now makes/watches BuildWorkload
- [] Delete now-unused code from controllers
- [] Ensure that the e2e tests install kpack-image-builder and that they pass

PR #1108, opened 2022-05-18
Build images via BuildWorkload resources

matt-royal:
#1029  

## What is this change about?

We now use BuildWorkload resources to orchestrate kpack image building

No

## Acceptance Steps
See issue


kieron-dev:
I've copied the scripts/assets/korifi-kpack-image-builder-kbld.yml file to the korifi-ci repo, as I was moving the controllers and api kbld files. They are specific to CI, so should live in the ci repo. See https://github.com/cloudfoundry/korifi-ci/commit/d9ccda78b2301b8c0a8809e56635fc58c6bf3a09

So you can safely remove that file from this PR.

PR #1109, opened 2022-05-19
Merge CRDs api groups into `korifi.cloudfoundry.org`

danail-branekov:
https://github.com/cloudfoundry/korifi/issues/1091

## What is this change about?
Change all korifi crd's api group to `korifi.cloudfoundry.org`

Changing the CRD API group is always a breaking change, however this is OK before we have an official release

## Acceptance Steps
* All tests passing
* Everything should work as it used to after korifi redeployment

@kieron-dev @georgethebeatle


matt-royal (approved):
LGTM. Thank you, @danail-branekov

PR #1111, opened 2022-05-19
Fix a flake in buildworkload_controller_test

danail-branekov:
`meta.FindStatusCondition` returns a nillable pointer which causes the
test to panic when a condition is not available yet. Therefore,
introduce a `mustHaveCondition` utility function that ensures that the
condition is available.
Also, for the sake of concise tests, use the `Eventually(func (g
Gomega){...}).Should(Succeed())` idiom

No

## What is this change about?
Fix a flake in buildworkload_controller_test

No

## Acceptance Steps
Green tests


matt-royal (approved):
LGTM

Issue #1090, opened 2022-05-16
[Feature]: Provide a warning to users who authenticate with long-expiry client certs


tcdowney:
### Background

Users authenticate with the API Shim by forwarding their Kubernetes credentials (either a token or client cert/key pair) via the `Authorization` header. The API shim then constructs a "user" Kubernetes based on these credentials per each request and uses that client to make API requests on behalf of the user. See the [Authentication proposal for more information](https://docs.google.com/document/d/1va6sE5uRi_iwGx5nzBcar145_jHq81i9SpwvpUdmsBA/edit#heading=h.mrjn69mrley4) on this.

These credentials are transmitted encrypted over TLS, not stored, and only exist in memory for the life of the request. That said, we still want to encourage users to user short-lived client certs/keys for authentication.

**As an** App Developer
**I want** the API Shim to alert me if I am using long-expiry client certs for authentication
**So that** I can improve my security practices and limit the impact of any potential credential leakage

---

### Default Warning Duration

**GIVEN** I have a client certificate with an expiration date that is greater than 1 week in the future
**WHEN I** make an authenticated API request
**THEN I** see a warning in the `X-Cf-Warnings` header with a message advising me to use a short-lived certificate

---

### Configurable Warning Duration

**GIVEN** I have a configured the API shim's `user_certificate_expiration_warning_duration` with some duration (e.g. 1 day, 1 year, etc.)
**WHEN I** use a client certificate with an expiration greater than the configured value and make an authenticated API request
**THEN I** see a warning in the `X-Cf-Warnings` header with a message advising me to use a short-lived certificate

---

### Cert that is short-lived

**GIVEN** I have a client certificate with an expiration date that is less than the default or configured duration
**WHEN I** make an authenticated API request
**THEN I** do not see a warning

### Dev Notes

---

* `user_certificate_expiration_warning_duration` is a proposed config name, feel free to suggest alternatives

tcdowney:
Validated that this works. The config property is called `userCertificateExpirationWarningDuration`.

PR #1117, opened 2022-05-20
Add check for long-lived client certificates

julian-hj:
#1090 

## What is this change about?
- The intent is to display a warning at login when a user with a
  long-lived certificate logs in.
- We do this by setting the X-Cf-Warnings header when we detect such a
  certificate. The CLI will relay the header string value to STDOUT.
- However the CLI doesn't check for warnings on the /whoami enpoint, so
  we settled on the /v3/orgs endpoint because that gets called by the
  CLI on each login.
- In the future we might consider enhancing the CLI to check for
  warnings on /whoami and then moving the check into the whoami handler.
- The default threshold for warnings is 7 days, but that value is
  configurable on the API shim configuration as
  `userCertificateExpirationWarningDuration`
- Note: we could not unit test this code because the authentication
  middleware isn't wired into the unit tests, so there's not a good way
  to pass auth info into the handler. We settled on an e2e test as the
  best option.

  [#1090]

Mostly not; it pins the kind node image to `v1.23.4` which might cause some interesting side effects. The reason for this is that we want to create client certs with a shorter expiry time and in order to do that we need to use the `expirationSeconds` field on the `CertificateSigningRequest` object which isn't supported in k8s before v1.22

If we run e2e tests on an older cluster, there's some possibility we could hit snags. It's also possible that this PR will fail e2e because we added some new mandatory env variables that might not get set by the PR check pipeline. 🤞
 
## Acceptance Steps
See #1090 

@clintyoshimura @matt-royal @tcdowney

### DIFF
--- a/api/apis/org_handler.go
+++ b/api/apis/org_handler.go
@@ -2,14 +2,16 @@ package apis
 
 import (
 	"context"
+	"crypto/x509"
+	"encoding/pem"
 	"fmt"
-	"net/http"
-	"net/url"
-
 	"github.com/go-logr/logr"
 	"github.com/gorilla/mux"
 	"github.com/gorilla/schema"
+	"net/http"
+	"net/url"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"time"
 
 	"code.cloudfoundry.org/korifi/api/apierrors"
 	"code.cloudfoundry.org/korifi/api/authorization"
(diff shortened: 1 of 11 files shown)


tcdowney (approved):
LGTM

I added a comment suggestion and had a question.

kpack-image-builder/config/manager/kustomization.yaml
```diff
@@ -13,4 +13,7 @@ kind: Kustomization
 images:
 - name: cloudfoundry/korifi-kpack-image-builder
   newName: korifi-kpack-image-builder
-  newTag: latest
+  newTag: C2297A37-57DE-4D45-ADB4-E2BBE5312282
```
tcdowney, line 16: What is this change?

akrishna90 (commented):
Looks good. I left a comment to fix up the `kustomization.yml`

kpack-image-builder/config/manager/kustomization.yaml
julian-hj: Oh--this didn't need to be in here. It's just a change that gets applied each time you build.

kpack-image-builder/config/manager/kustomization.yaml
```diff
@@ -13,4 +13,7 @@ kind: Kustomization
...
+- name: controller
+  newName: korifi-kpack-image-builder
+  newTag: 45D0A3AA-B3FB-4F53-ADE0-AA6EE85339F0
```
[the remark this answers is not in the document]
julian-hj: 👍

PR #1123, opened 2022-05-23
Don't deploy KIB in deploy-on-kind.sh for now

gcapizzi:
The kpack-image-builder deployment is currently breaking CI on `main`, so we're taking it out for now, until we're sure it's working.


kieron-dev (approved):
Yep - agree we need this to get main green. Then we can release confidently and take time fixing this properly.

PR #1127, opened 2022-05-23
Explicitly create service account token

matt-royal:
No related issue.

## What is this change about?

This PR changes our e2e test setup so that it explicitly creates the ServiceAccount token, rather than assuming that the k8s cluster will automatically make a token secret. The automatic secret creation was removed in k8s 1.24.

This PR also changes the kind image back to the default, since we no longer require k8s 1.23 or older.

No

## Acceptance Steps
Sadly, we won't really know if this PR works until it's merged and we confirm that the github actions against the commit on main pass.

### DIFF
--- a/scripts/account-creation.sh
+++ b/scripts/account-creation.sh
@@ -6,16 +6,6 @@ fi
 
 source "$SCRIPT_DIR/common.sh"
 
-getToken() {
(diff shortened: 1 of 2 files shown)


acosta11 (approved):
LGTM

PR #1128, opened 2022-05-23
lower resource requests for korifi deployments

julian-hj:
no

## What is this change about?
Lowering resource requests in an effort to get e2e tests working again in github actions.

We've seen that deployments are consistently failing trying to deploy the kpack-build-controller-manager, which is the last deployment in the installation. I'm speculating that the environment is just out of CPU, as it has a [quota of 2 cores](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) and both of our controller managers were asking for .5 cores. This change reduces the request to something a bit more modest that should still work.

- reduce the memory and CPU request sizes for the api shim, controller
  and kpack-build controller deployments
- this brings the request sizes more in line with other pods such as HNC
  and kube-rbac-proxy
- it should allow us to deploy in lower memory/cpu environments (such as
  github)

no.

## Acceptance Steps
CI passes (finally)

@matt-royal @acosta11 @gcapizzi

### DIFF
--- a/api/config/base/deployment.yaml
+++ b/api/config/base/deployment.yaml
@@ -22,7 +22,10 @@ spec:
         ports:
         - containerPort: 9000
           name: web
-        resources: {}
+        resources:
+          requests:
+            cpu: 50m
+            memory: 100Mi
         env:
         - name: APICONFIG
           value: "/etc/korifi-api-config"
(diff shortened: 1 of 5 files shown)


matt-royal (approved):
LGTM

PR #1134, opened 2022-05-24
Listing droplets and processes must tolerate 403 errors

gcapizzi:
The list of namespaces we list a resource in is just a list of namespaces containing a `RoleBinding` for the current user. This does not mean the user will be authorised for the list action we are attempting in the namespace. Therefore we should silently ignore forbidden errors during list.


kieron-dev (approved):
lgtm

PR #1140, opened 2022-05-24
Add Remaining Controller Constructors

### LINKED ISSUE #921: [Chore] Introduce constructors for all controllers.
Without constructors injecting new dependencies is a painful process over going nil reference errors in tests

gnovv:

closes #921 

## What is this change about?

Adds remaining constructors for CF Controllers

No

## Acceptance Steps

Tests pass, push app works

### DIFF
--- a/kpack-image-builder/config/manager/kustomization.yaml
+++ b/kpack-image-builder/config/manager/kustomization.yaml
@@ -13,6 +13,6 @@ configMapGenerator:
   name: manager-config
 
 images:
-  - name: cloudfoundry/korifi-kpack-image-builder
-    newName: cloudfoundry/korifi-kpack-image-builder
-    newTag: latest
+- name: cloudfoundry/korifi-kpack-image-builder
+  newName: cloudfoundry/korifi-kpack-image-builder
+  newTag: latest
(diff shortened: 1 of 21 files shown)


kpack-image-builder/config/manager/kustomization.yaml
```diff
@@ -13,6 +13,6 @@ configMapGenerator:
   name: manager-config
 
 images:
-  - name: cloudfoundry/korifi-kpack-image-builder
-    newName: cloudfoundry/korifi-kpack-image-builder
-    newTag: latest
+- name: cloudfoundry/korifi-kpack-image-builder
+  newName: korifi-kpack-image-builder
+  newTag: 122DDD51-2036-4A7B-8234-CED0D5879E6D
```
gcapizzi, line 18: This is why e2es are failing 🙂

gnovv: @gcapizzi is this safe to commit to main? 
I see that your [PR](https://github.com/cloudfoundry/korifi/pull/1141) doesn't need it?

gcapizzi: Yeah my PR is unrelated. I thought this change didn't belong here as it removes the `cloudfoundry/` prefix from `newName` (which is the reason of the failure) + it sets `newTag` to some random ID instead of `latest`.

Birdrock (approved):
/lgtm

PR #1163, opened 2022-06-01
Ignore the HTTP result on status code not 2XX

danail-branekov:
No

## What is this change about?
Resty client would only set the `result` when the HTTP status code is
2XX. Therefore, it makes no sense to verify the result when the response
code is `Forbidden`

Doing that causes flakes as the result might have been set by a previous
test that succeeded.

No

## Acceptance Steps
Green e2es

@kieron-dev


kieron-dev (approved):
lgtm

Issue #1112, opened 2022-05-19
[Feature]: Operator can install Korifi without needing HNC


julian-hj:
### Background

**As an** Operator
**I want** fewer dependencies
**So that** I can more easily run Korifi

### Acceptance Criteria

**GIVEN** I have docker and kind installed
**WHEN I** use the `deploy-on-kind.sh` script
**THEN I** see that HNC no longer gets installed
**AND I** can still use the product to create orgs and spaces and `cf push`

### Dev Notes

* A separate story will cover the updates to installation instructions.
* #769 covers the spike work we did to try this out. That work did not include watching existing roles, rolebindings, and secrets to propagate changes. It also didn't include the finalizer we will need to delete namespaces when a CFOrg or CFSpace gets deleted.
* Proposed implementation is covered in more detail in [this proposal document](https://docs.google.com/document/d/1AVZPcoOphbWU8tVJ2gM7UkEC0EvHaki6scWgp8DuCDY/edit#)
* Consider switching Eirini namespaced roles to a ClusterRole so that we don't have to propagate them. We will still need the rolebinding to be created in new namespaces, but shouldn't need to watch for updates.
* We will need to watch the root namespace for the appearance of new RoleBindings

clintyoshimura:
Changes that need to be done:

- Orgs
  - [ ] Propagation of resources at creation time
  - [ ] Propagation of added/changed/deleted resources
  - [ ] Finalizer deletion of managed namespaces  

- Spaces
  - [ ] Propagation of resources at creation time
  - [ ] Propagation of added/changed/deleted resources
  - [ ] Finalizer deletion of managed namespaces  
 
 - [ ] remove installing HNC from `deploy-on-kind` script
 - [ ] Remove code that waits for HNC in API
 - [ ] Remove code that waits for HNC in tests

clintyoshimura:
In order to switch the eirini roles to ClusterRoles we may need to make changes to eirini-controller. That might be more appropriate to do as a separate story.

julian-hj:
A/C passed

PR #1167, opened 2022-06-01
Remove HNC

akrishna90:
#1112

## What is this change about?
Removing HNC

- CFSpace reconciler
    - watches/propagates resources from `CFOrg` namespace
- CFOrg reconciler
     - watches/propagates resources from `root` namespace
- Secret matching controller-config `packageRegistrySecretName` is propagated
- All roles are propagated
- All role bindings are propagated, except role bindings with `cloudfoundry.org/propagate-cf-role` label set to `false`.
- Controller reconciles on changes to `secrets`, `roles` & `rolebindings`
- Remove unused hnc webhooks
- Remove hnc-installation steps from `scripts/install-dependencies.sh`
- Update INSTALL.md document

Yes. 
Any clusters currently running `korifi`, might have to delete and reinstall `koirifi` without HNC. 

## Acceptance Steps
see issue #1112 

@matt-royal @clintyoshimura


scripts/install-dependencies.sh
```diff
@@ -109,19 +109,12 @@ echo "*******************"
...
+echo "Installing Eirini"
 echo "*******************"
 
-kubectl apply -k ${DEP_DIR}/hnc/cf
-kubectl rollout status deployment/hnc-controller-manager --watch --namespace hnc-system
-while ! kubectl get hncconfigurations.hnc.x-k8s.io config; do
-  sleep 1
-done
-kubectl patch hncconfigurations.hnc.x-k8s.io config --type=merge -p '{"spec":{"resources":[{"mode":"Propagate", "resource": "secrets"}]}}'
+"${SCRIPT_DIR}/generate-eirini-certs-secret.sh" "*.eirini-controller.svc"
```
akrishna90, line 115: Removing HNC, I ran into `eirini-webhooks-certs` not found error during installation. 
I had to revert the change done as part of PR #1115 to get Eirini to install. 
cc @kieron-dev

controllers/webhooks/workloads/integration/helpers/subnamespace_anchors.go
```diff
@@ -2,27 +2,16 @@ package helpers
```
matt-royal, line 2: This file should probably be renamed, since we're no longer dealing with subnamespace anchors

controllers/controllers/workloads/cfbuild_controller.go
```diff
@@ -60,6 +60,13 @@ func NewCFBuildReconciler(client CFClient, scheme *runtime.Scheme, log logr.Logg
 //+kubebuilder:rbac:groups=korifi.cloudfoundry.org,resources=buildworkloads,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=korifi.cloudfoundry.org,resources=buildworkloads/status,verbs=get
 
+//+kubebuilder:rbac:groups=korifi.cloudfoundry.org,resources=buildreconcilerinfos,verbs=get;list;watch;create;update;patch;delete
```
danail-branekov, line 63: It is pretty much possible to be missing something, but I don't get why removing HNC needs changes in the buildreconcilerinfos/clusterbuilders RBAC permissions

controllers/controllers/workloads/cforg_controller.go
```diff
@@ -23,43 +23,61 @@ import (
...
 
-//+kubebuilder:rbac:groups="",resources=namespaces,verbs=get;list;watch
-//+kubebuilder:rbac:groups=hnc.x-k8s.io,resources=subnamespaceanchors,verbs=list;create;delete;watch
-//+kubebuilder:rbac:groups=hnc.x-k8s.io,resources=hierarchyconfigurations,verbs=get;list;watch;update;patch
+//+kubebuilder:rbac:groups="",resources=namespaces,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups="rbac.authorization.k8s.io",resources=roles,verbs=create;patch;delete;get;list;watch
+//+kubebuilder:rbac:groups="rbac.authorization.k8s.io",resources=rolebindings,verbs=create;patch;delete;get;list;watch
+
+//+kubebuilder:rbac:groups="",resources=events,verbs=create;update
```
danail-branekov, line 70: The org reconciler deals with namespaces, secrets, roles and rolebindings. Therefore putting kubebuilder rbac annotations for pods, statefulsets, events, etc. looks weird to me. I know that those permissions should be granted to the korifi service account, but I would rather have them in reconcilers that actually deal with such sort of objects

controllers/controllers/workloads/cforg_controller.go
```diff
@@ -78,82 +96,115 @@ func (r *CFOrgReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
...
+		return ctrl.Result{}, err
 	}
 
 	namespace, ok := getNamespace(ctx, r.client, cfOrg.Name)
 	if !ok {
 		return ctrl.Result{RequeueAfter: 100 * time.Millisecond}, nil
 	}
 
-	err = setCascadingDelete(ctx, r.client, req.Name)
+	err = propagateSecrets(ctx, r.client, r.log, cfOrg, r.packageRegistrySecretName)
```
danail-branekov, line 129: Namespace and related objects propagation is identical in both cforg and cfspace controllers, i.e. we have significant code duplication here. It has been partly addressed by extracting common functions into `shared.go` but from my point of view this is a bit "smelly". According to good OOP principles, using utility entities such as "tools" or "shared" is a smell and indicates code that the developer did not really know where to put.

I would therefore suggest extracting the common namespace handling (create/patch, propagate secrets, roles, rolebindings) in its own "namespace reconciler" that is just injected into the cforg/cfspace controllers. While running the reconciliation loop they could just delegate and if that succeeds, carry on with the org/space specific stuff

controllers/controllers/workloads/shared.go
```diff
@@ -38,29 +44,173 @@ func setStatusConditionOnLocalCopy(conditions *[]metav1.Condition, conditionType
...
-					Name:       object.GetName(),
-					UID:        object.GetUID(),
-				},
-			},
+			Name: object.GetName(),
+		},
+	}
+
+	result, err := controllerutil.CreateOrPatch(ctx, client, namespace, func() error {
+		if namespace.ObjectMeta.Labels == nil {
```
danail-branekov, line 55: nitpick: no need to use `ObjectMeta` in `namespace.ObjectMeta.Labels`, `namespace.Labels` is just fine. Same goes withe the rest of the meta fields: name, annotations, etc.

controllers/controllers/workloads/shared.go
```diff
@@ -38,29 +44,173 @@ func setStatusConditionOnLocalCopy(conditions *[]metav1.Condition, conditionType
 	})
 }
 
-func createSubnamespaceAnchor(ctx context.Context, client client.Client, req ctrl.Request, object client.Object, labels map[string]string) (v1alpha2.SubnamespaceAnchor, error) {
-	anchor := v1alpha2.SubnamespaceAnchor{
+func createOrPatchNamespace(ctx context.Context, client client.Client, log logr.Logger, object client.Object, labels map[string]string) error {
```
danail-branekov, line 47: Back in the CF-for-VMs days when we were using lager, there was sort of a rule that logger should be the first argument in the args list, this helped for better readability. I would suggest having the logger as first argument, the context as second (or vice versa, I can't really make up my mind) and then the rest. Same goes for the rest of the functions in this file

danail-branekov, line 47: Isn't the `object` arg here actually the namespace? If so, I would rename the `object` argument to `namespace` as this improves readability

controllers/controllers/workloads/shared.go
```diff
@@ -38,29 +44,173 @@ func setStatusConditionOnLocalCopy(conditions *[]metav1.Condition, conditionType
...
+
+	if err != nil {
+		return err
+	}
+
+	log.Info(fmt.Sprintf("Namespace/%s %s", object.GetName(), result))
+	return nil
+}
+
+func propagateSecrets(ctx context.Context, client client.Client, log logr.Logger, object client.Object, secretName string) error {
```
danail-branekov, line 73: Here the `object` argument is the source cforg/cfspace, in order to make that explicit I would rename the `object` argument to something like `sourceOrgSpace`

danail-branekov (changes requested):
see my comments

api/repositories/role_repository_test.go
```diff
@@ -134,8 +133,8 @@ var _ = Describe("RoleRepository", func() {
 					})
 
 					It("enables the role binding propagation, but not for cf_user", func() {
-						Expect(getTheRoleBinding(expectedName, cfOrg.Name).Annotations).NotTo(HaveKey(HavePrefix(hnsv1alpha2.AnnotationPropagatePrefix)))
-						Expect(getTheRoleBinding(cfUserExpectedName, rootNamespace).Annotations).To(HaveKeyWithValue(hnsv1alpha2.AnnotationNoneSelector, "true"))
+						Expect(getTheRoleBinding(expectedName, cfOrg.Name).Labels).To(HaveKeyWithValue(repositories.PropagateCFRoleLabel, "true"))
```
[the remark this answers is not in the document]
matt-royal: Good point. We moved this information to an annotation

controllers/controllers/workloads/cfbuild_controller.go
matt-royal: We tried removing these "extra" rbac annotations, but they end up being necessary. They're necessary since a service account can't create rolebindings that grant RBAC abilities that it doesn't already have. I moved the annotations around and added a clear comment explaining their purpose. In the long-run, we may want to find an automated way to ensure that the service account has all permissions granted by all cf_roles, but that's a large chunk of work.

controllers/controllers/workloads/shared.go
matt-royal: Good point. We took these out of this commit, but we use `ObjectMeta.*` in many places throughout the codebase and we decided not to touch existing code.

controllers/controllers/workloads/shared.go
matt-royal: We don't have a documented pattern for argument order in the codebase, but the most common order appears to be (ctx, client, logger, ...). It's also a golang convention that ctx should always be first if it's present, just like how error should be last. Perhaps we should have a team discussion on parameter order conventions, but for now we're going to leave this as is.

controllers/controllers/workloads/cforg_controller.go
matt-royal: We split out a chore to follow up on this: https://github.com/cloudfoundry/korifi/issues/1184

scripts/install-dependencies.sh
kieron-dev: I don't think reverting those changes is the correct thing to do here. The preferred method for generating webhook certs for eirini-controller-v0.3.0 is using cert-manager, and not by passing `webhooks.ca_bundle`. This is working on the main branch, so what change in this PR is breaking that?

kieron-dev: I think you may well see something like `MountVolume.SetUp failed for volume "certs" : secret "eirini-webhooks-certs" not found` when starting eirini-controller, but that is just a temporary error if cert-manager is slow to create the certs. Eirini-controller will start up eventually. If that is the problem, you should undo these changes.

akrishna90: Thanks @kieron-dev. That was exactly the error i observed on `eirini-controller`. 
The `deploy-on-kind` script kept failing on install of eirini because of it.
cc @matt-royal

matt-royal: Thank you for the context, @kieron-dev. We reverted the revert in our newest commit

PR #1178, opened 2022-06-03
Inline the apply manifest action into the manifest handler

danail-branekov:
https://github.com/cloudfoundry/korifi/issues/1058

## What is this change about?
Experiment with inlining the apply mnifest action into the manifest handler, see reasons in https://github.com/cloudfoundry/korifi/issues/1058

This draft PR is mergeable, however it also has the goal to start a discussion on whether people would like inlining actions into handler more than having them as separate actions.

* Introduce a `MnifestRepository` interface to conveniently abstract
  references to multiple repos. The interface is implemented by a struct
  that embeds concrete repo implementations
* Inline the action code into the handler
* Merge the action unit test into the handler one

No

## Acceptance Steps
* Green tests
* Agreement on whether we want to merge rest of the actions


matt-royal:
Of all of the actions, ApplyManifest is the one I'm most concerned about inlining because it's the most complex. Our original motivation for actions was to allow for code reuse and also to simplify the handlers. While the first goal of code reuse never worked out, I personally like having the simpler handler with the complex business logic encapsulated behind a simple interface. It's not important to me that we call the entity containing the logic an "Action", that it has an `Invoke` method, or whether we use it as an Interface or a Function.

tcdowney:
I left a comment on the issue, but realized it might not be visible so cross-linking here:
https://github.com/cloudfoundry/korifi/issues/1058#issuecomment-1146239138

kieron-dev:
@danail-branekov and I are in the middle of an alternative approach along those lines, simply refactoring the actions to implement some interfaces rather than function types. Just that is a great improvement for code navigability and should be uncontroversial. We'll post a link to the PR when it's ready.

danail-branekov:
Considering the feedback, we decided to have another round with an alternative approach, see https://github.com/cloudfoundry/korifi/pull/1188

Issue #1174, opened 2022-06-02
[Chore] Refactor Validating Webhooks to be Managed Webhooks Instead of Handlers


davewalter:
### Background

Our current validating webhooks are written at a fairly low level by registering an admission handler with the webhook server. For example, https://github.com/cloudfoundry/korifi/blob/17e6911025665a5c11b75552fcfa7ca6f4f6c141/controllers/webhooks/services/cfserviceinstance_validation.go#L51-L55

The `controller-runtime` package has a cleaner interface that allows us to register validators for resource types on different structs that can therefore have access to a Kubernetes client for validation against existing resources. For example, https://github.com/cloudfoundry/korifi/blob/17e6911025665a5c11b75552fcfa7ca6f4f6c141/controllers/webhooks/services/cfservicebinding_validation.go#L29-L34

The [interface for the validator](https://github.com/kubernetes-sigs/controller-runtime/blob/f46919744bee01060c9084a285e049afffd38c9d/pkg/webhook/admission/validator_custom.go#L30-L35) involves writing three separate functions for create, update and delete, instead of a single monolithic handler function.

```go
// CustomValidator defines functions for validating an operation.
type CustomValidator interface {
	ValidateCreate(ctx context.Context, obj runtime.Object) error
	ValidateUpdate(ctx context.Context, oldObj, newObj runtime.Object) error
	ValidateDelete(ctx context.Context, obj runtime.Object) error
}
```

### Action to take

Rewrite all the existing validating webhooks to use the `controller-runtime` interface/registration mechanism.

### Impact

* Simpler/cleaner interface for implementing validating webhooks.
* Aligns with our mutating webhook implementations.

PR #1185, opened 2022-06-03
Refactor Validating Webhooks

davewalter:
#1174

## What is this change about?
This PR refactors the validating webhooks to use the pattern described in [this article](https://betterprogramming.pub/writing-custom-kubernetes-controller-and-webhooks-141230820e9) instead of registering custom handlers.

No

## Acceptance Steps
Tests pass and validation rules are still applied.

@Birdrock


davewalter:
ToDo:
* Refactor CFSpace validating webhook (blocked on #1167)
* Refactor similar validating webhooks (the ones that make use of a duplicate validator) to avoid code duplication. For example, the code that converts the error returned by the duplicate validator could be moved into the duplicate validator itself.

controllers/webhooks/networking/cfroute_validation.go
```diff
@@ -59,188 +58,146 @@ var logger = logf.Log.WithName("route-validation")
...
+	if !ok {
+		return apierrors.NewBadRequest(fmt.Sprintf("expected a CFRoute but got a %T", obj))
+	}
+
+	domain, err := v.validateRoute(ctx, route)
+	if err != nil {
+		return err
+	}
+
+	return generateDuplicateValidationError(route, domain, v.duplicateValidator.ValidateCreate(ctx, logger, v.rootNamespace, uniqueName(*route)))
```
kieron-dev, line 95: It might improve readability to extract the duplicateValidator call from the method arg here. Then I would also be inclined to do the nil check here (even though that would be repeated in the update and delete validation methods) and remove the nil check from the generateDuplicateValidationError method to make it more focused.

kieron-dev (commented):
Overall, this is a really nice change. It's great to lose the code checking whether to decode the old or new objects based on the operation type.

When looking at this with @danail-branekov, we talked about whether it would be nice to separate the different validations into composable types. It might be good, but I guess that's not really the object of this PR though and would require a bit of framework code.

Should the *_validation files be renamed *_validator to match the types btw?

controllers/webhooks/networking/cfroute_validation.go
davewalter: I was contemplating moving the logic from `generateDuplicateValidationError` into the `duplicateValidator.Validate*` methods as there seemed to be a lot of duplication of handling of the error across all the webhooks.

davewalter:
Thanks for taking a look at this @kieron-dev and @danail-branekov.

> When looking at this with @danail-branekov, we talked about whether it would be nice to separate the different validations into composable types. It might be good, but I guess that's not really the object of this PR though and would require a bit of framework code.

I definitely think that this has helped to surface some potential refactors such as this, but I agree that it would be better to create follow-on chores for them, rather than delay getting this merged.

> Should the *_validation files be renamed *_validator to match the types btw?

Definitely. I'll add it to the list of items to complete before merging this.

danail-branekov:
Note that we are about to introduce a `CFTask` validator with PR https://github.com/cloudfoundry/korifi/pull/1200, could you please consider it as well when rebasing the PR?

davewalter:
> BTW, it is worth mentioning that the [kubebuilder book](https://book.kubebuilder.io/cronjob-tutorial/webhook-implementation.html) suggests that validation is implemented on top of the crd type itself, rather than having a separate validator type. This has the benefit that the validators would not need to cast the generic runtime object to a specific type before actually performing the validation.

Most (if not all) of the validating webhooks make use of K8s clients, duplicate validators, and placement validators to check for other resources on the cluster. If we implement the validators on top of the CRD types, we would have to add these things to the type structs, which would pollute them. This felt like a cleaner solution.

matt-royal (approved):
Dave and I discussed the PR on Zoom. I made some suggestions for refactoring the json encoding and decoding of errors, but otherwise this looks good to me.

PR #1188, opened 2022-06-06
Replace func types with interfaces for actions

danail-branekov:
https://github.com/cloudfoundry/korifi/issues/1058

## What is this change about?
After considering the feedback we got from PR https://github.com/cloudfoundry/korifi/pull/1178, we decided to give it another round, this time by just replacing functions with proper interfaces. 

A thing that we have not addressed here is the name `actions` - "services" in this package are not really actions because they do not implement a common interface.

Again, although this PR is mergeable, we would like to get some feedback on that approach.

### DIFF
--- a/api/actions/manifest.go
+++ b/api/actions/manifest.go
@@ -63,7 +63,7 @@ func (a *Manifest) Apply(ctx context.Context, authInfo authorization.Info, space
 	}
 
 	if exists {
-		err = a.updateApp(ctx, authInfo, spaceGUID, appRecord, appInfo)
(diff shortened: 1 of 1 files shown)


tcdowney:
This looks good to me, I like that we still maintain some of the code structure/organization benefits that we get from the actions and I don't mind getting rid of the `Invoke` function on them.

tcdowney:
This looks good to me, I like that we still maintain some of the code structure/organization benefits that we get from the actions and I don't mind getting rid of the `Invoke` function on them.

PR #1189, opened 2022-06-06
Kind local registry improvements

danail-branekov:
No

## What is this change about?
1. Use persistent storage. This _seems_ to stop the kpack errors when
   builds fail due to a kpack image not being found in the local
   registry
2. Move containerd config change to the cluster config so we don't have
   to restart containerd on the control-plane or wait 10 seconds.

We hope that this would solve the problem of droplets not being found thus making e2e tests time out

No

## Acceptance Steps
Consistent green e2e tests run

@kieron-dev


kieron-dev (approved):
Looks good

Issue #1094, opened 2022-05-17
[Feature]: CF API user can get logs from a specified start_time


acosta11:
### Blockers/Dependencies

For build logs, depends on ability to provide pod log options to the k8s client call. See #1078.

For special case start_time, depends on log timestamp sort to provide the latest set of logs. Also #1078.

### Background

**As a** CF APi client
**I want** to be able to specify a start time for my log read requests
**So that** I can review application logs for a particular time window without having to follow the underlying kubernetes resource graph to find my running pod and figure out how to query logs with kubectl

### Acceptance Criteria

**GIVEN** a kubernetes cluster with a default installation of the cf k8s shim
**WHEN I** make a request of the form
```
GET /api/v1/read/<app_guid>?start_time=<some epoch nano time>
```
**THEN I** receive a response of the form
```
{"envelopes":{"batch":[{"timestamp":1652713883018191820,"log":{"payload":"<encoded payload>","type":0}},
...
{"timestamp":1652713883018191820,"log":{"payload":"<encoded_payload>","type":0}}]}}
```
where each timestamp is greater than or equal to the provided start_time.

## Special Case - no start time
**GIVEN** a kubernetes cluster with a default installation of the cf k8s shim
**WHEN I** make a request of the form
```
GET /api/v1/read/<app_guid>descending=true&envelope_types=LOG&limit=100&start_time=-6795364578871345152
```
**THEN I** receive a response of the form
```
{"envelopes":{"batch":[{"timestamp":1652713883018191820,"log":{"payload":"<encoded payload>","type":0}},
...
{"timestamp":1652713883018191820,"log":{"payload":"<encoded_payload>","type":0}}]}}
```
where the envelopes are the latest 100 available logs.

### Dev Notes

This likely entails using the `sinceTime` pod log option from the kubernetes client-go (See the protobuff definition on which the go struct is based: [github link](https://github.com/kubernetes/api/blob/be84346886a48ff00fc676f9cb37f10e79dfd0da/core/v1/generated.proto#L3288) where applicable.

Note that this means we're blocked on passing pod log options in general to the k8s client responsible for fetching build logs. As a stopgap, we can continue to assume the latest build logs are the best candidates and add a timestamp filter in the pod repo responsible for converting those logs to envelopes. If this seems complex, we can defer the build logs implementation until after #1078.

start_time may not always be provided, act sensibly w.r.t descending and number of items returned

[...]

PR #1193, opened 2022-06-06
Added implementation of start_time to app logs

julian-hj:
#1094 

## What is this change about?
Added implementation of start_time to app logs
- truncate app logs before start_time
- clean up sleeps in unit tests
- enforce proper chronological ordering of build and run logs (instead
  of assuming that they will be in the correct order.)
- enforce length limit (since we glue together build and run logs, each
  of which might be as long as the limit.)

no.

## Acceptance Steps
#1094

@clintyoshimura

### DIFF
--- a/api/actions/read_app_logs.go
+++ b/api/actions/read_app_logs.go
@@ -2,6 +2,7 @@ package actions
 
 import (
 	"context"
+	"sort"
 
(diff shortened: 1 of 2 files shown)


clintyoshimura (approved):
LGTM

PR #1194, opened 2022-06-07
Wait for cert-manager to be ready during install-deps

kieron-dev:
No

## What is this change about?
The removal of HNC installation means the installation process happens more quickly and eirini might be installed before its dependency, cert-manager, is able to serve its registered webhooks. Eirini-controller creates a cert-manager certificate, so will fail if the webhook is not listening yet.

This change simply waits till cert-manager is ready after installing it.

No


kieron-dev:
@akrishna90 - this addresses the problem you were seeing when removing HNC.

gcapizzi (approved):

Issue #1146, opened 2022-05-26
[Chore]: Write e2e test to validate that app environment is being set effectively


matt-royal:
This description was EDITED as the story seems to be done. The instructions below cover what is left to be done as a chore

We already seem to have a [test](https://github.com/cloudfoundry/korifi/blob/d03b3dffb5343e830f31fe2d951d20d5a3d29c25/tests/e2e/apps_test.go#L387-L416) that attempts to validate that by setting an env var, restarting the app and then getting the env var, but this test does not seem to be doing the right thing, as it is getting the env vars from the api and not from the running app. 

Testing the env endpoint makes sense, but does not require app restart and is potentially a better fit for an api unit test

This test should be reworked to push an app that can dump its effective environment, like [this one](https://github.com/emalm/cf-acceptance-tests/tree/cf-k8s-node/assets/node)

clintyoshimura:
We will need to enable the e2e test that was removed as part of the the preceding issue:
https://github.com/cloudfoundry/korifi/pull/1222/files#diff-862c21f893b8783a8b6ceaf31c991167437eed916f4c9e10974fdbd09bb6cabaL306
In order to pass, we have to update the get app environment handler to return the vcap services secret ref. Currently the env is populated on the runtime workload but no longer exposed as plaintext by the api.

acosta11:
After https://github.com/cloudfoundry/korifi/commit/f0c1505948d99b3586ae0582892220f3ffa2c76d made it to `main`, the RunWorkload env is configured appropriately with secret references, so the remaining scope is only E2E testing with a custom app that dumps its env to an api client. We will circle back and reprioritize that testing enhancement once higher priority issues get finished this week.

gcapizzi:
This is effectively a chore now, as the user-facing bits have been completed, so I'm moving this to the chores column.

akrishna90:
We noticed that assets folder contained zips for apps that were being deployed in e2e tests. Having it this way would be hard to maintain as there wouldn't be a way to keep track of changes to zip. Matt brought up a great point about not maintaining out own version and instead just vendir-ing required apps from CATS.

Action Plan:
- [x] update `vendir.yml` to sync the required apps
- [ ] update `SynchronizedBeforeSuite` in e2e suite tests with a helper func that zips these apps to a tmpdir
- [ ] update the tests to looks for zips in the new tmpdir
- [ ] add the new test for fetching env variables from app

Issue #1101, opened 2022-05-17
[Feature]: kubectl user can see that BuildWorkload.spec.env uses a dedicated Secret for `VCAP_SERVICES`


matt-royal:
### Blockers/Dependencies

[#1029]

### Background

**As a** kubectl user
**I want** to see that my `VCAP_SERVICES`-encoded service credentials are stored in a secret
**So that** they aren't visible in plain text via kubectl

### Acceptance Criteria

**GIVEN** a working deployment of Korifi
**AND** a CFApp that has a CFServiceBinding to a CFServiceInstance
**WHEN I** use kubectl to describe the CFApp
**THEN I** should see that it has a `status.vcapServicesSecretName` field
**WHEN I** inspect the `Secret` named there in the same namespace
**THEN I** should see that the `Secret` contains only the `VCAP_SERVICES` key, and that its value is correct
**WHEN I** bind a second service to the app
**THEN I** should see that the same `Secret` I inspected before now includes the credential for the service instance I just bound

---

**WHEN I** create a build for that app (e.g. via `cf push`)
**THEN I** should see that the resulting `BuildWorkload` resource has a `spec.env` key that fetches the `VCAP_SERVICES` env var from the same `Secret`

For example:
```yaml
kind: BuildWorkload
metadata:
  name: 51947487-a683-4b04-a717-4e88461e4953
  namespace: cf-space-43626fac-8c65-4955-a7ec-9f753cbe76a0
spec:
  env:
  - name: VCAP_SERVICES
    valueFrom:
      secretKeyRef:
        key: VCAP_SERVICES
        name: my-app-vcap-service-secret
  #...
```

- [x] Make cool diagram
![image](https://user-images.githubusercontent.com/64824162/171736151-1d8cb80a-8355-4425-94ff-6adf64fe0340.png)
- [x] Update CFApp CRD to add status field `vcapServicesSecretName`
- [x] Update CFApp controller to create VCAPServicesSecret and update status field
- [x] Reconcile VCAPServicesSecret body in CFServiceBinding controller
  - [VCAP_SERVICES env var structure](https://docs.cloudfoundry.org/devguide/deploy-apps/environment-variable.html#VCAP-SERVICES)
- [x] Update CFBuildWorkloadController to pull in VCAPServicesSecret and add to Workload spec
- [x] Refactor the builder.go utility to decompose the app env and vcap services env secret generation
- [ ] Update ProcessController to reconcile LRPs appropriately with the changes to builder.go

[...]

PR #1222, opened 2022-06-13
Use a dedicated VCAP_SERVICES Secret

acosta11:
#1101 

## What is this change about?
Use a dedicated secret for vcap services instead of exposing sensitive information in plaintext on env var definitions

Yes, temporarily regresses the app environment response to exclude VCAP_SERVICES as noted in a todo in the E2E tests which will be subsequently restored by #1146 .

## Acceptance Steps
See #1101

@cloudfoundry/cf-k8s team


clintyoshimura (approved):
LGTM

controllers/controllers/services/cfservicebinding_controller.go
```diff
@@ -36,21 +36,29 @@ import (
...
+}
+
 // CFServiceBindingReconciler reconciles a CFServiceBinding object
 type CFServiceBindingReconciler struct {
 	client.Client
-	Scheme *runtime.Scheme
-	Log    logr.Logger
+	Scheme  *runtime.Scheme
+	Log     logr.Logger
+	Builder VCAPServicesSecretBuilder
```
danail-branekov, line 50: Provided that we have a constructor for this struct, I don't think that we need `Builder` to be public. Same goes to `Scheme` and `Log`

controllers/controllers/services/cfservicebinding_controller.go
```diff
@@ -164,6 +173,76 @@ func (r *CFServiceBindingReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+	if err != nil {
+		r.Log.Error(err, "failed to build vcap services secret", "CFServiceBinding", cfServiceBinding)
+	}
+
+	vcapServicesSecret := new(corev1.Secret)
+	// Note: is there a reason to fetch the secret name from the service instance spec?
+	err = r.Client.Get(ctx, types.NamespacedName{Name: cfApp.Status.VCAPServicesSecretName, Namespace: req.Namespace}, vcapServicesSecret)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			// Consider a follow up story to avoid this logical branch by creating and populating the secret when absent
```
danail-branekov, line 203: Why not create the story and remove this comment?

controllers/controllers/services/cfservicebinding_controller.go
```diff
@@ -164,6 +173,76 @@ func (r *CFServiceBindingReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+			meta.SetStatusCondition(&cfServiceBinding.Status.Conditions, metav1.Condition{
+				Type:    VCAPServicesSecretAvailableCondition,
+				Status:  metav1.ConditionFalse,
+				Reason:  "SecretNotFound",
+				Message: "Secret does not exist",
+			})
+		} else {
+			// Current status describes the presence of the service instance secret
```
danail-branekov, line 211: How is this comment related to setting `false` to the `VCAPServicesSecretAvailableCondition` condition?

controllers/controllers/services/cfservicebinding_controller_test.go
```diff
@@ -185,7 +196,7 @@ var _ = Describe("CFServiceBinding.Reconcile", func() {
 					})
 					It("it creates a servicebinding.io ServiceBinding with the type/provider filled in", func() {
 						Expect(fakeClient.CreateCallCount()).To(Equal(1), "Client.Create call count mismatch")
-						Expect(fakeClient.PatchCallCount()).To(Equal(1), "Client.Patch call count mismatch")
+						Expect(fakeClient.PatchCallCount()).To(Equal(2), "Client.Patch call count mismatch")
```
danail-branekov, line 199: Should the test not check the arguments to the patch operation in addition to checking the count? Why is the call count important but calls' args are not?

controllers/controllers/services/cfservicebinding_controller_test.go
```diff
@@ -218,7 +229,7 @@ var _ = Describe("CFServiceBinding.Reconcile", func() {
 					Expect(reconcileResult).To(Equal(ctrl.Result{}))
 					Expect(reconcileErr).NotTo(HaveOccurred())
 
-					Expect(fakeStatusWriter.UpdateCallCount()).To(Equal(1))
+					Expect(fakeStatusWriter.UpdateCallCount()).To(Equal(2))
```
danail-branekov, line 232: Why not check the args for the second `Update` call?

controllers/controllers/services/cfservicebinding_controller_test.go
```diff
@@ -230,9 +241,15 @@ var _ = Describe("CFServiceBinding.Reconcile", func() {
...
+				It("patches the existing servicebinding.io ServiceBinding", func() {
 					Expect(fakeClient.CreateCallCount()).To(Equal(0), "Client.Create call count mismatch")
-					Expect(fakeClient.PatchCallCount()).To(Equal(2), "Client.Patch call count mismatch")
+					Expect(fakeClient.PatchCallCount()).To(Equal(3), "Client.Patch call count mismatch")
+				})
+				It("patches the existing VCAP Services Secret", func() {
+					Expect(fakeBuilder.BuildVCAPServicesEnvValueCallCount()).To(Equal(1))
+					_, appArg := fakeBuilder.BuildVCAPServicesEnvValueArgsForCall(0)
+					Expect(appArg.Name).To(Equal(cfAppName))
+					// Expect patch/update to be called on the VCAP Services Secret from CFApp Status
```
danail-branekov, line 252: I do not see the expectation code that matches that comment. If the expectation code is there, the comment makes no sense

controllers/controllers/services/cfservicebinding_controller_test.go
```diff
@@ -318,6 +335,47 @@ var _ = Describe("CFServiceBinding.Reconcile", func() {
...
+					"Status": Equal(metav1.ConditionFalse),
+				})))
+			})
+		})
+		When("the vcap services secret isn't found", func() {
+			BeforeEach(func() {
+				fakeClient.GetReturnsOnCall(4, apierrors.NewNotFound(schema.GroupResource{}, cfApp.Status.VCAPServicesSecretName))
+			})
+
+			// Do we actually want to requeue this request? A missing secret is a system consistency error that we
```
danail-branekov, line 362: I would suggest addressing those questions in a chore/story rather than in a comment in tests

controllers/controllers/services/cfservicebinding_controller_test.go
```diff
@@ -318,6 +335,47 @@ var _ = Describe("CFServiceBinding.Reconcile", func() {
...
+			// may want to address in the reconciliation loop. For now assume the system will not reach this state.
+			It("requeues the request", func() {
+				Expect(reconcileResult).To(Equal(ctrl.Result{RequeueAfter: 2 * time.Second}))
+				Expect(reconcileErr).NotTo(HaveOccurred())
+
+				Expect(fakeStatusWriter.UpdateCallCount()).To(Equal(2))
```
danail-branekov, line 368: I feel the expectations below could be put into another `It("sets the  VCAPServicesSecretAvailable condition to false"....)`

controllers/controllers/services/integration/cfservicebinding_integration_test.go
```diff
@@ -116,69 +147,95 @@ var _ = Describe("CFServiceBinding", func() {
...
-			It("eventually resolves the secretName and updates the CFServiceBinding status", func() {
-				Eventually(func() korifiv1alpha1.CFServiceBindingStatus {
-					updatedCFServiceBinding := new(korifiv1alpha1.CFServiceBinding)
-					Expect(
-						k8sClient.Get(context.Background(), client.ObjectKeyFromObject(cfServiceBinding), updatedCFServiceBinding),
-					).To(Succeed())
+		It("eventually resolves the secretName and updates the CFServiceBinding status", func() {
+			Eventually(func() korifiv1alpha1.CFServiceBindingStatus {
+				updatedCFServiceBinding := new(korifiv1alpha1.CFServiceBinding)
+				Expect(
```
danail-branekov, line 153: According the to Ginkgo documentation, when using expectations in `Eventually` one should follow the pattern of 

```
Eventually(func(g Gomega) {
    model, err := client.Find(1138)
    g.Expect(err).NotTo(HaveOccurred())
    g.Expect(model.Reticulate()).To(Succeed())
    g.Expect(model.IsReticulated()).To(BeTrue())
    g.Expect(model.Save()).To(Succeed())
}).Should(Succeed())
```

instead of calling `Expect` directly

controllers/controllers/workloads/cfapp_controller.go
```diff
@@ -283,3 +290,54 @@ func (r *CFAppReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		}
+	}
+	originalCFApp := cfApp.DeepCopy()
+
+	cfApp.Status.VCAPServicesSecretName = vcapServicesSecretName
+
+	if cfApp.Status.ObservedDesiredState != cfApp.Spec.DesiredState {
+		cfApp.Status.ObservedDesiredState = cfApp.Spec.DesiredState
+	}
+	if cfApp.Status.Conditions == nil {
```
danail-branekov, line 334: I don't think that `cfApp.Status.Condition` can ever be nil in a `Reconcile` context. I remember trying to update `Status` with a `nil` `Condition` caused some k8s validation to fail

controllers/controllers/workloads/cfapp_controller_test.go
```diff
@@ -240,6 +249,29 @@ var _ = Describe("CFAppReconciler", func() {
...
+					secretErr = errors.New(failsOnPurposeErrorMessage)
+					_, reconcileErr = cfAppReconciler.Reconcile(ctx, req)
+				})
+
+				It("should return an error", func() {
+					Expect(reconcileErr).To(MatchError(failsOnPurposeErrorMessage))
+				})
+			})
+
+			When("fetch vcap services Secret returns a NotFoundError and create Secret returns an error", func() {
```
danail-branekov, line 264: I would suggest rephrasing this context to something like

```
When("vcap services secret does not exist"
   It("creates the vcap services secret"
  When("cresting the vcap services secret fails
    It("returns an error
```
```

File: controllers/controllers/workloads/env/builder.go [hunk 12]
```diff
@@ -8,17 +8,17 @@ import (
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 type vcapServicesPresenter struct {
-	UserProvided []serviceDetails `json:"user-provided,omitempty"`
+	UserProvided []ServiceDetails `json:"user-provided,omitempty"`
 }
 
-type serviceDetails struct {
+type ServiceDetails struct {
```
danail-branekov, line 21: The `serviceDetails` type has been made public only for the sake of being used in the integration test. That looks a bit smelly to me

tests/e2e/apps_test.go
```diff
@@ -303,7 +303,10 @@ var _ = Describe("Apps", func() {
 				Expect(result.State).To(Equal("STARTED"))
 			})
 
-			It("has the VCAP_SERVICES env var set", func() {
+			// TODO: Update the get app environment handler to return the vcap services secret ref. Negotiated as part
```
danail-branekov, line 306: All the discussion in the TODO should belong to an issue in the backlog. I believe they do, so the comment makes no sense. 
I also do not see the benefit of pending a test vs deleting it:
* If one wants to restore the test, git log is their friend
* pent tests are still compiled, so should be still maintained to at least compile in future refactoring
* while being pent, the test itself might become outdated and could trick anyone reading the code into not realising that the test is actually not run

controllers/controllers/services/cfservicebinding_controller.go
```diff
@@ -36,21 +36,29 @@ import (
...
-	return &CFServiceBindingReconciler{Client: client, Scheme: scheme, Log: log}
+func NewCFServiceBindingReconciler(client client.Client, scheme *runtime.Scheme, log logr.Logger, builder VCAPServicesSecretBuilder) *CFServiceBindingReconciler {
+	return &CFServiceBindingReconciler{Client: client, Scheme: scheme, Log: log, Builder: builder}
 }
 
 const (
-	BindingSecretAvailableCondition   = "BindingSecretAvailable"
-	ServiceBindingGUIDLabel           = "korifi.cloudfoundry.org/service-binding-guid"
-	ServiceCredentialBindingTypeLabel = "korifi.cloudfoundry.org/service-credential-binding-type"
+	BindingSecretAvailableCondition      = "BindingSecretAvailable"
```
danail-branekov, line 58: I might be overlooking something but I do not see the point of having the `BindingSecretAvailableCondition` and `VCAPServicesSecretAvailableCondition` conditions at all. Looking at the code, I can see them only being set, I would imagine that someone should be looking at them in order to justify their existence.
Or are they around for the sake of traceability?

danail-branekov (changes requested):
Please see my comments

controllers/controllers/services/cfservicebinding_controller.go
clintyoshimura: Good catch. Fixed

controllers/controllers/services/cfservicebinding_controller.go
```diff
@@ -164,6 +173,76 @@ func (r *CFServiceBindingReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
+		return ctrl.Result{RequeueAfter: 2 * time.Second}, nil
+	}
+
+	vcapServicesData, err := r.Builder.BuildVCAPServicesEnvValue(ctx, cfApp)
+	if err != nil {
+		r.Log.Error(err, "failed to build vcap services secret", "CFServiceBinding", cfServiceBinding)
+	}
+
+	vcapServicesSecret := new(corev1.Secret)
+	// Note: is there a reason to fetch the secret name from the service instance spec?
```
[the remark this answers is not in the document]
clintyoshimura: That seems to be a copy/paste error. Removed.

controllers/controllers/services/cfservicebinding_controller.go
clintyoshimura: Created chore to track this issue: https://github.com/cloudfoundry/korifi/issues/1237
Removing comment.

controllers/controllers/services/cfservicebinding_controller.go
```diff
@@ -36,21 +36,29 @@ import (
...
-	Scheme *runtime.Scheme
-	Log    logr.Logger
+	Scheme  *runtime.Scheme
+	Log     logr.Logger
+	Builder VCAPServicesSecretBuilder
 }
 
-func NewCFServiceBindingReconciler(client client.Client, scheme *runtime.Scheme, log logr.Logger) *CFServiceBindingReconciler {
-	return &CFServiceBindingReconciler{Client: client, Scheme: scheme, Log: log}
+func NewCFServiceBindingReconciler(client client.Client, scheme *runtime.Scheme, log logr.Logger, builder VCAPServicesSecretBuilder) *CFServiceBindingReconciler {
```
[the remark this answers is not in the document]
clintyoshimura: That's a good call. We see this pattern across multiple controllers so we created a chore to address it here: https://github.com/cloudfoundry/korifi/issues/1238
If it can't be addressed generically, we'll create a chore to do it specifically for this controller.

controllers/controllers/services/cfservicebinding_controller_test.go
clintyoshimura: The 2 patch function calls sets the owner ref and updates the VCAP services secret. The patch functionality is actually tested in the integration test. That verifies the owner refs are set and the VCAP services secret is updated. We didn't want to duplicate the same testing conditions in the unit test as well.

controllers/controllers/services/cfservicebinding_controller_test.go
clintyoshimura: I think the comment might have been there to explain why the It statement mentioned patch since the patch call count is verified in the previous It statement. Like mentioned previously, the result of the patch command (Updating the VCAP services secret) is tested in the integration test so the arguments aren't verified in the unit test. I'll remove the comment since it's just creating confusion.

controllers/controllers/services/cfservicebinding_controller_test.go
clintyoshimura: Should now be covered by https://github.com/cloudfoundry/korifi/issues/1237
Removed.

controllers/controllers/services/cfservicebinding_controller_test.go
clintyoshimura: Good call. That seems to make it clearer. Done

controllers/controllers/services/integration/cfservicebinding_integration_test.go
clintyoshimura: Yeah unfortunately our usage of Eventually seems to be pretty inconsistent across tests. Some places use the g Gomega arg and some don't. The docs seem to indicate we should be passing it in so I'll fix the tests here. I created this chore to go back and fix the other instances: https://github.com/cloudfoundry/korifi/issues/1240

controllers/controllers/workloads/cfapp_controller.go
```diff
@@ -283,3 +290,54 @@ func (r *CFAppReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		For(&korifiv1alpha1.CFApp{}).
 		Complete(r)
 }
+
+func (r *CFAppReconciler) CreateVCAPServicesSecretForApp(ctx context.Context, cfApp *korifiv1alpha1.CFApp) error {
```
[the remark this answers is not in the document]
clintyoshimura: Good call. Fixed.

controllers/controllers/workloads/cfapp_controller.go
clintyoshimura: I'm not 100% sure on this one, but I thought we saw this behaviour during a test. Not sure if it was an integration or e2e test though.

controllers/controllers/workloads/cfapp_controller_test.go
clintyoshimura: Isn't "creates the vcap services secret" part of the happy path though? This whole block is inside a When block for "on the unhappy path".

controllers/controllers/workloads/env/builder.go
clintyoshimura: Yeah this one was a bit odd. We were originally matching the JSON, but because it's a list with no guaranteed order, that failed once we added more than one CFServiceBinding. The only way we got that to work was to unmarshall it into a struct and use gomega's ConsistOf functionality.

tests/e2e/apps_test.go
clintyoshimura: That makes sense. I'll remove it here and add a note to the relevant story that the test will have to be replaced once the functionality is in place.

controllers/controllers/services/cfservicebinding_controller.go
clintyoshimura: I believe those status are for debugging through the K8s API. Regarding `VCAPServicesSecretAvailableCondition` not ever being set to `true`, you're right and I think the one on line 230 should be `VCAPServicesSecretAvailableCondition` and not `BindingSecretAvailableCondition` so I fixed that.

clintyoshimura:
> Please see my comments. I would also recommend having a look at the codeclimate issues as well - some could be fixed, the ones that make no sense could be marked as `wontfix` on the codeclimate UI

I took a look at these and they don't seem to be relevant. 2 of the errors were 5 return statements instead of 4. 2 of them were very short blocks of common code when creating controllers. The last 2 were very short blocks to get secrets. I didn't seem to have an option to mark as wontfix in codeclimate though.

Issue #1225, opened 2022-06-14
[Bug]: CFRoutes cannot be deleted if the CFDomain is deleted first


julian-hj:
### What happened?

When deleting an installation, if the CFDomain gets deleted first, then CFRoutes become undeletable.

### What you expected to happen

We should be able to delete CFRoutes even if the CFDomain no longer exists

### Acceptance Criteria

GIVEN a korifi installation with a space containing a route
WHEN I delete the cfdomain, and then delete the space
THEN I see that the underlying namespace was successfully deleted, along with the stuff in it.

### How to reproduce it (as minimally and precisely as possible)

```
./scripts/deploy-on-kind.sh -l -d kind
cf login
cf co o && cf target -o o && cf create-space s && cf target -s s
cf p node -p ~/workspace/cf-acceptance-tests/assets/node --no-start
cf p node2 -p ~/workspace/cf-acceptance-tests/assets/node --no-start
cf p node3 -p ~/workspace/cf-acceptance-tests/assets/node --no-start
kubectl delete cfdomains.korifi.cloudfoundry.org -n cf <hit tab here to find the guid>
cf delete-org o
kubectl get ns
```
Observe that the space namespace still exists, and still has apps and routes

PR #1227, opened 2022-06-14
CFRoute finalizer tolerates missing CFDomain

julian-hj:
[#1225]

## What is this change about?
- The CFDomain object may go missing either because the user deleted it,
  or because it got deleted first during uninstall.
- Our CFRoute controller should allow removal of routes in this
  scenario.
- This commit fixes two different failure cases that occur when the
  CFDomain no longer exists.

no

## Acceptance Steps
[#1225]

### DIFF
--- a/controllers/controllers/networking/cfroute_controller.go
+++ b/controllers/controllers/networking/cfroute_controller.go
@@ -77,6 +77,10 @@ func (r *CFRouteReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 	err = r.Client.Get(ctx, types.NamespacedName{Name: cfRoute.Spec.DomainRef.Name, Namespace: cfRoute.Spec.DomainRef.Namespace}, &cfDomain)
 	if err != nil {
 		if apierrors.IsNotFound(err) {
+			if isFinalizing(cfRoute) {
+				r.Log.Info("Warning: CFDomain not found during CFRoute deletion; proceeding to finalize", "route", req.NamespacedName, "domain", cfRoute.Spec.DomainRef)
+				return r.finalizeCFRoute(ctx, cfRoute, nil)
+			}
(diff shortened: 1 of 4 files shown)


controllers/controllers/networking/cfroute_controller.go
```diff
@@ -188,17 +192,19 @@ func (r *CFRouteReconciler) finalizeCFRoute(ctx context.Context, cfRoute *korifi
...
 
-	fqdnHTTPProxy, foundFQDNProxy, err := r.getFQDNProxy(ctx, cfRoute.Spec.Host, cfDomain.Spec.Name, cfRoute.Namespace, false)
-	if err != nil {
-		return ctrl.Result{}, err
-	}
-
-	// Cleanup the FQDN HTTPProxy on delete
-	if foundFQDNProxy {
-		err := r.finalizeFQDNProxy(ctx, cfRoute.Name, fqdnHTTPProxy)
+	if cfDomain != nil {
```
gcapizzi, line 195: A few questions:
- Do we need `cfDomain` here? It looks like `finalizeFQDNProxy` could just get `cfRoute.Status.FQDN`.
- If no, can we get rid of this `if` and always finalise the FQDN proxy?
- If yes, can we call `finalizeCFRoute` even before trying to retrieve the domain, right after retrieving the route?

julian-hj: @gcapizzi thanks, that's a good suggestion. I was thinking we needed the domain in order to determine the fqdn, but you are right that we already have that information denormalized into the route status.

I will give that change a try.

tcdowney: I assume if it is missing in the route status then we haven't reconciled the `CFRoute` and there is no way we're potentially missing an `HTTPProxy` we need to update?

PR #1244, opened 2022-06-17
Fix a timer memory leak in task repository

danail-branekov:
No

## What is this change about?
* task initialised timeout timer is created before running the select
  loop which means it is going to be created just once rather than on
  every loop
* introduce a kill channel for the controller-like loop in the
  repositories test

No

## Acceptance Steps
Run repositories test while monitoring memory usage, the test process should not use anything abnormal

@kieron-dev


kieron-dev (approved):
lgtm

PR #1253, opened 2022-06-20
Do not validate for existing app of a task

danail-branekov:
Instead, the task controller would issue an event about the missing app.
Eventually, when the app appears, the reconciler should work the task
out.

Issue:
https://github.com/cloudfoundry/korifi/issues/1082#issuecomment-1160600570

Co-authored-by: Danail Branekov <danailster@gmail.com>

https://github.com/cloudfoundry/korifi/issues/1082#issuecomment-1160600570

## What is this change about?
Do not validate for existing app of a task. Instead, the task controller would issue an event about the missing app.
Eventually, when the app appears, the reconciler should work the task
out.

No

## Acceptance Steps
See story for acceptance 

@kieron-dev


kieron-dev (approved):
Looks even better to me.

PR #1271, opened 2022-06-22
Reconcile role binding deletions for CFOrg and CFSpace

akrishna90:

https://github.com/cloudfoundry/korifi/issues/1204

## What is this change about?
Reconcile role binding deletions for CFOrg and CFSpace
- added a new label `propogated-from` on role-bindings for all propagated role-bindings
- reconcile when role bindings are deleted and check if the propagated role-bindings still exists in the parent namespace, if it doesn't go ahead and delete.

No

## Acceptance Steps

See issue

@Birdrock

### DIFF
--- a/controllers/controllers/workloads/shared.go
+++ b/controllers/controllers/workloads/shared.go
@@ -4,19 +4,18 @@ import (
 	"context"
 	"fmt"
 
-	"github.com/go-logr/logr"
-	rbacv1 "k8s.io/api/rbac/v1"
-	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
-
-	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/types"
-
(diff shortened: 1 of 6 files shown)


Birdrock (approved):
/lgtm

danail-branekov (approved):

controllers/controllers/workloads/shared.go
```diff
@@ -166,6 +171,35 @@ func propagateRoleBindings(ctx context.Context, kClient client.Client, log logr.
...
+
+	err = kClient.List(ctx, propagatedRoleBindings, &client.ListOptions{Namespace: orgOrSpace.GetName(), LabelSelector: labelSelector})
+	if err != nil {
+		log.Error(err, fmt.Sprintf("Error listing role-bindings from namespace %s", orgOrSpace.GetName()))
+		return err
+	}
+
+	for _, propagatedBinding := range propagatedRoleBindings.Items {
+		toDelete := true
+		for _, parentBinding := range roleBindings.Items {
```
kieron-dev, line 190: It's probably better to have a map pre-created here so we can do this check with O(1) complexity

akrishna90: Done

Issue #1266, opened 2022-06-22
[Feature]: Task creation should fail correctly for unstaged apps


gcapizzi:
### Background

**As a** developer
**I want** to run tasks
**So that** I can execute one-off workloads

### Acceptance Criteria

**GIVEN** an app `app1` with no current droplet set
**WHEN I** [create a task](https://github.com/cloudfoundry/korifi/issues/1017) for `app1`
**THEN I** get the following response:
```
HTTP/1.1 422 Unprocessable Entity
Content-type: application/json
```
```json
{
  "errors": [
    {
      "code": 10008,
      "detail": "Task must have a droplet. Assign current droplet to app.",
      "title": "CF-UnprocessableEntity"
    }
  ]
}
```

**GIVEN** an app `app1` which points to a non-existing current droplet
**WHEN I** [create a task](https://github.com/cloudfoundry/korifi/issues/1017) for `app1`
**THEN I** get the following response:
```
HTTP/1.1 422 Unprocessable Entity
Content-type: application/json
```
```json
{
  "errors": [
    {
      "code": 10008,
      "detail": "Task must have a droplet. Assign current droplet to app.",
      "title": "CF-UnprocessableEntity"
    }
  ]
}
```

### Dev Notes

See [this Slack thread](https://cloudfoundry.slack.com/archives/C0297673ASK/p1655894912154469) for a discussion on how to implement this.

PR #1276, opened 2022-06-23
Reject tasks on unstaged apps in the API shim

gcapizzi:

#1266 

## What is this change about?

This changes the task creation endpoint so that it returns a `422 Unprocessable Entity` if the target app is not staged. The implementation leverages the `Staged` status condition introduced in #1269.

Not really. The error message has changed from a generic 500 to the correct one.

@kieron-dev


kieron-dev (approved):
lgtm

PR #1277, opened 2022-06-23
Configure Korifi with `restricted` pod-security

danail-branekov:
https://github.com/cloudfoundry/korifi/issues/1221

## What is this change about?
Enable `restricted` pod security on Korifi "core" namespaces and pods

As this is a change in the deployment ymls - which we do not test - this PR has no tests

No. Although it technically enables the restricted pod security policy, Korifi's implementation does not violate it so no further changes are needed

## Acceptance Steps
See the story for acceptance


gcapizzi (approved):

kieron-dev (approved):
lgtm

Issue #1207, opened 2022-06-08
[Feature]: Processes that have a route mapped to them should default to having the `PORT` health check type


tcdowney:
### Background

CF for VMs defaults the health check on the `web` process type to be `PORT` since it typically will be serving web traffic. This is a convenient default because we will more accurately know when an app is "ready" and the ingress routing tier will be less likely to direct traffic to the app before it is healthy.

Currently we're setting the default health check type for all apps to `PROCESS` which only checks for an app container to having a running process in it (I think `PID 1` but don't quote me on that 🙂). For many apps this is fine, but for apps that are slow to initialize (Java apps) the process will be running before they are ready to service traffic.

**As an** App Developer
**I want** the default health check type `CFProcess` resources that have routes mapped to them to be `PORT` 
**So that** I have a more stable and reliable `cf push` experience out of the box

---

## Processes created as an output of staging
In these scenarios there is no manifest configuration of processes.

### Route mapped to an app/process type

**GIVEN** I have an app with a route mapped to one of its `web` process (`web` in this example, could be any of them)
**WHEN I** I `cf push` the app
**THEN I** see that the health check type for the `web` process is `PORT` (should also verify that the readiness probes are correct on the underlying `Pods`)

### No route mapped to an app/process type

**GIVEN** I have an app with no route mapped to its `web` process (`web` in this example, could be any of them)
**WHEN I** I `cf push` the app
**THEN I** see that the health check type for the `web` process is `PROCESS` (should also verify that the readiness probes are correct on the underlying `Pods`)

---

## Processes created via manifest
In these scenarios there is a manifest that declares the process. This triggers a different codepath that will create it.

### Route mapped to an app/process type

**GIVEN** I have an app with a route mapped to one of its `web` process (`web` in this example, could be any of them)
**WHEN I** I `cf push` the app
**THEN I** see that the health check type for the `web` process is `PORT` (should also verify that the readiness probes are correct on the underlying `Pods`)

### No route mapped to an app/process type

**GIVEN** I have an app with no route mapped to its `web` process (`web` in this example, could be any of them)
**WHEN I** I `cf push` the app
**THEN I** see that the health check type for the `web` process is `PROCESS` (should also verify that the readiness probes are correct on the underlying `Pods`)

---

### Manual health check overrides
**GIVEN** I have an app and have declared health check configuration to be `HTTP` in its manifest for the `web` process
**WHEN I** I `cf push` the app
[...]

PR #1279, opened 2022-06-23
Processes that have a route mapped to them default to having PORT health check type

clintyoshimura:
#1207 

## What is this change about?
Processes that have a route mapped to them should default to having the PORT health check type

no

## Acceptance Steps
See issue

@akrishna90


akrishna90 (approved):
paired with @clintyoshimura, lgtm!

PR #1284, opened 2022-06-28
Enforce pod security on workloads

danail-branekov:
https://github.com/cloudfoundry/korifi/issues/1220

## What is this change about?
* Set `restricted` pod-security labels on workloads namespace so that pods with containers that do not match the `restricted` policies are rejected
* Introduce a temporary mutating webhook to inject proper security context to kpack build pods to workaround https://github.com/pivotal/kpack/issues/972

No

## Acceptance Steps
See story

@kieron-dev


kpack-image-builder/config/webhook/object_selector_patch.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: admissionregistration.k8s.io/v1
+kind: MutatingWebhookConfiguration
+metadata:
+  name: mutating-webhook-configuration
+webhooks:
+- name: mkpackbuildpod.korifi.cloudfoundry.org
+  objectSelector:
+    matchExpressions:
+      - key: kpack.io/build
```
tcdowney, line 9: Can we make this apply only to pods that have some specific labels we set?
https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-objectselector

There are clusters that use kpack more broadly than just for Korifi use cases and this feels too broad.

tcdowney:
What is the urgency for this? Do we know if the kpack folks would be willing to accept a contribution for https://github.com/pivotal/kpack/issues/972 to get that support in sooner?

kpack-image-builder/config/webhook/object_selector_patch.yaml
kieron-dev: We've added an additional selector on `korifi.cloudfoundry.org/build-workload-name` which means we only make kpack build pods created by korifi.

kieron-dev (approved):
LGTM

Issue #1236, opened 2022-06-16
[Feature]: Improve "long-expiry cert warning" message


tcdowney:
### Background

In https://github.com/cloudfoundry/korifi/issues/1090 we added a warning to the `X-CF-WARNINGS` header for users authenticating with certs that have long expiration dates. Currently the warning message is:

> "Warning: Client certificate has an unsafe expiry date. Please use a short-lived certificate"

We think it would be useful if the message included the expiration date of the cert as well as the (configurable) max cert duration.

**As an** App Developer
**I want** the "long-expiry cert warning" message to contain more details
**So that** I can better act upon its advice

---

### Acceptance Criteria

**GIVEN** I am authenticating with a client cert that has an expiration date longer than the configured duration
**WHEN I** authenticate with the API and list orgs
**THEN I** see a warning message that includes the cert's expiry date and the duration that is acceptable

---

### Dev Notes

* Code for this is here: https://github.com/cloudfoundry/korifi/blob/ed779634dfb5fc335bf39391d292b1d97e4deea9/api/handlers/org_handler.go#L100
  * We should have all the info we need (user's cert and config for max duration) available at that point so this should be easy

PR #1285, opened 2022-06-28
Adds additional details to the long cert warning.

PureMunky:
Please critique this, I wanted to spend some time digging into the weeds a little and did it solo.

#1236 

## What is this change about?
Updating the messaging so that the user has a little more detail.

No

## Acceptance Steps
See #1090 

@cloudfoundry/cf-k8s

### DIFF
--- a/api/handlers/org_handler.go
+++ b/api/handlers/org_handler.go
@@ -100,7 +100,7 @@ func (h *OrgHandler) orgListHandler(ctx context.Context, logger logr.Logger, aut
 	notAfter, certParsed := decodePEMNotAfter(authInfo.CertData)
 
 	if !isExpirationValid(notAfter, h.userCertificateExpirationWarningDuration, certParsed) {
-		resp = resp.WithHeader("X-Cf-Warnings", fmt.Sprintf("Warning: Client certificate has an unsafe expiry date (%s). Please use a short-lived certificate less than %s.", notAfter.Format(time.RFC3339), h.userCertificateExpirationWarningDuration))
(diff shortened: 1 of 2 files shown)


api/handlers/org_handler.go
```diff
@@ -96,8 +97,10 @@ func (h *OrgHandler) orgListHandler(ctx context.Context, logger logr.Logger, aut
 	}
 
 	resp := NewHandlerResponse(http.StatusOK).WithBody(presenter.ForOrgList(orgs, h.apiBaseURL, *r.URL))
-	if !isExpirationValid(authInfo.CertData, h.userCertificateExpirationWarningDuration) {
-		resp = resp.WithHeader("X-Cf-Warnings", "Warning: Client certificate has an unsafe expiry date. Please use a short-lived certificate")
+	notAfter, certParsed := decodePEMNotAfter(authInfo.CertData)
+
+	if !isExpirationValid(notAfter, h.userCertificateExpirationWarningDuration, certParsed) {
+		resp = resp.WithHeader("X-Cf-Warnings", fmt.Sprintf("Warning: Client certificate has an unsafe expiry date (%s). Please use a short-lived certificate less than %s.", notAfter.Format(time.RFC3339), h.userCertificateExpirationWarningDuration))
```
tcdowney, line 103: Do you have an example of what this might return the user? When we IPMed this one we figured the hardest part would be wording this well. 😅 

If I'm reading this correctly, this will be something like `"... Please use a short-lived certificate less than 168h"`? Since that's user configurable it could be something like `86400s`. I'm wondering if there are any go helpers to make that duration friendlier.

If not, this is definitely an improvement over what we've currently got. 🙂

PureMunky: I was digging into the time.ParseDuration function and it looks like it should make this more readable. I believe that we're already using this function to read from the config but I may be wrong on that.

```
time.ParseDuration("86400s") // Output: 24h0m0s
```

PR #1296, opened 2022-06-30
Rename KorifiControllerNamespace config param

davewalter:
No

## What is this change about?
This PR renames the KorifiControllerNamespace config parameter as it is only used for the workloads TLS secret namespace.

The configuration parameter name change will be a breaking change for existing deployments.

## Acceptance Steps
Deploy Korifi following the docs and push an app.

@clintyoshimura


clintyoshimura (approved):
LGTM

PR #1304, opened 2022-07-01
Introduce task state conditions on the task object

danail-branekov:
https://github.com/cloudfoundry/korifi/issues/1031

## What is this change about?
Introduce status conditions on the task object:
* `Started` - when the eirini task starts
* `Succeeded` - when the task succeeds
* `Failed` - when the task fails

No

## Acceptance Steps
See story

@kieron-dev


kieron-dev (approved):
lgtm

Issue #1037, opened 2022-05-05
[Feature]: Developer can list all their tasks


gcapizzi:
### Background

**As a** developer
**I want** to run tasks
**So that** I can execute one-off workloads

### Acceptance Criteria

**GIVEN** organisation `org1` exists with spaces `space1` and `space2`
**AND** task `task1`, associated with app `app1` and `droplet1`, exists in `space1`
**AND** task `task2`, associated with app `app2` and `droplet2`, exists in `space2`
**WHEN I** make the following request to our shim API:
```
GET /v3/tasks
Authorization: ...
Content-type: application/json
```
**THEN I** get the following response:
  ```
  HTTP/1.1 200 OK
  Content-type: application/json
  ```
  ```json
  {
    "pagination": {
      "total_results": 2,
      "total_pages": 1,
      "first": {
        "href": "https://api.example.org/v3/tasks?page=1&per_page=2"
      },
      "last": {
        "href": "https://api.example.org/v3/tasks?page=1&per_page=2"
      },
      "next": null
      "previous": null
    },
    "resources": [
      {
        "guid": "<task1-guid>",
        "sequence_id": 20220505162517123,
        "name": "<task1-guid>",
        "state": "SUCCEEDED",
        "memory_in_mb": 1024,
        "disk_in_mb": 1024,
        "result": {
          "failure_reason": null
        },
        "droplet_guid": "<droplet1-guid>",
        "created_at": "2016-05-04T17:00:41Z",
        "updated_at": "2016-05-04T17:00:42Z",
        "relationships": {
          "app": {
            "data": {
              "guid": "<app1-guid>"
            }
          }
        },
        "links": {
          "self": {
            "href": "https://api.example.org/v3/tasks/<task1-guid>"
          },
          "app": {
            "href": "https://api.example.org/v3/apps/<app1-guid>"
          },
          "droplet": {
            "href": "https://api.example.org/v3/droplets/<droplet1-guid>"
          }
        }
      },
      {
        "guid": "<task2-guid>",
        "sequence_id": 20220505162518456,
        "name": "<task2-guid>",
        "state": "FAILED",
        "memory_in_mb": 1024,
        "disk_in_mb": 1024,
        "result": {
          "failure_reason": null
        },
        "droplet_guid": "<droplet2-guid>",
        "created_at": "2016-05-04T17:00:41Z",
        "updated_at": "2016-05-04T17:00:42Z",
        "relationships": {
          "app": {
            "data": {
[...]

PR #1317, opened 2022-07-04
Remove potential race in e2e task test

danail-branekov:
#1037

## What is this change about?
We must check list tasks includes our expected results, rather than that it equals just our expected results since other tests might be running in parallel

No

## Acceptance Steps
e2es pass


kieron-dev (approved):
lgtm

PR #1319, opened 2022-07-05
Make filtering by correlationId configurable

georgethebeatle:
...so that we can run unfiltered in CI to get more info about flaky tests


kieron-dev (approved):
lgtm

PR #1322, opened 2022-07-05
Fix flaky space validation test

danail-branekov:
https://github.com/cloudfoundry/korifi/issues/1303

## What is this change about?
Space validation involves validating whether the space's org exists. Due
to the asynchronous nature of kubernets, the space validating webhook
might not "see" the org immediately after it is being created, and
therefore intermittently reject space creation with `org not found`
error

In real life one could avoid this error by awaiting the `Ready`
condition on the org, however in envtests there is no org conrtoller
running, hence the `Ready` status is never set.

This fix puts space creation into an async assertion (`Eventually`)
which would result into retrying space creation a few times until an
expectation is met.

No

## Acceptance Steps
No space validation integration tests flakes

### DIFF
--- a/controllers/webhooks/workloads/integration/cfspace_validation_integration_test.go
+++ b/controllers/webhooks/workloads/integration/cfspace_validation_integration_test.go
@@ -10,6 +10,7 @@ import (
 	"github.com/google/uuid"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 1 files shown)


danail-branekov:
Not merging this. As [discussed on slack](https://cloudfoundry.slack.com/archives/C0297673ASK/p1657035966215229) this is a genuine problem, not just a flake

Issue #1040, opened 2022-05-05
[Feature]: Developer can cancel a task via the Kubernetes API 


gcapizzi:
### Background

**As a** developer
**I want** to run tasks via the Kubernetes API
**So that** I can execute one-off workloads without needing the `cf` CLI

### Acceptance Criteria

**GIVEN** `task1` is a task with a condition of type `"Created"` and status `"True"`
**AND** `dev1` is a `SpaceDeveloper` in `space1`
**AND** `dev2` is *not* a `SpaceDeveloper` in `space1`
**AND** `mgr` is a `SpaceManager` in `space1`
- **WHEN I** set `spec.canceled` field to `true` for `task1` as `dev1`
  **THEN I** see the task has a new condition of type `"Canceled"` and status `"True"`
  **AND** the task execution is interrupted
- **WHEN I** try to do it as `dev2` or `mgr`
  **THEN I** get a permission error

**GIVEN** a cancelled task
**WHEN** the task execution has been interrupted
**THEN I** see the task has a new condition of type `"Failed"` and status `"True"`

**GIVEN** a task with a condition of type `"Succeeded"` or `"Failed"` and status `"True"`
**WHEN I** set its `spec.canceled` field to `true`
**THEN I** get an error

kieron-dev:
**Warning** There is a possible problem with this implementation.

Sketch of implementation:
* New bool field `canceled` on task spec
* When this is set to `true`, the korifi task reconciler deletes the eirini task and sets the cancelled and failed status conditions
* Deleting the eirini task causes the k8s job to be garbage collected

Potential problems:
* We are anticipating the k8s garbage collector. We set the failed status on the korifi task regardless of what's happening to the k8s job
* Killing the job is an abrupt way of stopping a task. We already know that we can lose the latest output from the job if we kill it too soon after the process terminating (hence the eirini task TTL work), so that could happen here.
* There is a new way (since k8s 1.22) of stopping a task using the suspend field on its spec. This would work well if we didn't have to support earlier k8s versions. This would SIGTERM the job process allowing it to terminate gracefully.

So there might be a case for implementing the cancel logic in the eirini-controller, rather than what we've done in the korifi task controller. There is some reluctance to do this re-work now given we have an almost acceptable solution and we have further re-work required to unify the korifi and eirini-controller task resources.

PR #1327, opened 2022-07-06
Allow cancelling a task via kubectl

kieron-dev:

#1040

## What is this change about?

Set cftask.spec.canceled = true, and the eirini task will be deleted and the canceled status condition set on the cftask.

This also implements a webhook to prevent cancelling tasks in the succeeded or failed states

No

## Acceptance Steps

See #1040

@georgethebeatle


controllers/controllers/workloads/cftask_controller.go
```diff
@@ -249,6 +253,37 @@ func (r *CFTaskReconciler) ensureInitialized(ctx context.Context, cfTask *korifi
...
+	}
+
+	meta.SetStatusCondition(&cfTask.Status.Conditions, metav1.Condition{
+		Type:   korifiv1alpha1.TaskCanceledConditionType,
+		Status: metav1.ConditionTrue,
+		Reason: "task_canceled",
+	})
+
+	if meta.FindStatusCondition(cfTask.Status.Conditions, korifiv1alpha1.TaskStartedConditionType) != nil {
+		meta.SetStatusCondition(&cfTask.Status.Conditions, metav1.Condition{
```
danail-branekov, line 282: I am a bit worried that the status in the cache might not be up to date and miss the `Started` condition being available, hence `Failed` would be never set. Is this something we need to worry about?

danail-branekov (approved):
lgtm, besides the conditions consistency concern

PR #1332, opened 2022-07-06
Fix panics when logging with format strings

julian-hj:

No.

## What is this change about?

- the logger requires Sprintf when formatting error strings and panics if we log with an odd number of trailing arguments.
- we *think* we found all of the offending log lines and fixed them here.
- see [this example](https://go.dev/play/p/yzVGHCN3_D-) to replicate the panic we were getting in the API logs.

No.

## Acceptance Steps

CI should pass e2e tests (The API shim was panicing)
Also look for gpanic in the API shim logs

@acosta11


acosta11 (approved):
LGTM

PR #1334, opened 2022-07-07
Fix service instance env flake

georgethebeatle:

No

## What is this change about?

This change fixes the following flake:

https://ci.korifi.cf-app.com/teams/main/pipelines/main/jobs/run-e2es-main/builds/109#L62c509cc:254:286

The VCAP_SERVICES env var is coming from a secret that might not have
been set at the time of query. This fix is waiting for the VCAP_SERVICES
key to appear before inspecting its value. While this fixes the flake is
this behaviour desired? This means that users will have to retry until
they finally see their service instance related env vars

No

## Acceptance Steps

Green tests

@eirini

### DIFF
--- a/tests/e2e/apps_test.go
+++ b/tests/e2e/apps_test.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/go-resty/resty/v2"
 	. "github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
 	. "github.com/onsi/gomega"
 	. "github.com/onsi/gomega/gstruct"
 )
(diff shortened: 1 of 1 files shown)


tests/e2e/apps_test.go
```diff
@@ -446,6 +447,13 @@ var _ = Describe("Apps", func() {
 					SetResult(&result).
 					Get("/v3/apps/" + appGUID + "/env")
 			}).Should(HaveRestyStatusCode(http.StatusOK), BeNil())
+
+			Eventually(func(g gomega.Gomega) {
```
davewalter, line 451: Should this block replace the original one on line 445 as it covers the HTTP status code check?

Also, you are reimporting the `gomega` package, which is already dot-imported.

georgethebeatle:
Team agreed that this is a valid bug and test should not work around it

PR #1337, opened 2022-07-08
Include body in create space error

georgethebeatle:
https://github.com/cloudfoundry/korifi/issues/1291

## What is this change about?
Improve error logging when create{Org,Space}Raw fails by including the body from the http response

No


kieron-dev (approved):
Looks fantastic!

PR #1343, opened 2022-07-08
Avoid implicit memory aliasing in a for loop

davewalter:
## What is this change about?
This PR fixes an issue with implicit memory aliasing in a for loop.

No

## Acceptance Steps
Deploy and run tests.

### DIFF
--- a/tests/e2e/apps_test.go
+++ b/tests/e2e/apps_test.go
@@ -469,6 +469,5 @@ var _ = Describe("Apps", func() {
 				},
 			})))
 		})
-
 	})
 })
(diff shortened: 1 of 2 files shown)


Birdrock (approved):
/lgtm

tests/e2e/apps_test.go
```diff
@@ -469,6 +469,5 @@ var _ = Describe("Apps", func() {
 				},
 			})))
 		})
-
```
Birdrock, line 472: Best part of the PR, right here!

PR #1345, opened 2022-07-11
Filter tasks by sequence id

danail-branekov:

https://github.com/cloudfoundry/korifi/issues/1051

## What is this change about?
Filter tasks by sequence id

No

## Acceptance Steps
See story acceptance


api/handlers/task_handler_test.go
```diff
@@ -367,6 +367,44 @@ var _ = Describe("TaskHandler", func() {
...
+
+				When("the sequence_ids parameter cannot be parsed to ints", func() {
+					BeforeEach(func() {
+						var err error
+						req, err = http.NewRequestWithContext(ctx, "GET", "/v3/apps/the-app-guid/tasks?sequence_ids=asdf", nil)
+						Expect(err).NotTo(HaveOccurred())
+					})
+
+					It("returns an Internal Server Error", func() {
+						expectUnknownError()
```
kieron-dev, line 391: Should this be bad request?

kieron-dev (approved):
Looks good. Just unsure about whether invalid args should trigger a 503. Probably better as a 4xx error?

Issue #1349, opened 2022-07-11
[Bug]: Race condition with VCAP_SERVICE secret 


akrishna90:
### What happened?

@georgethebeatle observed a [flaky test](https://ci.korifi.cf-app.com/teams/main/pipelines/main/jobs/run-e2es-main/builds/109#L62c509cc:254:286) that's binding a service instance to an app and then immediately getting the app env and expecting the VCAP_SERVICES env var to be present with the expected key/value pairs. 

 A [PR](https://github.com/cloudfoundry/korifi/pull/1334) was opened that would prevent it from flaking (using `Eventually`), but felt more like hiding a potential problem, hence the PR was closed without merging

Real world scenario were this could become problematic - Actual running app won’t get the new environment until it is restarted, but probably there could be a race condition where a customer uses the CLI to bind a service, and then immediately restarts the app. If we don’t wait for the binding to reconcile, then we might restart prematurely. 

### What you expected to happen

Consistently bind the updated VCAP_SERVICE secret to the apps on restart. 

Consistently pass the flaky E2E test without `Eventually`. 

### Acceptance Criteria

GIVEN I have deployed `korifi`
WHEN I run `E2E` tests
THEN I Consistently see that the tests pass. 

GIVEN I have deployed `korifi` and a running app
WHEN I bind a service to the app and restart
THEN I consistently see that the pods start with the updated `VCAP_SERVICE` env var.

### How to reproduce it (as minimally and precisely as possible)

Run E2E tests until it flakes

### Anything else we need to know?

See oss slack regarding this. - https://cloudfoundry.slack.com/archives/C0297673ASK/p1657204242265429

### Environment

Revision of codebase: latest on main
Kubernetes version (use `kubectl version`):
Cloud provider or hardware configuration:
Others:

PR #1350, opened 2022-07-11
Wait for service bindings to reconcile

julian-hj:
Co-authored-by: Julian Hjortshoj <hjortshojj@vmware.com>

[#1349]

## What is this change about?
- When the API shim creates a CFServiceBinding, wait for the vcap
  services secret status to become "true" before returning, so that we
  know that the environment is available.
- This should address race conditions getting the environment or
  restarting the app after binding a service.

no
## Acceptance Steps
[#1349]

@acosta11


acosta11 (approved):
LGTM, we'll see how the tests run in CI.

Issue #1303, opened 2022-07-01
[Bug] Fix flaky space validation test


danail-branekov:
It appears that the validatingwebhook for CFSpace sometimes uses cached results and fails to retrieve the CFOrg object, and then rejects the CFSpace.

We know that in actuality, the CFOrg must exist, or the namespace where we are placing the CFSpace would not exist, so we believe that this is an issue of cache consistency.

We added the validation in order to give customers fast feedback that we don't allow CFSpaces to exist outside of CFOrg managed namespaces, but we may have to move this logic to the CFSpace reconciler, if we cannot fix the caching behavior of the client in the webhook.

**A/C**
- Given a cluster with a lot of traffic
- I can consistently make an org and then immediately make a space within that org.

**References**
https://ci.korifi.cf-app.com/teams/main/pipelines/main/jobs/run-tests-main/builds/100

```
• [FAILED] [0.048 seconds]
CFSpaceValidatingWebhook updating a space [BeforeEach] when the new space name already exists in the org namespace fails
/tmp/build/2970e19c/root/controllers/webhooks/workloads/integration/cfspace_validation_integration_test.go:129

  Begin Captured GinkgoWriter Output >>
    1.6566731875358715e+09	DEBUG	controller-runtime.webhook.webhooks	received request	{"webhook": "/validate-korifi-cloudfoundry-org-v1alpha1-cforg", "UID": "59699161-32f6-4fae-b85e-5278862d5d9b", "kind": "korifi.cloudfoundry.org/v1alpha1, Kind=CFOrg", "resource": {"group":"korifi.cloudfoundry.org","version":"v1alpha1","resource":"cforgs"}}
    1.6566731875421512e+09	DEBUG	controller-runtime.webhook.webhooks	wrote response	{"webhook": "/validate-korifi-cloudfoundry-org-v1alpha1-cforg", "code": 200, "reason": "", "UID": "59699161-32f6-4fae-b85e-5278862d5d9b", "allowed": true}
    1.6566731875520577e+09	DEBUG	controller-runtime.webhook.webhooks	received request	{"webhook": "/validate-korifi-cloudfoundry-org-v1alpha1-cfspace", "UID": "4fd5a380-c1fa-440e-8886-a177e5260385", "kind": "korifi.cloudfoundry.org/v1alpha1, Kind=CFSpace", "resource": {"group":"korifi.cloudfoundry.org","version":"v1alpha1","resource":"cfspaces"}}
[...]

PR #1353, opened 2022-07-11
Attempt to de-flake CFSpace Validator integration tests

matt-royal:

#1303 

## What is this change about?

See issue

No

## Acceptance Steps

We can't get the test to flake, so we may not be able to accept this directly. We will know that we failed if we eventually see it flake again.

@clintyoshimura

### DIFF
--- a/controllers/webhooks/workloads/integration/cfspace_validation_integration_test.go
+++ b/controllers/webhooks/workloads/integration/cfspace_validation_integration_test.go
@@ -3,6 +3,7 @@ package integration_test
 import (
 	"context"
 	"fmt"
+	"k8s.io/apimachinery/pkg/types"
 
 	korifiv1alpha1 "code.cloudfoundry.org/korifi/controllers/api/v1alpha1"
 	. "code.cloudfoundry.org/korifi/controllers/webhooks/workloads/integration/helpers"
@@ -26,12 +27,6 @@ var _ = Describe("CFSpaceValidatingWebhook", func() {
 		ctx = context.Background()
 
 		orgNamespace = "test-org-" + uuid.NewString()
-		Expect(k8sClient.Create(ctx, &v1.Namespace{
-			ObjectMeta: metav1.ObjectMeta{
-				Name: orgNamespace,
-			},
-		})).To(Succeed())
-
 		Expect(k8sClient.Create(ctx, &korifiv1alpha1.CFOrg{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      orgNamespace,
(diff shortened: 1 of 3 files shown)


clintyoshimura (approved):
LGTM

matt-royal:
We weren't able to get the test to flake, but we reasoned out that it shouldn't ever happen in production. See the comment we added for details. Short of moving this behavior out of the validator and into the reconciler (with a Condition and Event to indicate the validation failure), we couldn't think of another good solution. The test now ensures that the validator client has the CFOrg present in its cache before running the test, since that will always be the case in practice. This should avoid the flake at the cost of a lightly-broken abstraction.

danail-branekov:
While this change might solve the flake (as it is using the manager client to make sure that the org is available in its cache), I am still not convinced that this is the correct approach to solve this. Putting the manager client leaking into the test aside, is it possible that users who apply an yaml with bunch of orgs and spaces to face this issue? I think that this is a thing that has been requested by our users - be able to create all orgs, spaces, users (i.e. role bindings) in a single `kubectl apply -f some.yaml`.

As discussed on the slack thread mentioned in the description, K8S is a system based on the eventual reconciliation to a valid state, we should try to embrace this concept instead of going against the grain. Therefore, I believe that creating a space whose parent org for whatever reason is not available in the validator cache should not result into space creation error, instead, it should remain in `uninitialized`/`not ready`/whatever state and eventually be reconciled to `ready` whenever the space controller sees the org.

As per CF cli users experience, early failure is ensured by the repo itself - it would first try to [get the parent org](https://github.com/cloudfoundry/korifi/blob/b707d8550ba2c107a35312697bc9ba874bc4de62/api/repositories/space_repository.go#L76-L79) and if it is not there for whatever reason, the repo would error immediately without even trying to create the space.

controllers/webhooks/workloads/integration/cfspace_validation_integration_test.go
```diff
@@ -41,6 +36,22 @@ var _ = Describe("CFSpaceValidatingWebhook", func() {
...
+				Labels: map[string]string{korifiv1alpha1.OrgNameLabel: orgNamespace},
+			},
+		})).To(Succeed())
+
+		// Ensure that the client used by the validating webhook has the CFOrg in its cache.
+		// In practice this will always be the case because...
+		//   1. The controllers and webhooks share the same client
+		//   2. The CFSpace is created in the namespace that the CFOrgReconciler creates
+		//   3. To create the namespace, the CFOrgReconciler has to first fetch the CFOrg (ensuring it's in the cache)
+		Eventually(func() error {
```
danail-branekov, line 52: Instead of leaking the webhook client into the test, would simply putting space creation in `Eventually(<create the space via the standard k8s client>).Should(Succeed())` not have the same effect?

julian-hj (approved):
👍

controllers/webhooks/workloads/integration/cfspace_validation_integration_test.go
matt-royal: Either solution would get the test passing, but I prefer that we ensure the client cache is synchronized first since it will not mask other kinds of failures with CFSpace validation. Once we establish that the cache is synchronized, we know that any failures to create the CFSpace are from a different, possibly legitimate bug.

matt-royal:
I wasn't in the cross-team sync this morning, but it sounds like y'all agreed that we should merge this so I went ahead and did it. If we keep seeing the flake we can try another solution.

Issue #1022, opened 2022-05-04
[Feature]: Developer can get a task


gcapizzi:
### Background

**As a** developer
**I want** to run tasks
**So that** I can execute one-off workloads

### Acceptance Criteria

**GIVEN** `app1` exists in `org1` and `space1`
**AND** `droplet1` is the droplet used by `app1`
**AND** `dev1` is a `SpaceDeveloper` in `space1`
**AND** `dev2` is *not* a `SpaceDeveloper` in `space1`
**AND** `task1` exists for `app1`
- **WHEN I** make the following request to our shim API as `dev1`:
  ```
  GET /v3/tasks/<task1-guid>
  Authorization: ...
  Content-type: application/json
  ```
  **THEN I** get the following response:
    ```
    HTTP/1.1 200 OK
    Content-type: application/json
    ```
    ```json
    {
      "guid": "<task1-guid>",
      "sequence_id": 202005041656123,
      "name": "<task1-name>",
      "command": "<task1-cmd>",
      "state": "RUNNING",
      "memory_in_mb": 1024,
      "disk_in_mb": 1024,
      "result": {
        "failure_reason": null
      },
      "droplet_guid": "<droplet1-guid>",
      "created_at": "2016-05-04T17:00:41Z",
      "updated_at": "2016-05-04T17:00:42Z",
      "relationships": {
        "app": {
          "data": {
            "guid": "<app1-guid>"
          }
        }
      },
      "links": {
        "self": {
          "href": "https://<shim-domain>/v3/tasks/<task1-guid>"
        },
        "app": {
          "href": "https://<shim-domain>/v3/apps/<app1-guid>"
        },
        "droplet": {
          "href": "https://<shim-domain>/v3/droplets/<droplet1-guid>"
        }
      }
    }
    ```
- **WHEN I** try to do the same as `dev2`
  **THEN I** get the following response:
  ```
  HTTP/1.1 404 Not Found
  Content-Type: application/json
  ```
  ```json
  {
    "errors": [
      {
        "code": 10010,
        "title": "CF-ResourceNotFound",
        "detail": "Task not found"
      }
    ]
  }
  ```

PR #1354, opened 2022-07-12
SpaceManager should be able to get and list tasks

gcapizzi:

This should have been part of #1022, #1037 and #1038.

## What is this change about?

This allows users who are space managers to get and list tasks in their spaces.

No.


georgethebeatle (approved):
LGTM, how come we did not realize we are missing it? Are we missing tests somewhere?

gcapizzi:
I think we just don't test the different role permissions, as they're technically not part of the code.

Issue #1286, opened 2022-06-28
[Feature]: RoleBinding propagation should be opt-in instead of opt-out


tcdowney:
### Background

Our current approach of propagating all RoleBindings down from `cf` -> `cf-org-` -> `cf-space-` namespaces by default has caused several issues.

1. `kubectl` users who are applying `RoleBindings` manually need to be aware that they must set the `cloudfoundry.org/propagate-cf-role: "false"` annotation to opt-out of propagating for certain bindings. For example, if someone manually applies a "CF User" `RoleBinding` that should only exist in the root namespace and forgets the annotation then the controllers will propagate that to every namespace which confuses our [namespace permissions checker](https://github.com/cloudfoundry/korifi/blob/8dd17c2f4978250931d2c62badce71a5cfd946e0/api/authorization/namespace_permissions.go). That's one example, but in general I feel an "opt-in" mode of propagation is less likely to surprise users and cause unintended issues.
2. Some Kubernetes distributions (such as OpenShift) and role management systems will be applying their own `RoleBindings` automatically that confer permissions that our controller does not have permission to copy. These systems automatically manage these bindings and prevent / make it difficult to opt-out via the annotation. If we switch to an "opt-in" system then these externally managed `RoleBindings` will just be ignored.

**As a** Platform Operator
**I want** `RoleBinding` propagation to be "opt-in" instead of "opt-out"
**So that** I can install Korifi on a wider variety of Kubernetes distributions and do not need to worry about other RBAC management systems conflicting with it.

---

#### No annotation

**GIVEN** I have installed Korifi
**WHEN I** make a new `RoleBinding` in one of the CF namespaces (root `cf` or an org namespace)
**THEN I** see it **IS NOT** propagated to the child namespaces

#### `cloudfoundry.org/propagate-cf-role: "true"`

**GIVEN** I have installed Korifi
**WHEN I** make a new `RoleBinding` in one of the CF namespaces (root `cf` or an org namespace)
**AND I** annotate it with `cloudfoundry.org/propagate-cf-role: "true"`
**THEN I** see it **IS** propagated to every child namespace

#### `cloudfoundry.org/propagate-cf-role: "false"` (or any value that's not truthy)

**GIVEN** I have installed Korifi
**WHEN I** make a new `RoleBinding` in one of the CF namespaces (root `cf` or an org namespace)
**AND I** annotate it with `cloudfoundry.org/propagate-cf-role: "false"`
[...]

PR #1360, opened 2022-07-12
Use opt-in strategy for role-binding propagation

akrishna90:

#1286

## What is this change about?

Only role-bindings with annotation `cloudfoundry.org/propagate-cf-role` set to `true` will be propagated

No

## Acceptance Steps

See issues #1286


akrishna90:
This PR blocked by the `eirini-controller` change we have here - https://github.com/cloudfoundry/eirini-controller/pull/21
Once `eirini-controller` PR is merged and a new release is cut, we can merge this PR

Issue #1187, opened 2022-06-06
[Feature]: Set CPU weights based on App memory on RunWorkloads


akrishna90:
### Blockers/Dependencies

#1145 

### Background

**As a** app developer
**I want** apps to have some amount of guaranteed CPU
**So that** they do not get completely starved for CPU resources

---
For CPU, the CPU requests on the pod should at least be nonzero to guarantee some degree of CPU availability for the app process. In CF, CPU has not been allocatable directly, and instead has been allocated proportionally to the memory allocation.

### CF on VMs
In CF on VMs, this model has been more flexible than Kubernetes, because the Diego cells do not treat CPU as an explicitly allocatable resource and so do not place an independent bound on the total quantity of CPU shares. Nevertheless, in most cases 8 GB of memory has corresponded to a CPU weight of 100 on a Diego LRP, and then into 1024 CPU shares on a Linux host. We commonly see that 4 CPU x 32 GB is a common size for Diego cell VMs, as a moderately sized high-memory VM configuration on many clouds (GCP n2-highmem, Azure E-series, AWS r-series), so maxing out the memory allocation on the Diego cell means that those 1024 CPU shares correspond to a minimum of 1 CPU of availability.

### Kubernetes
In Kubernetes, the kubelet makes the host CPU available as an explicitly allocatable quantity (in units as granular as millicores), minus any overhead for its own CPU usage. Internally, it translates the CPU requests into CPU shares at 1024 per CPU core.  Since Pods can burst their CPU usage above their requests up to the CPU limit (assuming no other CPU contention), it's also ok to reduce the CPU-to-memory allocation ratio slightly to avoid making CPU the constraining resource to allocate. 

A default of **100m** of CPU request per **1 GiB(1024 MiB)** of memory limit/request would be close to the behavior of CF in practice without going over. We should also make sure that each app has some minimum CPU request (say, **5m**) even if its memory allocation is very small (say, **32 MiB**).

~~- CPU Limits
For CPU limits, which are not exclusively allocated, we should consider setting some high limits as a guard against extreme noisy neighbors without otherwise constraining the flexibility of CPU consumption. One scheme that might work is to set the limit as double the CPU request value, with a minimum limit of 1 CPU to avoid constraining low-memory apps too much.~~

### Acceptance Criteria

**GIVEN** I `cf push test-app -m 1024M`
**WHEN I** inspect the `pod` for `test-app` 
**THEN I** I see `100m` CPU Request set on the pods
[...]

PR #1361, opened 2022-07-12
Set CPU requests and limits based on app memory

clintyoshimura:
#1187

## What is this change about?
Set the StatefulSet CPU requests and limits based on the app memory in RunWorkloads

no

## Acceptance Steps
See issue

@acosta11

### DIFF
--- a/controllers/controllers/workloads/cfprocess_controller.go
+++ b/controllers/controllers/workloads/cfprocess_controller.go
@@ -21,7 +21,6 @@ import (
 	"crypto/sha1"
 	"errors"
 	"fmt"
-	"math"
 	"sort"
 	"strconv"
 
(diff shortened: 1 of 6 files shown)


acosta11 (approved):
LGTM, and we added a note for the team to consider our strategy before merging so leaving auto-merge turned off for now.

danail-branekov:
@kieron-dev and I had a look at the PR and have a couple of concerns:

1) The cpu/memory ratio is not configurable, i.e. an app would always get 100 millicores per 1G of memory. Would such a ratio always work for every cluster? If the ratio is configurable, then operators would be able to do the maths so that they accomodate the apps they expect to run on the cluster.
2) Do we really need to set cpu limits? Back in time when we were working on the Garden CPU throttling feature we came to the conclusion that
    a) setting hard limits on containers causes initialisation/startup routines for apps to occasionally timeout (for example, Spring apps)
    b) that leaves CPU underutilised as there would be noone to use it (imagine a cluster with one or two apps that are limited to a CPU)
    c) it is hard (even impossible) to come up with reasonable limits that would allow apps to burst effectively while restricting noisy neighbours

clintyoshimura:
@danail-branekov @kieron-dev @emalm @tcdowney @acosta11 
After some discussion, we added a proposal regarding the CPU limit calculation on the issue (https://github.com/cloudfoundry/korifi/issues/1187#issuecomment-1183539480).
We can continue the discussion there.
Thanks

clintyoshimura:
Based on discussion of updated issue requirements, closing this PR in favor of #1382

PR #1364, opened 2022-07-13
`CFTask.canceled` should not be a required field

gcapizzi:

Problem introduced in #1040 (PR #1327).

## What is this change about?

This removes the `required` constraint on the `canceled` field of `CFTask`, as a task should not be considered canceled by default and it shouldn't be necessary to specify `canceled: false` to create a `CFTask`.


danail-branekov (approved):

kieron-dev (approved):
lgtm

PR #1377, opened 2022-07-14
Do not retry on webhook validation errors

danail-branekov:
No

## What is this change about?
The k8s retrying client does not retry on Korifi webhook validation errors

Webhook validation errors are received by the shim as forbidden errors that
have a status reason containing the json marshalled validation error and status
code `Forbidden`.

It makes no sense for the retrying client to retry the operation (e.g. `Patch`)
on such forbidden errors as the validating webhook would consistently return
the same error over and over again

No

## Acceptance Steps
Green tests


kieron-dev (approved):
Yep - this makes a lot of sense.

PR #1379, opened 2022-07-14
Remove no-op controllers overlays

georgethebeatle:

No

## What is this change about?

We historically needed overlays for controllers so that the kpackImageTag config property could be changed between deployment targets. Since that configuration option moved to the kpack-image-builder, controllers no longer needs overlays to modify the configmap.

We leave the overlay to add debug functionality.

The makefile has been modified to just perform `deploy-controllers` when `deploy-controllers-kind-local` is called. This could be removed at a later date if nobody's scripts are using it.

No

## Acceptance Steps

`make test-e2e` works on a workstation.


kieron-dev:
We now need an overlay for controllers on kind to set the taskTTL. So let's abandon this.

PR #1401, opened 2022-07-19
Await `Staged` condition on setting the app droplet

danail-branekov:
https://github.com/cloudfoundry/korifi/issues/1385

## What is this change about?
Await the `Staged` app condition on set app droplet

By doing that we ensure that the app controller has accepted the request
to set the app droplet and it is aware that the app is staged. Thus
subsequent creation of objects that depend on the app being staged (such
as tasks) would not fail.

Awaiting for a status condition of a task requires adding the `watch`
permission on apps for the `cf_admin` and `space-developer` roles

No

## Acceptance Steps
n/a


kieron-dev (approved):
Looks good.

With @georgethebeatle, we added generics 😲 as it seemed a perfect fit.

PR #1409, opened 2022-07-20
Use context.WithTimeout in client.Watch

georgethebeatle:

No

## What is this change about?

No need to create our own timer and channel as we can use a context with
timeout

No

## Acceptance Steps

Green tests

@kieron-dev


kieron-dev (approved):
lgtm

Issue #1420, opened 2022-07-21
[Bug] CFRoute Finalizer Blocks Uninstall of Korifi


davewalter:
### What happened?

When users attempt to uninstall Korifi without first deleting custom resources created using its CRDs, the deletion of CFRoute resources is prevented by the finalizer, which is attempting to clean up other resources. Here are some sample errors from the `korifi-controller` pod:

```
manager 1.658434462711873e+09    INFO    controllers.CFRoute    Reconciling deletion of CFRoute/cf-route-252af5ce-53a1-494b-981a-49329cc711d7
manager 1.6584344627145164e+09    DEBUG    controller-runtime.webhook.webhooks    received request    {"webhook": "/mutate-korifi-cloudfoundry-org-v1alpha1-cfroute", "UID": "fc7523c3-ee2a-44cf-9009-20d974e060d8", "kind": "korifi.cloudfoundry.org/v1alpha1, Kind=CFRoute", "resource": {"group":"korifi.cloudfoundry.org","version":"v1alpha1","resource":"cfroutes"}}
manager 1.658434462714725e+09    INFO    cfroute-resource    Mutating CFRoute webhook handler    {"name": "cf-route-252af5ce-53a1-494b-981a-49329cc711d7"}
manager 1.6584344627150435e+09    DEBUG    controller-runtime.webhook.webhooks    wrote response    {"webhook": "/mutate-korifi-cloudfoundry-org-v1alpha1-cfroute", "code": 200, "reason": "", "UID": "fc7523c3-ee2a-44cf-9009-20d974e060d8", "allowed": true}
manager 1.6584344627168453e+09    DEBUG    controller-runtime.webhook.webhooks    received request    {"webhook": "/validate-korifi-cloudfoundry-org-v1alpha1-cfroute", "UID": "e90dc041-422f-48d4-a91b-5afe0cf50367", "kind": "korifi.cloudfoundry.org/v1alpha1, Kind=CFRoute", "resource": {"group":"korifi.cloudfoundry.org","version":"v1alpha1","resource":"cfroutes"}}
manager 1.6584344627170675e+09    ERROR    route-validation    Error while retrieving CFDomain object    {"error": "CFDomain.korifi.cloudfoundry.org \"5b5032ab-7fc8-4da5-b853-821fd1879201\" not found"}
manager code.cloudfoundry.org/korifi/controllers/webhooks/networking.(*CFRouteValidator).fetchDomain
[...]

PR #1421, opened 2022-07-21
Cleanup finalizing checks

davewalter:
#1420 

## What is this change about?
This PR cleans up the finalizing checks in the controllers and adds checks to all validating webhooks to allow the update if the resource is currently being deleted (non-zero deletion timestamp) as the change is likely the removal of the finalizer, which shouldn't be blocked.

No

## Acceptance Steps
See #1420 

@Birdrock

### DIFF
--- a/controllers/controllers/networking/cfroute_controller.go
+++ b/controllers/controllers/networking/cfroute_controller.go
@@ -73,7 +73,7 @@ func (r *CFRouteReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
-	if isFinalizing(cfRoute) {
+	if !cfRoute.GetDeletionTimestamp().IsZero() {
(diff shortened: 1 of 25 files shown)


Birdrock (dismissed):

controllers/controllers/networking/cfroute_controller.go
```diff
@@ -461,10 +461,6 @@ func (r *CFRouteReconciler) fetchServicesByMatchingLabels(ctx context.Context, l
 	return &serviceList, nil
 }
 
-func isFinalizing(cfRoute *korifiv1alpha1.CFRoute) bool {
```
tcdowney, line 464: Why remove the helper function? I feel like it provides some quicker context to what `!cfRoute.GetDeletionTimestamp().IsZero()` means. But maybe could be renamed to `isBeingDeleted` or `markedForDeletion` or something. 🤷

tcdowney (approved):
LGTM once CodeClimate check is green

I have a preference to making the change I suggested, but not enough to block a review over it if y'all disagree.

Issue #1250, opened 2022-06-20
[Chore]: Migrate pairup and postfacto to the eirini acceptance environment


georgethebeatle:
### Background

We are using pairup and postfacto as an integral part of our daily work. These apps are still running on eirini infrastructure on top of cf-for-k8s. This means we still need to maintain an old acceptance environment, which is a waste

### Action to take

- Migrate pairup and postfacto to the korifi acceptance environment
- Setup some uptime monitor such as https://www.statuscake.com/
- Tear down unneeded eirini infrastructure

### Impact

We reduce out overall cost of maintainance by retiring old infrastructure

kieron-dev:
I had a look at doing uptime checks in GCP. It's really easy: https://console.cloud.google.com/monitoring/uptime?project=cf-on-k8s-wg. After I added it, it also offered to check SSL cert validity, so I turned that on too. You can see that in the alerts tab.

PR #1436, opened 2022-07-26
Allow korifi apps to use websockets

georgethebeatle:
#1250

## What is this change about?
It turns out that Contour is disallowing websockets by default. This
breaks apps like postfacto. This commit allows all apps to use
websockets.

No

## Acceptance Steps
Deploy an app that uses websockets (e.g postfacto) and see it wokring. With postfacto working means that different browsers see live updates on the retro board.

@kieron-dev

### DIFF
--- a/controllers/controllers/networking/cfroute_controller.go
+++ b/controllers/controllers/networking/cfroute_controller.go
@@ -298,7 +298,8 @@ func (r *CFRouteReconciler) createOrPatchRouteProxy(ctx context.Context, cfRoute
 					Conditions: []contourv1.MatchCondition{
 						{Prefix: cfRoute.Spec.Path},
 					},
-					Services: services,
+					Services:         services,
+					EnableWebsockets: true,
 				},
 			}
 		}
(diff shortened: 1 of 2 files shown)


tcdowney:
Re: my questions on [Gateway API](https://projectcontour.io/) support for websockets from the WG meeting. I did some quick digging, doesn't look like backend protocol tweaks like this are supported yet, but it is on their radar.

Relevant issues:
* https://github.com/kubernetes-sigs/gateway-api/issues/205
* https://github.com/kubernetes-sigs/gateway-api/issues/1282

I don't think uncertainty there should block us from enabling websockets on Contour because it's a pretty big CF app compatibility gap not to support them, but this will delay us from adopting Gateway API in any non-experimental capacity I think.

tcdowney:
Other question I had: turning this config on broadly doesn't break apps that don't support websockets, right?

Birdrock (approved):
As discussed at the bi-weekly Cloud Foundry on Kubernetes Working Group Forum meeting 7/26, this seems reasonable. The default of Envoy is to disable Websockets, so Contour follows in this regard. Enabling Websockets does increase surface area, but is a widespread and fairly well known area.

kieron-dev:
> Other question I had: turning this config on broadly doesn't break apps that don't support websockets, right?

Postfacto is using websockets _and_ http on the same port, and it works ok. I don't see any reason why it might break other apps not using websockets. I think we should just apply this and keep an eye on it.

Issue #1427, opened 2022-07-22
[Feature]: Operator can switch to alternative Build reconciler


matt-royal:
### Background

**As an** operator
**I want** to be able to use alternative components to build my apps
**So that** I can take advantage of new/different features

### Acceptance Criteria

**GIVEN** a default deployment of korifi
**WHEN** I push an app
**THEN** I see that my image is built via a kpack Image (i.e. via kpack-image-builder)
**AND** I can see that a BuildWorkload resource was made with the `spec.reconcilerName` set to `kpack-image-builder`

---

**GIVEN** a deployment of korifi where I've changed the build reconciler via config to "non-existent-reconciler"
**WHEN** I push an app
**THEN** I can see that a BuildWorkload resource was made with the `spec.reconcilerName` set to "non-existent-reconciler"
**AND** I can see that my build never completes and eventually times out

### Dev Notes

These tasks are included in this story:

- [ ] Add `spec.reconcilerName` to the `BuildWorkload` type. It's a required field.
- [ ] Add a configuration option to `korifi/controllers` for which build reconciler it will use. This should default to `kpack-image-builder`. All new `BuildWorkload` resources created by `korifi/controllers` will have this value set as their `spec.reconcilerName`
- [ ] Update `kpack-image-builder` so that it ignores `BuildWorkload` resources where the `spec.reconcilerName` is not `kpack-image-builder`.

This story will allow alternative `BuildWorkload` reconcilers to be deployed to the same cluster. All reconcilers MUST ignore Workloads that are not explicitly assigned to them via `spec.reconcilerName`. For now an operator can change back and forth via the config, but in the future we can add more sophisticated ways to assign individual builds to different reconcilers.

This pattern is similar to the ingress and gateway resources

PR #1440, opened 2022-07-26
Implement the spec reconcileName in BuildWorkload

clintyoshimura:
#1427 

## What is this change about?
Implement the reconcileName in BuildWorkloads so builders will only produce images if the reconciler name matches

no

## Acceptance Steps
See issue

@davewalter

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,3 +1,3 @@
-export KUBEBUILDER_ASSETS=$PWD/testbin/bin
+source <(setup-envtest use -p env --bin-dir "$PWD/testbin")
 export APICONFIG=$PWD/config/base/apiconfig/
 export CONTROLLERSCONFIG=$PWD/config/base/controllersconfig/
--- a/controllers/controllers/workloads/cfbuild_controller_test.go
+++ b/controllers/controllers/workloads/cfbuild_controller_test.go
@@ -110,7 +110,9 @@ var _ = Describe("CFBuildReconciler", func() {
 			fakeClient,
 			scheme.Scheme,
 			zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)),
-			&config.ControllerConfig{},
+			&config.ControllerConfig{
+				BuildReconciler: "myCustomBuildReconciler",
+			},
 			fakeEnvBuilder,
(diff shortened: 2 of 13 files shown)


.envrc
```diff
@@ -1,3 +1,3 @@
-export KUBEBUILDER_ASSETS=$PWD/testbin/bin
```
Birdrock, line 1: Darwin is specific to Mac OS and will be nonfunctional on linux systems. It is also specific to a version, which may change.

We should probably think about how to resolve this or remove the .envrc file from the project and add a .gitignore for people to configure for themselves.

Birdrock (changes requested):

.envrc
Birdrock: Looks like it highlighted just one half of the diff...

davewalter: We replaced this with a call to `setup-envtest` to retrieve the appropriate value.

davewalter (approved):

Birdrock (approved):

controllers/controllers/workloads/cfbuild_controller_test.go
```diff
@@ -170,6 +172,14 @@ var _ = Describe("CFBuildReconciler", func() {
 
 				Expect(actualWorkload.Spec.Env).To(Equal(buildEnv))
 			})
+
+			It("sets the build reconciler from the controller config", func() {
+				Expect(fakeClient.CreateCallCount()).To(Equal(1), "fakeClient Create was not called 1 time")
+				_, obj, _ := fakeClient.CreateArgsForCall(0)
```
matt-royal, line 178: I thought we'd agreed not to add these optional assertion descriptions except for truly confusing/unclear circumstances. They mostly just clutter the code.

davewalter: IMHO, it is more about the clarity of the test output than the code. Any time you see a test failure where Ginkgo expected 0 to equal 1 (or some variation of this), it can be extremely confusing as to what is failing without digging into the code.

I think that at least part of the problem is that we are not consistent about adding these helper descriptions.

Issue #1439, opened 2022-07-26
[Bug]: Fix issue with running ruby apps


akrishna90:
### Background

We pinned the version of ruby buildpacks to `v0.11` in this PR #1438 

With Ruby apps, the builds succeeds but the pods fail to start with the latest version. In our controllers, after the build succeeds, we [extract the `start-command` from the built image](https://github.com/cloudfoundry/korifi/blob/c5de457248ebeb7aaad21768a905bce0911b2ec0/kpack-image-builder/controllers/imageprocessfetcher/image_process_fetcher.go#L51-L57) and set it on the cf-process. This seems to work fine with ruby:0.11 but is broken on the latest version.

Inspecting it further, we verified that the `start-command` on the image is the same as the start command we set on the cfprocess

```
   Paketo Rackup Buildpack 0.4.0
     Writing start command
     Assigning launch processes:
       web (default): bash -c bundle exec rackup --env RACK_ENV=production -p "${PORT:-9292}"
```

Removing the command from the cfprocess fixes the issue. As a workaround, we decided to pin the version to 0.11 until we figure out how we want to fix this.

We identified the difference with the `start command` b/w v0.11 & latest
With v0.11
```
bundle exec rackup --env RACK_ENV=production -p "${PORT:-9292}"
```

With latest version
```
 bash -c bundle exec rackup --env RACK_ENV=production -p "${PORT:-9292}"
```

It appears that quoting the command after `bash -c` fixes the issue
```
 bash -c "bundle exec rackup --env RACK_ENV=production -p \"${PORT:-9292}\""
```

### Action to take

@davewalter had the suggestion about not setting the command on the cfprocess resource. It appears that the `command` attribute is not a required field on the `RunWorkload` anyway. 

We have opened a discussion with the `packeto-buildpack` team
 See details - https://paketobuildpacks.slack.com/archives/CULAS8ACD/p1658873060962669

### Impact

This will allow us to use the latest buildpacks for ruby

Birdrock:
>If the -c option is present, then commands are read from string. If there are arguments after the string, they are assigned to the positional parameters, starting with $0.

Birdrock:
It appears this is a result of a change in the rackup buildpack: https://github.com/paketo-buildpacks/rackup/blame/8cc50547eeeda8781e0ec944997514394f2f95f0/build.go#L53-L63

Birdrock:
[...]

PR #1442, opened 2022-07-27
Fix command string truncation in bash string cmds

Birdrock:
#1439 

## What is this change about?
Fixes an issue with how command and args were processed. The issue lead to a situation where args would be dropped.

No.

## Acceptance Steps
Push a ruby rackup app (e.g.: CATS hello-world)

@akrishna90 @davewalter @matt-royal

### DIFF
--- a/kpack-image-builder/controllers/imageprocessfetcher/image_process_fetcher.go
+++ b/kpack-image-builder/controllers/imageprocessfetcher/image_process_fetcher.go
@@ -3,7 +3,6 @@ package imageprocessfetcher
 import (
 	"encoding/json"
 	"fmt"
-	"strings"
 
 	korifiv1alpha1 "code.cloudfoundry.org/korifi/controllers/api/v1alpha1"
 
@@ -66,8 +65,11 @@ func (f *ImageProcessFetcher) Fetch(imageRef string, credsOption remote.Option, 
 
 // Reconstruct command with arguments into a single command string
 func extractFullCommand(process launch.Process) string {
-	commandWithArgs := append([]string{process.Command}, process.Args...)
-	return strings.Join(commandWithArgs, " ")
+	cmdString := process.Command
+	for _, a := range process.Args {
+		cmdString = fmt.Sprintf(`%s %q`, cmdString, a)
+	}
+	return cmdString
 }
 
 func extractExposedPorts(imageConfig *v1.Config) ([]int32, error) {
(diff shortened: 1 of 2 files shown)


kpack-image-builder/controllers/imageprocessfetcher/image_process_fetcher.go
```diff
@@ -66,8 +65,11 @@ func (f *ImageProcessFetcher) Fetch(imageRef string, credsOption remote.Option,
 
 // Reconstruct command with arguments into a single command string
 func extractFullCommand(process launch.Process) string {
-	commandWithArgs := append([]string{process.Command}, process.Args...)
-	return strings.Join(commandWithArgs, " ")
+	cmdString := process.Command
+	for _, a := range process.Args {
+		cmdString = fmt.Sprintf(`%s %q`, cmdString, a)
```
kieron-dev, line 70: I wonder if this is resilient to all arg content. Is it possible that a valid arg could contain quoting that would become invalid when wrapped in further quotes?

Birdrock: We should expect any valid args would be properly quoted and escaped. It is possible that manifest commands could come in without escaping, but there wouldn't be any args, so the quoting shouldn't come into play.

Regardless, this is more of a band-aid fix to enable buildpacks that provide a `bash -c` style cmd and args. We should avoid string processing of this sort whenever possible. We should have discussion around how to move forward.

davewalter: I think Kieron's concern is that, by wrapping a string with quotes without escaping the string being wrapped, something like `"properly \"escaped\" string"` becomes `""properly \"escaped\" string""`, which is not the same as `"\"properly \\\"escaped\\\" string\""`.

Birdrock: Looking at [fmt](https://pkg.go.dev/fmt) docs, `%q	a single-quoted character literal safely escaped with Go syntax.` I think we should be safe.

davewalter: I just tried this out in the Go playground and it looks to be doing what we want it to: https://go.dev/play/p/UZnII9jxCZe

davewalter (approved):

PR #1443, opened 2022-07-27
Remove test consistently failing on concourse

georgethebeatle:

No

## What is this change about?

The test being removed is able to pass loclally as well as in gh
actions, but fails consistently in our Concourse, because our test are
running as `root` in the ci image.

There seems to be no easy way of fixing this so we remove it to unblock
ci.

Running our tests as non-root is certainly a possibility, but we are not
convinced that it is worth the cost.

No

## Acceptance Steps

N/A

@cloudfoundry/cf-k8s


kieron-dev (approved):
Approving to unblock CI.

We do need to work out how to reintroduce this so it works with the golang image running as root, or change the concourse images to run as non-root (and fix all the permissions / caching throughout)

PR #1454, opened 2022-07-28
Rewrite the API docs

gcapizzi:
- Make the structure consistent
- Remove any information already present in the official docs
- Always link to the official docs
- Use the same order as the official docs
- Add missing endpoints
- Update outdated entries

This is hard to evaluate from the diff, but you can take a look at the [new version of the document](https://github.com/cloudfoundry/korifi/blob/api-docs-cleanup/docs/api.md).

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -393,6 +393,34 @@ This endpoint is fully supported.
 
 This endpoint is fully supported.
 
+## [Tasks](https://v3-apidocs.cloudfoundry.org/#tasks)
+
+### [Create a task](https://v3-apidocs.cloudfoundry.org/#create-a-task)
+
+#### Supported parameters:
+
+-   `command`
+
+### [Get a task](https://v3-apidocs.cloudfoundry.org/#get-a-task)
+
+This endpoint is fully supported.
+
+### [List tasks](https://v3-apidocs.cloudfoundry.org/#list-tasks)
+
+#### Supported query parameters:
+
+No query parameters are supported.
+
+### [List tasks for an app](https://v3-apidocs.cloudfoundry.org/#list-tasks-for-an-app)
+
+#### Supported query parameters:
+
+-   `sequence_ids`
+
+### [Cancel a task](https://v3-apidocs.cloudfoundry.org/#cancel-a-task)
+
+These endpoints are fully supported.
+
 ## User Identity
 
 > **Warning**
--- a/docs/api.md
+++ b/docs/api.md
@@ -1,468 +1,425 @@
-# CF K8s API Endpoint Documentation
+# Korifi API Documentation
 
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -1,468 +1,427 @@
...
+This document lists all the CF API endpoints supported by Korifi and their parameters.
 
-## Resources
+## Cloud Foundry V3 API
 
-### Root
+### [Apps](https://v3-apidocs.cloudfoundry.org/version/3.117.0/#apps)
 
-Docs: https://v3-apidocs.cloudfoundry.org/version/3.117.0/index.html#root
+#### [Create an app](https://v3-apidocs.cloudfoundry.org/version/3.117.0/#create-an-app)
```
tcdowney, line 9: Nothing we can do about it, but a big gripe I have with the V3 API docs is that there is no way to link to the latest release.

docs/api.md
```diff
@@ -1,468 +1,427 @@
...
-        "data": {
-          "guid": "7304bc3c-7010-11ea-8840-48bf6bec2d78"
-        }
-      }
-    }
-  }'
-```
+### [Manifests](https://v3-apidocs.cloudfoundry.org/version/3.117.0/#manifests)
+
+#### [Applying a manifest](https://v3-apidocs.cloudfoundry.org/version/3.117.0/#create-a-route)
```
akrishna90, line 134: the link is incorrect. It need to point to - https://v3-apidocs.cloudfoundry.org/version/3.117.0/#apply-a-manifest-to-a-space

docs/api.md
```diff
@@ -1,468 +1,427 @@
...
+-   `destinations[].protocol`
+
+#### [Remove destination for a route](https://v3-apidocs.cloudfoundry.org/version/3.117.0/#remove-destination-for-a-route)
 
 This endpoint is fully supported.
 
-### User Identity
+### [Service Instances](https://v3-apidocs.cloudfoundry.org/version/3.117.0/#service-instances)
+
+Korifi only supports user-provided service instances. [Fields](https://v3-apidocs.cloudfoundry.org/version/3.117.0/#fields) are not supported.
```
tcdowney, line 312:
```suggestion
Korifi only supports user-provided service instances. Managed service operations and the [Fields](https://v3-apidocs.cloudfoundry.org/version/3.117.0/#fields) parameter are not supported.
```

docs/api.md
gcapizzi: Yeah 😞 I'll still update all links to point to the current latest!

docs/api.md
gcapizzi: Fixed, thanks!

docs/api.md
tcdowney: Wow that's great!

PR #1456, opened 2022-07-29
Simplify login code in smoke tests

georgethebeatle:

No

## What is this change about?

Do the equivalent of echo cf-admin | cf login in the smoke test code

No

## Acceptance Steps

Green tests

@kieron-dev


kieron-dev (approved):
Yep - looks good

PR #1457, opened 2022-07-29
Fix bug reading stats for recently-scaled-down process

matt-royal:

No
## What is this change about?

This PR solves a bug when running `cf scale -i` to a scale an App down. Before this change, the following sequence would consistently lead to an error:

1. `cf scale -i 4`
2. Wait for all instances to come up (by polling `cf app`)
3. `cf scale -i 1`

The user would see output like this:
```
$ cf scale node -i 1
Scaling app node in org o / space s as cf-admin...


Error unmarshalling the following into a cloud controller error: upstream connect error or disconnect/reset before headers. reset reason: connection termination
FAILED
```

No

## Acceptance Steps

Follow the steps in the "What is this change about?" section above. On this branch, there should be no error when scaling down the App.

### DIFF
--- a/api/repositories/pod_repository.go
+++ b/api/repositories/pod_repository.go
@@ -117,6 +117,9 @@ func (r *PodRepo) ListPodStats(ctx context.Context, authInfo authorization.Info,
 		if podState == "DOWN" {
 			continue
 		}
+		if index >= len(records) {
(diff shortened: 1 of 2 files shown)


davewalter (approved):
Looks good!

PR #1466, opened 2022-08-01
Require `staged` condition on app on task reconcile

danail-branekov:
https://github.com/cloudfoundry/korifi/issues/1464

## What is this change about?
The task reconciler now returns an error when the app does not have the
`staged` status condition. The `staged=true` condition indicates that
the app's status has its droplet and `VCAP_SERVICES` secret set.

The task controller also explicitly checks for the droplet reference
being not empty to make sure that the task can transition into
`initialized` state.

Also, reorder objects creation on integartion tests to mitigate
https://github.com/cloudfoundry/korifi/issues/954

No

## Acceptance Steps
The flake is fixed

@kieron-dev


kieron-dev (approved):
lgtm

PR #1500, opened 2022-08-05
Use `Patch` in service instance controller

georgethebeatle:
https://github.com/cloudfoundry/korifi/issues/714

## What is this change about?
* Use `Patch` in service instance controller
* Remove the redundant `CfServiceInstance.Status.Binding` field - it has
  been only referenced in tests
* Improve the controller integration test
* Simplify the controller implementation

No

## Acceptance Steps
No conflict errors caused by updating service instance status by the controller

@danail-branekov

### DIFF
--- a/controllers/api/v1alpha1/cfserviceinstance_types.go
+++ b/controllers/api/v1alpha1/cfserviceinstance_types.go
@@ -17,7 +17,7 @@ limitations under the License.
 package v1alpha1
 
 import (
-	v1 "k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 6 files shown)


controllers/api/v1alpha1/cfserviceinstance_types.go
```diff
@@ -49,10 +48,6 @@ type InstanceType string
 
 // CFServiceInstanceStatus defines the observed state of CFServiceInstance
 type CFServiceInstanceStatus struct {
-	// A reference to the Secret containing the credentials (same as spec.secretName).
-	// This is required to conform to the Kubernetes Service Bindings spec
-	Binding v1.LocalObjectReference `json:"binding"`
```
tcdowney, line 54: This is required to meet the "ProvisionedService" duck type which lets us interop with other tooling in the "K8s Service Binding" ecosystem. Kpack, the Service Binding reconciler that volume mounts service bindings as secrets, etc. all require this when you use them in certain ways.

https://servicebinding.io/spec/core/1.0.0/#provisioned-service

Those projects support working with anything that has a `spec.binding.status` field which is why we have this.

kieron-dev (approved):

PR #1503, opened 2022-08-05
Delete 'VCAP_SERVICES' Secret when App is deleted

tcdowney:
Issue: https://github.com/cloudfoundry/korifi/issues/1498

## What is this change about?
Sets an ownerReference on this Secret to the CFApp that it belongs to so that it gets cleaned up when the app is deleted.

No

## Acceptance Steps
1. `cf push` an app
2. Verify that it has a Secret containing `VCAP_SERVICES`: `kubectl -n $(cf space s --guid) get secrets`
  There will be one with a `-vcap-services` suffix corresponding for the app
3. `cf delete APP_NAME -f`
4. Verify that the Secret is deleted

### DIFF
--- a/controllers/controllers/workloads/cfapp_controller.go
+++ b/controllers/controllers/workloads/cfapp_controller.go
@@ -401,7 +401,7 @@ func (r *CFAppReconciler) createVCAPServicesSecretForApp(ctx context.Context, cf
 	err := r.Client.Get(ctx, vcapServicesSecretLookupKey, vcapServicesSecret)
 	if err != nil {
 		if !apierrors.IsNotFound(err) {
-			r.Log.Error(err, "unable to fetch vcap services Secret")
+			r.Log.Error(err, "unable to fetch 'VCAP_SERVICES' Secret")
 			return err
 		}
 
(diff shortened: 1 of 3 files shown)


acosta11 (approved):
LGTM

The only thing I was thinking about whether is whether this case would be a valuable E2E test. But because we only use the cf api without any inspection of underlying Kubernetes resources, this check would have to live in integration tests only, I think.

tcdowney:
> The only thing I was thinking about whether is whether this case would be a valuable E2E test. But because we only use the cf api without any inspection of underlying Kubernetes resources, this check would have to live in integration tests only, I think.

Yeah, I took a look at the e2e tests initially to add it before it dawned on me that there was no good way to peak behind the curtains and inspect it without `kubectl`. We don't really have any integration tests apart from the smoketests that would be able to do this I think.

I think a test case that proves you can push a running app and delete it and everything is cleaned up is valuable though, just not sure where it should go.

tcdowney:
@acosta11 I made a chore to try and capture some of this sentiment. I think it needs discussed further, but feel free to add some of your ideas there as well:
https://github.com/cloudfoundry/korifi/issues/1505

PR #1517, opened 2022-08-09
Add additional diagrams and content to architecture docs

tcdowney:
## Readable Version
**[Rendered Version](https://github.com/cloudfoundry/korifi/blob/update-arch-docs/docs/architecture.md)**

No, but adding some additional "deeper dives" into specific areas is something folks in the community have asked for.

## What is this change about?
- Added new sections for routing, services, core Korifi, logs/metrics, etc.
- Added additional diagrams and update existing ones

No

## Acceptance Steps
None

TBD

## Misc
- Diagrams available here: https://miro.com/app/board/uXjVOhxvHB8=/?share_link_id=665497294647
  - Can be used to get higher resolution versions of them.


danail-branekov:
I think it would be useful to have a section targeted towards kubectl users so that we make it clear that CF API is not the only way to interract with Korifi. Maybe a list of custom resources one needs to apply to simulate pushing an app would be a nice one.

Otherwise, looks good.

danail-branekov (approved):

gcapizzi (approved):

tcdowney:
> I think it would be useful to have a section targeted towards kubectl users so that we make it clear that CF API is not the only way to interact with Korifi. Maybe a list of custom resources one needs to apply to simulate pushing an app would be a nice one.
> 
> Otherwise, looks good.

That's a good suggestion 👍. I'll add a bit to explain some of that and give some examples. The full list of things you need to do and watch for to push an app using only custom resources might be out of scope for this -- could be a decent blog post or tutorial though -- kind of like a Korifi version of https://www.cloudfoundry.org/blog/build-cf-push-learn-procfiles/.

PR #1535, opened 2022-08-16
Fix kpack builder check for reconcilerName

julian-hj:
related to #1427

## What is this change about?
- Previously, the check to ignore BuildWorkload objects intended for
  other reconcilers was nested in an if block so that it would only
  apply to new objects.
- This resulted in the kpack reconciler failing to ignore objects that
  were partially reconciled by another reconciler.
- In an unrelated change, this commit also makes a bunch of test helper
  methods private, since they were not used anywhere outside the
  package.

No. 

## Acceptance Steps
There's a new envtest that tests the previously failing case. That's probably the easiest way to reproduce this. Otherwise you would have to manually make the BuildWorkload and update the status and make the corresponding Image with all the right fields.

cc @matt-royal @clintyoshimura


davewalter (approved):

clintyoshimura (approved):
LGTM

Issue #1525, opened 2022-08-10
[Chore]: Add AppWorkload.spec.buildRef


matt-royal:
### Background

To support the idea of a combined builder-runner component (as opposed to separate deployments to reconcile BuildWorkload and AppWorkload), we need to provide a way to link the AppWorkload to the CFBuild that generated its image.

### Action to take

- Add the same `spec.buildRef` field to `AppWorkload` that exists on `BuildWorkload`
- Change the CFProcess controller to set this field on `AppWorkload`

### Impact

After this change, alternative build+run implementations will be able to connect a BuildWorkload and AppWorkload together based on their namespace and `spec.buildRef` fields.

gcapizzi:
It sounds wrong to have the lower level resources (`AppWorkload`/`BuildWorkload`) depend on the higher level ones (`CFBuild`): how does this work? Does the `buildRef` have to be a `CFBuild` or is it more of a duck type?

matt-royal:
@gcapizzi `buildRef` is a direct reference to a CFBuild in the same namespace. The goal in adding this was to have a single field to link Build and AppWorkloads from the same "origin" so that they can be connected easily by a combined builder+runner. There may be a better way to do this, though, that doesn't directly link to a higher level resource, like you mentioned. 

When I was first writing this, I had it as a `buildWorkloadRef`, but then I saw that `BuildWorkload` already had `buildRef` so I went with that. `BuildWorkload.metadata.name`. That would also allow us to remove the `buildRef` field on the `BuildWorkload`, since we clearly weren't using it before (it was empty in practice). What do you think?

gcapizzi:
Definitely much better for `AppWorkload` to refer to `BuildWorkload` as at least they're at the same level of abstraction. Question: if it's just correlation we're after, can't we just have some sort of correlation id? Maybe as a label/annotation?

matt-royal:
A correlation ID could work, but I'm not sure I see a benefit to making it more abstract when the link we want to establish is that the AppWorkload is running the output of a specific BuildWorkload.

RE labels and annotations, I've generally avoided using them in a "load bearing" way, since they're not a codified part of a CRD that we can easily validate (short of a validating webhook)

gcapizzi:
The problem I see is that this link is not necessary something runners and builders will want to establish most of the times, as they're usually separate and decoupled implementations - if a specific implementation does, then I'd rather see that solved via metadata.

matt-royal:
[...]

PR #1540, opened 2022-08-18
Revert "Add AppWorkload.spec.buildRef"

matt-royal:
 #1525

## What is this change about?
 Reverting a previous change for the issue, since there was agreement this wasn't a good solution. We don't yet have an alternative, but that will be in a separate issue if/when we decide on one

 Possibly, it removes a field from the AppWorkload resource. It's likely that the new CRD won't be able to apply to an existing cluster.

## Acceptance Steps
 None


tcdowney (approved):
LGTM

Issue #1477, opened 2022-08-02
[Bug]: Race condition fetching logs with `cf start`


julian-hj:
### What happened?

pushing an app with `--no-start` and then running `cf start` on the app results in a 404 error before logs start to stream.

```
$ cf start node2
Starting app node2 in org o / space s as kubernetes-admin...

Staging app and tracing logs...
Failed to retrieve logs from Log Cache: unexpected status code 404

   Build reason(s): CONFIG
   CONFIG:
   	+ env:
   	+ - name: VCAP_SERVICES
...
```

### What you expected to happen

no 404 error

### Acceptance Criteria

GIVEN I have `cf push`ed an app with the `--no-start` option
WHEN I `cf start` the same app
THEN I see logs streaming without an initial 404 error.

### How to reproduce it (as minimally and precisely as possible)

cf p node --no-start
cf start node

### Anything else we need to know?

If this turns out to just be a race condition with the log endpoint getting touched before some part of the app exists, probably the correct fix is to just return 200 with an empty list of envelopes, instead of 404. That aligns with what cf4vms log cache does.

If the user does not have access to view logs for pods they should still get a 403.

### Environment

Revision of codebase:
Kubernetes version (use `kubectl version`):
Cloud provider or hardware configuration:
Others:

julian-hj:
Confirmed fixed.

PR #1548, opened 2022-08-19
Ignore build not found errors when fetching logs

clintyoshimura:
#1477 

## What is this change about?
Fetching logs is returning a 404 error when the build is not found when trying to fetch the build logs. However, this can occur when starting an app depending on the timing. The read app logs function should return only the app logs if the build cannot be found instead. 

no

## Acceptance Steps
See issue

@Birdrock


api/actions/app_logs_test.go
```diff
@@ -209,7 +209,7 @@ var _ = Describe("ReadAppLogs", func() {
 		})
 	})
 
-	When("GetLatestBuildByAppGUIDReturns returns a Forbidden error", func() {
+	When("GetLatestBuildByAppGUID returns a Forbidden error", func() {
```
davewalter, line 212: `GetLatestBuildByAppGUIDReturns` is the name of the function that is returning the error.

api/actions/app_logs.go
```diff
@@ -36,14 +37,17 @@ func (a *AppLogs) Read(ctx context.Context, logger logr.Logger, authInfo authori
...
-	}
-
-	buildLogs, err := a.buildRepo.GetBuildLogs(ctx, authInfo, app.SpaceGUID, build.GUID)
-	if err != nil {
-		return nil, apierrors.LogAndReturn(logger, err, "Failed to fetch build logs", "AppGUID", appGUID, "BuildGUID", build.GUID)
+		if !errors.As(err, new(apierrors.NotFoundError)) {
+			return nil, apierrors.LogAndReturn(logger, apierrors.ForbiddenAsNotFound(err), "Failed to fetch latest app CFBuild from Kubernetes", "AppGUID", appGUID)
+		}
+	} else {
+		buildLogs, err = a.buildRepo.GetBuildLogs(ctx, authInfo, app.SpaceGUID, build.GUID)
```
davewalter, line 47: If there isn't a build for the given app GUID yet, I think that it is safe to assume that there won't be any logs either. Also, the `build.GUID` parameter to the call to `GetBuildLogs` will be empty.

You could probably simply return an empty `repositories.LogRecord` array and a nil error inside the check for an error from `GetLatestBuildByAppGUID` (lines 42-46).

davewalter (changes requested):
See inline comments.

api/actions/app_logs.go
clintyoshimura: The GetBuildLogs command only runs if there is no error from GetLatestBuildByAppGUID (Including not found errors) so it defaults to the var buildLogs which is an empty list that will eventually get merged with the result of GetRuntimeLogsForApp below. If we're sure that there will be never be any app logs from GetRuntimeLogsForApp then we can return an empty list and no err here instead.

api/actions/app_logs_test.go
clintyoshimura: The function name is GetLatestBuildByAppGUID and the mock function to set the return value is GetLatestBuildByAppGUIDReturns right? Which should we use in the When statement?

api/actions/app_logs.go
davewalter: Thanks for the clarification. While that does make sense now, I think it is more Golang-idiomatic to short-circuit and handle the empty list return inside the non-nil error check so that you know that the code that runs after that does have a build to work with. Personally, I find that style easier to read/understand than using else statements.

clintyoshimura: I think I understand. Since there is no build there won't be an app so we can just return an empty list immediately instead of querying the app logs and merging the empty build log list. Fixed

davewalter (approved):

PR #1549, opened 2022-08-19
Use the build-workload-name label to find build logs

matt-royal:
No

## What is this change about?
This PR changes how we fetch build logs, so that we use the `korifi.cloudfoundry.org/build-workload-name` label instead of `image.kpack.io/image`. Both labels have the same value. Without this change, an alternative builder implementation would need to set a kpack-specific label on their build pods to make logs available. As we want to support non-kpack builders, this requirement doesn't make sense.

No

## Acceptance Steps
Tests passing should be sufficient. There is no user-visible change


clintyoshimura (approved):
LGTM

Issue #1510, opened 2022-08-08
[Feature]: Propagate service accounts to Spaces based on annotation


matt-royal:
### Background

The CFSpace controller in `korifi` currently creates only the `kpack-service-account` in each Space's namespace. This chore switches to a more flexible system where any service account in the root namespace will be copied into all space namespaces if it has the correct annotation.

### Action to take

* Remove the logic in the CFSpace controller that creates a `korifi-service-account` service account in each space's namespace
* Add new code to the CFSpace controller that finds any service accounts in the root (e.g. `cf`) namespace that has the annotation `cloudfoundry.org/propagate-service-account: true` and ensures that it also exists in the namespace for every Space. Feel free to change the annotation if you have a better option, but it should remain opt-in.
* Add a `kpack-service-account` in the `cf` namespace with this new annotation set. This will maintain the current behavior from before this chore was executed.

### Impact

This chore will allow us to create more service accounts, each with a minimal set of RBAC capabilities. Currently our only option is to jam more capabilities on the kpack-service-account or hard code additional service accounts into the CFSpace controller.

PR #1586, opened 2022-08-30
CFSpace controller propagates service accounts

julian-hj:
[#1510]
## What is this change about?
- previously we had hard-coded logic to create a `kpack-service-account`
  in each space.
- this commit looks for service accounts in the root namespace with a
  specific annotation, and copies them instead.
- the contoller manager role is extended to watch/patch/delete service accounts in order to keep spaces
  in sync with any changes in the root namespace.

no

## Acceptance Steps
[#1510]

@matt-royal

### DIFF
--- a/controllers/controllers/workloads/cfspace_controller.go
+++ b/controllers/controllers/workloads/cfspace_controller.go
@@ -21,26 +21,25 @@ import (
 	"fmt"
 	"time"
 
-	rbacv1 "k8s.io/api/rbac/v1"
-	k8serrors "k8s.io/apimachinery/pkg/api/errors"
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
-	"sigs.k8s.io/controller-runtime/pkg/handler"
-	"sigs.k8s.io/controller-runtime/pkg/reconcile"
-	"sigs.k8s.io/controller-runtime/pkg/source"
-
 	korifiv1alpha1 "code.cloudfoundry.org/korifi/controllers/api/v1alpha1"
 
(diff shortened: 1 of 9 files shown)


controllers/controllers/workloads/cfspace_controller.go
```diff
@@ -150,34 +153,78 @@ func (r *CFSpaceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
+	)
+
+	serviceAccounts := new(corev1.ServiceAccountList)
+	err = r.client.List(ctx, serviceAccounts, client.InNamespace(r.rootNamespace))
+	if err != nil {
+		r.log.Error(err, fmt.Sprintf("Error listing service accounts from namespace %s", space.GetNamespace()))
+		return err
+	}
+
+	parentRoleBindingMap := make(map[string]struct{})
```
danail-branekov, line 169: This map is actually a set of reconciled service account names and has nothing to do with role bindings. I would therefore suggest renaming it to `reconciledServiceAccounts` to make that clear.

danail-branekov (dismissed):
This change is lacking the unit test to verify errors are returned upon failure to list service accounts or role bindings. I also have a small comment on variable naming.

Other than that, looks all right to me.

controllers/controllers/workloads/cfspace_controller.go
julian-hj: Haha. Can you tell that this was copy/pasted? Thanks.

clintyoshimura (approved):
LGTM

PR #1608, opened 2022-09-01
Fix kind load targets to specify cluster name

georgethebeatle:
No

## What is this change about?
Fix kind load targets to specify cluster name

Otherwise `kind load docker-image` is trying to use a cluster named `kind`
which does not exist in the context of `make test-e2e`

No

## Acceptance Steps
`make test-e2e` succeeds when there is no kind cluster

@danail-branekov

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@
 IMG_CONTROLLERS ?= cloudfoundry/korifi-controllers:latest
 IMG_API ?= cloudfoundry/korifi-api:latest
 CRD_OPTIONS ?= "crd"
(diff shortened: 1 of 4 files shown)


gcapizzi:
Given this used to work previously, is it a breaking change in kind? If yes, do we need to make sure that everyone has a particular version installed?

danail-branekov:
> Given this used to work previously, is it a breaking change in kind?

It's complicated :)
see https://cloudfoundry.slack.com/archives/C0297673ASK/p1662048099497789?thread_ts=1662047778.029739&cid=C0297673ASK

PR #1616, opened 2022-09-02
Do not inject cert-mgr ca into CRDs

kieron-dev:
No

## What is this change about?
Turn off the kustomization path to insert cert-manager cainjection annotations in some CRDs.

These are used for CRD version conversion webhooks, but we don't use those.

No

## Acceptance Steps
All continues to work


tcdowney:
> These are used for CRD version conversion webhooks, but we don't use those.

There might be a time when we do want to use them to convert CRs to a new version, but since this is just an annotation it seems easy enough to add this back if/when we do support conversion webhooks.

georgethebeatle (approved):

PR #1618, opened 2022-09-03
Fix the deploy-on-kind script

matt-royal:
 No

## What is this change about?
 Fixes a bug introduced in commit 42e44ce4, which caused `scripts/deploy-on-kind.sh` to fail consistently unless the name `e2e` was used.

Also increases the CPU limit on the kpack-controller so that ClusterBuilder becomes ready quickly instead of the 10+ minutes it's been taking on some machines

No

## Acceptance Steps
 
`make test-e2e` passes
`./scripts/deploy-on-kind.sh` succeeds with a name besides e2e


clintyoshimura (approved):
LGTM

Issue #1171, opened 2022-06-02
[Feature]: Specifying any buildpack other than default errors


matt-royal:
### Blockers/Dependencies

#1054 

### Background

**As a** cf CLI user
**I want** to receive an error message when I select a buildpack other than default
**So that** I have understand that I cannot select a specific buildpack (right now we silently ignore the buildpack request)

### Acceptance Criteria

**GIVEN** a korifi cluster
**WHEN I** run `cf push -b paketo-buildpacks/go`
**THEN I** should see that the build fails with an error explaining that only the default buildpack is currently supported

---

**WHEN I** run `cf push`, `cf push -b default`, or `cf push -b null`
**THEN I** should see that the build succeeds
**AND** that `cf app <AppName>` shows no buildpacks (i.e default)

### Dev Notes

This story spans `korifi/api`, `korifi/controllers`, and `korifi/kpack-image-builder`.

- In `korifi/api`:
  - The API, payload and BuildRepository must respect the user-specified buildpack
  - When no buildpack is specified, the buildpacks should be left empty on `CFBuild`
  
- In `korifi/controllers`:
  - The `BuildWorkload` CRD now has a `spec.buildpacks` field
  - The `CFBuildReconciler` sets `BuildWorkload.spec.buildpacks` based on the value in `CFBuild.spec.lifecycle.data.buildpacks`
  
- In `korifi/kpack-image-builder`:
  - If the `BuildWorkload.spec.buildpacks` is any value other than blank, then the `Succeeded` status condition becomes false with a message that explains that only the default buildpack is currently supported.
  
When a buildpack is specified to `cf push`, the manifest will look like this...
```
applications:
- name: my-app
  buildpacks:
  - paketo-buildpacks/go
```

When no buildpack is specified, no `buildpacks` key will be set on the manifest.

When the `default` or `null` buildpack is specified, the `buildpacks: []` will be in the manifest

matt-royal:
This is an interview story. We may not merge the work or choose to take a different path in the future.

PR #1619, opened 2022-09-05
Error when buildpack is specified during first push

matt-royal:
 #1171 

## What is this change about?
 See issue

 No

## Acceptance Steps
 See issue

### DIFF
--- a/kpack-image-builder/controllers/buildworkload_controller.go
+++ b/kpack-image-builder/controllers/buildworkload_controller.go
@@ -141,7 +141,7 @@ func (r *BuildWorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 
 	if len(buildWorkload.Spec.Buildpacks) > 0 {
 		// Specifying buildpacks is not supported
-		failureStatusConditionMessage := `Only the "default" buildpack is supported`
+		failureStatusConditionMessage := `Only buildpack auto-detection is supported. Specifying buildpacks is not allowed.`
 		setSucceededConditionOnLocalCopy(&buildWorkload.Status.Conditions, metav1.ConditionFalse, "InvalidBuildpacks", failureStatusConditionMessage)
 		err = r.Client.Status().Update(ctx, buildWorkload)
 		if err != nil {
--- a/kpack-image-builder/controllers/buildworkload_controller.go
+++ b/kpack-image-builder/controllers/buildworkload_controller.go
@@ -139,6 +139,17 @@ func (r *BuildWorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		return ctrl.Result{}, nil
 	}
 
+	if len(buildWorkload.Spec.Buildpacks) > 0 {
+		// Specifying buildpacks is not supported
+		failureStatusConditionMessage := `Only the "default" buildpack is supported`
+		setSucceededConditionOnLocalCopy(&buildWorkload.Status.Conditions, metav1.ConditionFalse, "InvalidBuildpacks", failureStatusConditionMessage)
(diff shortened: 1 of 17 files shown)


kpack-image-builder/controllers/buildworkload_controller_test.go
```diff
@@ -20,7 +20,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
-var _ = Describe("BuildWorkloadReconciler", func() {
+var _ = Describe("BuildWorkloadReconciler", Focus, func() {
```
tcdowney, line 23: Looks like a stray `Focus` here is what's causing the tests to fail

kpack-image-builder/controllers/buildworkload_controller.go
```diff
@@ -139,6 +139,17 @@ func (r *BuildWorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 		return ctrl.Result{}, nil
 	}
 
+	if len(buildWorkload.Spec.Buildpacks) > 0 {
+		// Specifying buildpacks is not supported
+		failureStatusConditionMessage := `Only the "default" buildpack is supported`
```
tcdowney, line 144: I think this error message might be confusing. There's not really a concept of a "default" buildpack, right? Like there is a default builder that has a collection of buildpacks. Maybe that could be mentioned?

That's not really a CF concept though so if this error is bubbled up and displayed through `cf push`, maybe an error message that mentions only "buildpack auto-detection" is supported?

IDK, open to thoughts on this one, but I think something that mentions auto-detection and you can't specify individual buildpacks would be good.

tcdowney (changes requested):

kpack-image-builder/controllers/buildworkload_controller.go
tcdowney: I guess `'default'` and `'null'` are both valid CF App Manifest concepts to empty out the buildpacks array on the Cloud Controller side. 🤔 

https://github.com/cloudfoundry/cloud_controller_ng/commit/f402ddde20703b17e805b317fcaec6a25abcd5c7

matt-royal: I missed that when I was writing this story. The `cf push` command has built-in logic to turn `default` into an empty buildpacks array, so I assumed we didn't need this. Do you think we should add this? If so, I think that the CFProcess reconciler is the right place to handle turning `default` into a no-op

matt-royal: Actually, we could also do the `default` translation in the App repo, the payload or the handler. As long as we don't make every builder reimplement this logic.

matt-royal: I took a second look at the CCNG link above. The logic to handle `default` is only invoked when the Manifest has a `buildpack` key (not `buildpacks`). This commit doesn't support that field at all, so the logic wouldn't fit anyway. If we're concerned about supporting both `buildpack` and `buildpacks`, then I can split out a separate story to handle the singular version with the `default` logic included.

tcdowney (approved):
LGTM

kpack-image-builder/controllers/buildworkload_controller.go
tcdowney: Re: supporting the `buildpack` field in addition to `buildpacks`.

It is probably worth making a story to capture this context in case it comes up, but I don't think it's worth supporting at this time.  If we ever do it then we can handle `default`/`null`, but those were both legacy manifest values when we implemented them in CCNG in the first-place.

PR #1634, opened 2022-09-08
Improve the reason and message on BuildWorkload conditions

matt-royal:
 No

## What is this change about?
 Improve the reason and message fields on the BuildWorkload `Succeeded` condition. These are eventually shown to the user upon failure, and the old messages were not clear or helpful.

 Only if clients were depending on the old `reason` values we were using, which is unlikely

## Acceptance Steps
 `cf push` an app that you know will fail to stage and observe that the error message is now clearer.

### DIFF
--- a/kpack-image-builder/controllers/buildworkload_controller.go
+++ b/kpack-image-builder/controllers/buildworkload_controller.go
@@ -176,14 +176,13 @@ func (r *BuildWorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 
 	kpackReadyStatusCondition := kpackImage.Status.GetCondition(kpackReadyConditionType)
 	if kpackReadyStatusCondition.IsFalse() {
-		failureStatusConditionMessage := kpackReadyStatusCondition.Reason + ":" + kpackReadyStatusCondition.Message
-		setSucceededConditionOnLocalCopy(&buildWorkload.Status.Conditions, metav1.ConditionFalse, "kpack", failureStatusConditionMessage)
+		setSucceededConditionOnLocalCopy(&buildWorkload.Status.Conditions, metav1.ConditionFalse, "BuildFailed", "Check build log output")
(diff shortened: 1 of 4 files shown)


kpack-image-builder/controllers/buildworkload_controller.go
```diff
@@ -176,14 +176,13 @@ func (r *BuildWorkloadReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 
 	kpackReadyStatusCondition := kpackImage.Status.GetCondition(kpackReadyConditionType)
 	if kpackReadyStatusCondition.IsFalse() {
-		failureStatusConditionMessage := kpackReadyStatusCondition.Reason + ":" + kpackReadyStatusCondition.Message
-		setSucceededConditionOnLocalCopy(&buildWorkload.Status.Conditions, metav1.ConditionFalse, "kpack", failureStatusConditionMessage)
+		setSucceededConditionOnLocalCopy(&buildWorkload.Status.Conditions, metav1.ConditionFalse, "BuildFailed", "Check log output")
```
tcdowney, line 179: Is there anything more specific we can provide than `"Check log output"` here? Or is this due to us not having good visibility into the status of the backing "builder" implementation?

matt-royal: When I was making this change I tried pushing an app in an unsupported language (php) to see it fail. Unfortunately, there was no useful information on the kpack Image or kpack Build. Their conditions don't include reason or message and I couldn't find anything else that looked like it would be helpful.

tcdowney: Makes sense. Maybe tweaking it to say "Check build log output" will make it slightly more clear what logs we're pointing people at.

matt-royal: I pushed an updated message

tcdowney (approved):
LGTM

Issue #1179, opened 2022-06-03
[Feature]: Operator can get consistent logs from korifi-api-deployment


julian-hj:
### Background

**As a** Operator
**I want** predictable and consistent log format
**So that** I can use tools to to search and organize logs

### Acceptance Criteria

**GIVEN** a Korifi deployment
**WHEN I** `kubectl logs -n korifi-api-system korifi-api-deployment-xxxxxxxxxx-xxxxx`
**THEN I** see that the results are consistently formatted in json following [this proposal](https://docs.google.com/document/d/1FUZaMyYTL-vzoGzODS_V-YKapa6jlivsxc7DjMWxaDg/edit#)
**AND I** see that only system errors are tagged with `"level": "error"` with user errors tagged as `"level":"info"`

### Dev Notes

See [implementation notes in the proposal for suggestions](https://docs.google.com/document/d/1FUZaMyYTL-vzoGzODS_V-YKapa6jlivsxc7DjMWxaDg/edit#heading=h.lwm58dx8x7kx)

Some logging appears to come from the k8s client code, e.g. 
```
I0603 20:05:55.339392       1 request.go:601] Waited for 1.040934921s due to client-side throttling, not priority and fairness, request: GET:https://10.96.0.1:443/apis/acme.cert-manager.io/v1?timeout=32s
```
This logging is in a different format from the other log lines from the same executable. We may have some work to do in order to make sure we have loggers consistently plumbed through the layers in our code.

See [this comment from Kieron](https://github.com/cloudfoundry/korifi/pull/1647#discussion_r971010001) regarding log propagation. We might be able to shoehorn our handler logger into the `ctx` to get it plumbed down into the repo layer.

We could test this by capturing logs during e2e tests and having a test that validates the logs are in the correct format. @julian-hj may be writing a separate chore for this to do it after all these stories are done.

julian-hj:
A/C passed

julian-hj:
It seems that we still have some low-level k8s client logging from klog that wasn't using the standard logger.

To test this: modify the ingress cert so that it rolls over every 3 minutes and redeploy:

```
apiVersion: cert-manager.io/v1
kind: Certificate
metadata:
  name: korifi-api-internal-cert
  namespace: {{ .Release.Namespace }}
spec:
  duration: 90m
  renewBefore: 87m
...
```

Then do something with the cf cli repeatedly. After a few minutes, you'll start to see lines in the log like:

```
[...]

PR #1647, opened 2022-09-14
Never fail when fetching Pod metrics fails

gcapizzi:

No.

## What is this change about?

This makes `PodRepo.ListPodStats` always `continue` when its `metricsFetcher` errors.

Yes: the `/v3/processes/{guid}/stats` endpoint won't fail anymore when fetching metrics fails for any reason.

### DIFF
--- a/api/repositories/pod_repository.go
+++ b/api/repositories/pod_repository.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"io"
 	"strconv"
-	"strings"
 	"time"
 
 	"code.cloudfoundry.org/korifi/api/apierrors"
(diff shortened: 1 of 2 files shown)


georgethebeatle (approved):

api/repositories/pod_repository.go
```diff
@@ -125,12 +124,7 @@ func (r *PodRepo) ListPodStats(ctx context.Context, authInfo authorization.Info,
 
 		podMetrics, err := r.metricsFetcher(ctx, p.Namespace, p.Name)
 		if err != nil {
-			errorMsg := err.Error()
-			if strings.Contains(errorMsg, "not found") ||
-				strings.Contains(errorMsg, "the server could not find the requested resource") {
-				continue
-			}
-			return nil, err
+			continue
```
kieron-dev, line 127: Not related to this change, but there is no logging of the failure here. As that currently only happens in the api layer and we don't pass an error down to it here, we won't have a chance to log it.

kieron-dev (approved):

api/repositories/pod_repository.go
tcdowney: There's no reason why we shouldn't log in the repositories right? I think it might be good to pass the logger down so that when someone asks "why are my metrics always missing?" they can find some logs to answer that question.

kieron-dev: Yep - absolutely.

I think we have to find a good option for passing the loggers through the code. We inject one containing the correct correlation ID in the handlers. It would be good to pass that around, preferably without changing every method in the system to accept a log argument. Piggy-backing on context is an option built into logr.

julian-hj: I think we will need to do something to propagate loggers in order to finish #1179. @kieron-dev I will drop a link to your comment into that story as it sounds like it would be worth trying.

PR #1648, opened 2022-09-14
Use `authn.Keychain` from `go-containerregistry`

gcapizzi:

No.

## What is this change about?

This replaces the `authn.Keychain` implementation coming from Kpack with the one coming from `go-containerregistry`. This reduces our coupling with Kpack: we're now only using their resource types.

No.


tcdowney (approved):
LGTM

Surprised how much of a drop-in replacement that was 😁

Issue #1669, opened 2022-09-19
[Chore]: Make server timeouts configurable with defaults


Birdrock:
### Background

We currently are using the http package convenience methods for starting the API server. These assume an unlimited timeout. We should set defaults and make timeouts configurable.

### Impact

Closes a medium denial of service CVE.

### Dev Notes

Configuring timeouts for http server as noted in https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/

PR #1670, opened 2022-09-19
Make server timeouts configurable

Birdrock:
#1669 

## What is this change about?
Make server timeouts configurable and add defaults.

No.

## Acceptance Steps
Run tests as normal. Optionally try to replicate slow network with large files

@tcdowney

### DIFF
--- a/api/config/base/apiconfig/korifi_api_config.yaml
+++ b/api/config/base/apiconfig/korifi_api_config.yaml
@@ -1,5 +1,9 @@
 externalFQDN: "api.example.org"
 internalPort: 9000
+idleTimeout: 900
+readTimeout: 900
+readHeaderTimeout: 10
+writeTimeout: 900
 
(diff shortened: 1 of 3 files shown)


davewalter (approved):

tcdowney (approved):
LGTM

Just adding on that these 15 minute timeouts match the defaults in CF for VMs:
* https://github.com/cloudfoundry/capi-release/blob/5edae9a50f8dce3ff2401815a505df40b3347423/jobs/cloud_controller_ng/spec#L297
* https://github.com/cloudfoundry/routing-release/blob/3e607d9bfecabb3826624107e2ec50e0b39c3e8a/jobs/gorouter/spec#L358

Why is it so long? Uploading application source code during `cf push` can take a long time and in the real world with real (non-performant) networks often takes upwards of ten minutes.

Issue #1181, opened 2022-06-03
[Feature]: Operator can view basic performance metrics for Korifi controllers via Prometheus


julian-hj:
### Background

**As a** Korifi Operator
**I want** to scrape metrics from the Korifi controllers
**So that** I can monitor the performance of my system

For the scope of this story/epic, we will not add additional metrics to the controller-manager, or add a metrics endpoint to the API shim. We will wait to hear from consumers if such steps are necessary.

### Acceptance Criteria

**GIVEN** a running Korifi installation
**WHEN** I [install Prometheus on my cluster](https://artifacthub.io/packages/helm/prometheus-community/prometheus)
**AND I** query `controller_runtime_active_workers` from the Prometheus server web UI
**THEN I** can see statistics for Korifi controllers have been scraped
**AND I** can see that statistics for the `kpack-image-builder` controller manager have been scraped
**AND I** can see that statistics for the `statefulset-runner` controller manager have been scraped
**AND I** can see that statistics for the `job-task-runner` controller manager have been scraped

![Prometheus UI](https://user-images.githubusercontent.com/8484092/171951332-a4b2e564-2163-47e0-b00f-d7757b120111.png)

### Dev Notes

It appears that our controller manager is already serving metrics on the default port and path:

```
 root@korifi-controllers-controller-manager-85f564c6f6-jzhr6:/# curl localhost:8080/metrics
 # HELP controller_runtime_active_workers Number of currently used workers per controller
 # TYPE controller_runtime_active_workers gauge
 controller_runtime_active_workers \{controller="cfapp"} 0
 controller_runtime_active_workers\{controller="cfbuild"} 0
```

It seems likely that we only need to add the [correct annotations](https://artifacthub.io/packages/helm/prometheus-community/prometheus#scraping-pod-metrics-via-annotations) on our pods so that Prometheus can find them. From observation, it appears that only the prometheus.io/scrape annotation is fully necessary since the defaults for path and port are already correct.

Alternately, the controller runtime creates prometheus monitor configuration that could be enabled by uncommenting the [prometheus overlays in the default configuration](https://github.com/cloudfoundry/korifi/blob/main/controllers/config/default/kustomization.yaml#L25-L26). The downside to that approach however is that it uses a CRD which may not exist on non-prometheus clusters, so it might fail to install in that case. (This should be very easy to test.)

julian-hj:
A/C passed

PR #1671, opened 2022-09-20
Expose /metrics endpoint via HTTP

julian-hj:

#1181

## What is this change about?
- Add prometheus annotations on controller manager pods to advertise metrics on `:8080/metrics`
- Remove kube-rbac-proxy and associated overlays
- Remove vestigial prometheus overlays that use CRs to advertise metrics on port 8443
- This aligns with the metrics endpoints currently exposed by k8s system components.

No.

## Acceptance Steps
#1181

cc @tcdowney


controllers/config/manager/manager.yaml
```diff
@@ -23,6 +23,9 @@ spec:
     metadata:
       annotations:
         kubectl.kubernetes.io/default-container: manager
+        prometheus.io/scrape: "true"
+        prometheus.io/path: /metrics
+        prometheus.io/port: "8080"
```
tcdowney, line 28: Do we need the `prometheus.io/scheme: http` annotation, or is that the default?

tcdowney (approved):
LGTM

danail-branekov (approved):

controllers/config/manager/manager.yaml
julian-hj: I believe it is the default. But then again, the path is also the default, so having one and not the other is a little inconsistent.

Issue #1672, opened 2022-09-20
[Bug]: Scaling down processes to zero does not work


danail-branekov:
### What happened?

1. Deploy an app with at least one web process instance running:

```
❯ cf app node11
Showing health and status for app node11 in org org1 / space space1 as cf-admin...

name:              node11
requested state:   started
routes:            node11.vcap.me
last uploaded:     Tue 20 Sep 12:43:01 EEST 2022
stack:             io.buildpacks.stacks.bionic
buildpacks:

type:           web
sidecars:
instances:      1/1
memory usage:   1024M
     state     since                  cpu    memory   disk     logging      details
#0   running   2022-09-20T09:49:22Z   0.0%   0 of 0   0 of 0   0/s of 0/s

```

2. Scale the app down via applying the following manifest:

```
❯ cat manifest.yml
---
applications:
- name: node11
  processes:
  - type: web
    instances: 0

❯ cf apply-manifest -f manifest.yml
```

3. The number of running web processes has not changed (i.e. is still 1)

```
❯ cf app node11
Showing health and status for app node11 in org org1 / space space1 as cf-admin...

name:              node11
requested state:   started
routes:            node11.vcap.me
last uploaded:     Tue 20 Sep 12:43:01 EEST 2022
stack:             io.buildpacks.stacks.bionic
buildpacks:

type:           web
sidecars:
instances:      1/1
memory usage:   1024M
     state     since                  cpu    memory      disk     logging      details
#0   running   2022-09-20T09:52:00Z   0.0%   9.8M of 0   0 of 0   0/s of 0/s

```

### What you expected to happen

* Apply the manifest, see the web process has been scaled down to zero instances
* 

### Acceptance Criteria

GIVEN I have an app with at least one running web process instance
WHEN I apply a manifest that scales the web process instances down to 0
THEN I see that there are no web process instances running

### How to reproduce it (as minimally and precisely as possible)

See above

### Anything else we need to know?

Scaling down the web process to zero should be also possible for kubectl users via using something like

```
kubectl patch cfprocess cf-proc-$(cf app --guid node11)-web -n $(cf space --guid s) --type merge --patch '{"spec": {"desiredInstances": 0}}'
cf curl /v3/apps/$(cf app --guid node)/processes | jq -e '.resources[0].instances == 0'
```

### Environment

Revision of codebase: ba356b516cb12a170321549741b3530b4edc80ec

PR #1673, opened 2022-09-20
Allow scaling web processes down to 0

danail-branekov:

#1672

## What is this change about?

As zero is a valid value for `CFProcess.Spec.DesiredInstances` (set to
zero when the user wants to scale down process instances to 0) we cannot
rely on the field zero value. That is why the field has its type changed
to `*int` so that the CFProcess webhook would only default the instances
value if the pointer value is not set.

No

## Acceptance Steps
See story


matt-royal (approved):
Looks good to me. Thanks for the fix!

Issue #1552, opened 2022-08-22
[Bug]: Install instructions don't work on kind


gcapizzi:
### What happened?

Trying to follow the [install instructions](https://github.com/cloudfoundry/korifi/blob/main/INSTALL.md) on kind won't work. In particular, [ingress](https://kind.sigs.k8s.io/docs/user/ingress/) needs to be configured properly. Also, we should instruct people on how to properly use a local container registry.

### What you expected to happen

I expect to see instructions in our `INSTALL.md` specifying the right thing to do to install Korifi on kind. We don't want to advertise `deploy-on-kind.sh` (at least not yet), but provide reasonable instructions.

### Acceptance Criteria

GIVEN I have kind installed on my machine
WHEN I follow the instructions in `INSTALL.md` to install Korifi
THEN I see it working

### How to reproduce it (as minimally and precisely as possible)

Try to install Korifi on kind following `INSTALL.md`.

### Anything else we need to know?

[Slack thread](https://cloudfoundry.slack.com/archives/C0297673ASK/p1661178931841559) reporting the issue.

georgethebeatle:
It was discussed that this bug should wait for the distribution work which is planned for next release, because it will greatly influence what the readme looks like.

PR #1675, opened 2022-09-20
More `INSTALL*.md` fixes

gcapizzi:

#1552 

## What is this change about?

This further improves our install instructions. In particular, it:
* adds more details on how to configure Kpack, as the tutorial we used to refer to is not very good for the Korifi use case;
* rearranges information so that it's closer to where it's needed;
* bumps `cf` to `v8.5` so that we can get all our fixes + `cf auth`;
* simplifies the creation of the necessary TLS certificates by leveraging cert-manager.

No. These instructions worked successfully on my local Kind cluster.

## Acceptance Steps

Install Korifi using the new instructions:
* [`INSTALL.md`](https://github.com/cloudfoundry/korifi/blob/install-md-fixes/INSTALL.md)
* [`INSTALL_kind.md`](https://github.com/cloudfoundry/korifi/blob/install-md-fixes/INSTALL_kind.md)

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -123,9 +123,9 @@ EOF
 
 ## Dependencies
 
-### Cert-Manager
+### cert-Manager
 
-[Cert-Manager](https://cert-manager.io) allows us to automatically create internal certificates within the cluster. Follow the [instructions](https://cert-manager.io/docs/installation/) to install the latest version.
+[cert-Manager](https://cert-manager.io) allows us to automatically create internal certificates within the cluster. Follow the [instructions](https://cert-manager.io/docs/installation/) to install the latest version.
 
 ### Kpack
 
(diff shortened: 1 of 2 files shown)


INSTALL.md
```diff
@@ -79,39 +84,16 @@ You will need to update the following configuration files:
         # ...
     ```
 
-### Image Registry Configuration
```
tcdowney, line 82: I agree with removing this in the streamlined initial install instructions. It is pretty verbose.

That said, I think there is a lot of nuance here that we should keep around in some form -- maybe in a separate `registry-configuration.md` doc. 🤔 

Like `DockerHub` has some pretty significant deviations from other OCI registries, but the others have minor differences as well between each other (ECR requires different auth for example)  and it would be good to have a place to document all of that.

tcdowney: We already have `using-container-registry-signed-by-custom-ca.md` and that could probably consolidated into a more general registry config doc as well.

INSTALL.md
```diff
@@ -137,11 +118,22 @@ subjects:
...
     name: $ADMIN_USERNAME
+EOF
 ```
 
-### Container registry credentials
+## Dependencies
+
+### Cert-Manager
+
+[Cert-Manager](https://cert-manager.io) allows us to create internal automatically certificates within the cluster. Follow the [instructions](https://cert-manager.io/docs/installation/) to install the latest version.
```
tcdowney, line 128:
```suggestion
[Cert-Manager](https://cert-manager.io) allows us to automatically create internal certificates within the cluster. Follow the [instructions](https://cert-manager.io/docs/installation/) to install the latest version.
```

INSTALL.md
```diff
@@ -137,11 +118,22 @@ subjects:
...
 
-Use the following command to create a `Secret` that Korifi will use to connect to your container registry:
+### Kpack
+
+[Kpack](https://github.com/pivotal/kpack) powers our source-to-image build process. Follow the [instructions](https://github.com/pivotal/kpack/blob/main/docs/install.md) to install the latest version.
```
tcdowney, line 132: I know you just moved this line up, but after re-reading it I think it might be confusing to new users. Maybe something like this would be more clear:

```suggestion
[Kpack](https://github.com/pivotal/kpack) is used to build runnable applications from source code using [Cloud Native Buildpacks](https://buildpacks.io/). Follow the [instructions](https://github.com/pivotal/kpack/blob/main/docs/install.md) to install the latest version.
```

tcdowney (changes requested):
Overall LGTM, I think the kpack installation section is definitely clearer now around what is required. 👍 

Gave a couple minor wording suggestions and am also wondering if we should have a separate docs story to pull out some of the container registry config into its own docs. A place where we could cover the nuances between registries deeper and consolidate it with the existing `using-container-registry-signed-by-custom-ca.md` doc.

INSTALL.md
gcapizzi: If we feel like we need to document the different nuances, it makes sense to have a separate document. Alternatively we could have a section at the bottom? Something like _Notes on container registries_?

tcdowney (approved):

PR #1677, opened 2022-09-20
Add test for scaling a process to 0

matt-royal:
This behavior was broken before commit c8f84a, but now works as expected. This test will prevent any reversion

 This is related to #1672 , but it's not the exact same issue. The PR for that issue did resolve this problem, though.

## What is this change about?
 This PR adds a test to ensure that the ProcessRepo is able to scale processes to 0 instances. Before #1673 this behavior was broken.

 No

## Acceptance Steps
 Push an app, then use `cf scale -i 0` to scale it to 0. Confirm that the app was actually scaled down.

 @danail-branekov

### DIFF
--- a/api/repositories/process_repository_test.go
+++ b/api/repositories/process_repository_test.go
@@ -303,6 +303,21 @@ var _ = Describe("ProcessRepo", func() {
 				Expect(updatedCFProcess.Spec.MemoryMB).To(Equal(memoryScaleMB))
 			})
 
+			When("scaling down a process to 0 instances", func() {
+				It("works", func() {
+					scaleProcessMessage.ProcessScaleValues = repositories.ProcessScaleValues{Instances: pointerTo(0)}
+					scaleProcessRecord, scaleProcessErr := processRepo.ScaleProcess(context.Background(), authInfo, *scaleProcessMessage)
+					Expect(scaleProcessErr).ToNot(HaveOccurred())
+
+					Expect(scaleProcessRecord.DesiredInstances).To(Equal(0))
+
+					var updatedCFProcess korifiv1alpha1.CFProcess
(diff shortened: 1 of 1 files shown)


tcdowney (approved):
LGTM

acosta11 (approved):
LGTM

julian-hj (approved):

davewalter (approved):

danail-branekov:
I am not convinced that this test prevents the regression. Repositories suite does not install controllers and their webhooks (where the problem actually was).

I also find it a bit weird to have this test against a repository as repositories have no responsibilities to default values.

Having said that, I am pretty sure that this test is not going to fail without the previous fix. I think that an e2e test is the right tool to prevent this regression, but maybe the integration test against the process modification webhook from the fix PR is good enough.

matt-royal:
@danail-branekov This new test fails on the commit before you changed the `DesiredInstances` field to a pointer. The specific issue in the repo is that it isn't fetching the CFProcess resource before it patches it. When the `DesiredInstances` field was an int, the zero-value was 0, so it didn't see setting that field to `0` as a change. Now that it's a pointer, it sees it as a change.

I agree that we would also benefit from also having an e2e test for this behavior, since that would exercise handler, repo, webhooks and controllers all at once.

acosta11:
> I am not convinced that this test prevents the regression. Repositories suite does not install controllers and their webhooks (where the problem actually was).
> I also find it a bit weird to have this test against a repository as repositories have no responsibilities to default values.

A fair point. Behavior that depends on a webhook or some arbitrary controller would need an e2e test to exercise the full flow, which I would support adding.

I'm not sure where the line is for testing to drive out particular error/edge case handling code, but I tend to see unit testing the repositories error handling as a way to capture the interface at a relevant api boundary. So in this case I'm in favor of the unit test even if it feels contrived in the greater context of the bug. But I could be convinced to either side, to be fair.

Thanks for the follow up discussion on the bug!

danail-branekov:
Ah, now I see, yes, having the test now makes sense. Thanks!

Issue #1701, opened 2022-09-23
[Bug]: Health checks don't work as expected


gcapizzi:
### What happened?

Pushed an app that took 5 seconds to respond to any request using an `health-check-invocation-timeout` of `6`, and it failed.

### What you expected to happen

I expected my push to succeed as the time took to respond was inferior to the invocation timeout.

After inspecting [the Korifi code](https://github.com/cloudfoundry/korifi/blob/5ef219cfde47d954802c3555dc8dbf2b1a90361e/controllers/controllers/workloads/cfprocess_controller.go#L345-L371), I discovered that:

* we are ignoring the value of `health-check-invocation-timeout`;
* we are using the value of `timeout` to set `initialDelaySeconds` on a [liveness probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-a-liveness-command);
* we are setting the liveness probe `failureThreshold` to `4`;
* we are not setting the `periodSeconds`;
* we are also creating a [readiness probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-readiness-probes) with the same handler, no delay and a `failureThreshold` of `1`;
* we are not setting any [startup probe](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes).

None of this matches the [expected behaviour of Cloud Foundry](https://docs.cloudfoundry.org/devguide/deploy-apps/healthchecks.html#understand-healthchecks):

* Cloud Foundry will wait `health-check-invocation-timeout` seconds on each health check call before failing: this is equivalent to using the `timeoutSeconds` field on Kubernetes probes.
* Cloud Foundry will wait `timeout` seconds after the app has been started before failing: this is equivalent to using a startup probe in Kubernetes, but there is no equivalent setting for `timeout`. Instead, it is possible to set the `failureThreshold` which specifies the _number_ of attempts before failing.
* Cloud Foundry always restarts containers as soon as they fail a health check call: this is the equivalent of using liveness probes with a `failureThreshold` of `1` on Kubernetes. There is no feature in Cloud Foundry that mirrors readiness probes.
* Cloud Foundry will perform a health check call every 2 seconds during startup, and every 30 seconds afterwards. These values are hardcoded. This is the equivalent of using a startup probe with `periodSeconds` set to `2` and a liveness probe with `periodSeconds` set to `30`.

[...]

PR #1706, opened 2022-09-27
Create the correct healthcheck probes

kieron-dev:

#1701

## What is this change about?

CF allows `timeout` seconds for an app to start, with a probe each 2 seconds. Once started, CF probes the app each 30 seconds, killing the container if a probe fails. The `health-check-invocation-timeout` determines how long to wait for a response for these probes.

To reproduce this functionality in korifi we need to create:

1. A startup probe with:

-   periodSeconds: 2
-   timeoutSeconds: `health-check-invocation-timeout`
-   failureThreshold: (`timeout` / 2) rounded to the next integer.

2. A liveness probe with:

-   periodSeconds: 30
-   timeoutSeconds: `health-check-invocation-timeout`
-   failureThreshold: 1

The readiness probe is not required as this does not correspond to CF functionality. It would stop k8s routing traffic to an app failing the readiness probe, but would not terminate it. This might be a nice-to-have in the future.

No

## Acceptance Steps

See #1701

@georgethebeatle

### DIFF
--- a/api/handlers/space_manifest_handler_test.go
+++ b/api/handlers/space_manifest_handler_test.go
@@ -288,6 +288,82 @@ var _ = Describe("SpaceManifestHandler", func() {
 			})
 		})
 
+		When("a process's healthcheck timeout is not a positive integer", func() {
+			BeforeEach(func() {
+				requestBody = strings.NewReader(`---
+                applications:
+                - name: test-app
+                  processes:
+                  - type: web
+                    timeout: 0
+                `)
+			})
+
+			It("response with an unprocessable entity error", func() {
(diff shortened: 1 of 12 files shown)


kieron-dev:
> > The readiness probe is not required as this does not correspond to CF functionality. It would stop k8s routing traffic to an app failing the readiness probe, but would not terminate it. This might be a nice-to-have in the future.
> 
> I assume for slow starting apps the new `startupProbe` we're adding will keep them from getting traffic prematurely as well, right? The readiness probe would be more around an app that went "un-ready" sometime later in its lifecycle.

Yep - liveness and readiness kick in after startup suceeds

Issue #714, opened 2022-02-25
[Chore]: Use Patch instead of Update


georgethebeatle:
### Background

When performing Update you have to get the object first, then modify it and finally attempt to Update it. This opens up a gap for the object to be asynchronously changed by another thread resulting in conflicts. One way to work around that is to use Patch instead, as demonstrated [here](https://github.com/cloudfoundry/cf-k8s-controllers/blob/489b41eecfcbc08151254da2fe9c6513300ed1fa/controllers/controllers/networking/cfroute_controller.go#L171). If Patch is used the operation is atomic and conflicts are avoided.

### Action to take

Change the code to use Patch instead of Update everywhere

### Impact

Code is more robust

danail-branekov:
`Update` has been replaced by `Patch` in the route and service instance controllers, leaving the chore opened to address the rest of the `Update` occurrences.

georgethebeatle:
Today's progress is summarized in this commit: https://github.com/cloudfoundry/korifi/commit/f3cbb9a06d1492585eca66577d05c435c3938200

It works in e2es and integrations, but the pathcher tests are broken, because the test objects (a Pod and a Secret) lack type metadata and cannot be looked up form the Schema which seems to be populated properly. Tests need more looking into, otherwise looks promising.

danail-branekov:
With latest PR the chore has been implemented for the "core" controllers. What is left to do:
- use `PatchingReconciler` in `kpack-image-builder`, `statefulset-runner`, `job-task-runner` controllers
- use the `k8s.Patch` utility in repositories
- use the `k8s.Patch` utility in repositories in integration tests 
- check for other usages of `k8sClient.Update` and `k8sClient.Patch` and deal with them accordingly

PR #1707, opened 2022-09-27
Use `PatchingReconciler` in "core" controllers

danail-branekov:
NOTE: DO NOT merge prior releasing 0.3

#714

## What is this change about?
Use `PatchingReconciler` in all "core" controllers so that those controllers do
not need to patch the object being reconciled anymore. If they need to patch a
dependent object (e.g. a referenced secret), they use the `k8s.Patch` utility
that implements the standard patching k8s idiom.

No

## Acceptance Steps
Not functional change, green tests

### DIFF
--- a/controllers/controllers/networking/cfroute_controller.go
+++ b/controllers/controllers/networking/cfroute_controller.go
@@ -23,6 +23,7 @@ import (
 
 	korifiv1alpha1 "code.cloudfoundry.org/korifi/controllers/api/v1alpha1"
 	"code.cloudfoundry.org/korifi/controllers/config"
+	"code.cloudfoundry.org/korifi/tools/k8s"
 
 	"github.com/go-logr/logr"
 	contourv1 "github.com/projectcontour/contour/apis/projectcontour/v1"
@@ -34,6 +35,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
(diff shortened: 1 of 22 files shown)


controllers/controllers/networking/cfroute_controller.go
```diff
@@ -165,41 +152,27 @@ func createInvalidRouteStatus(cfRoute *korifiv1alpha1.CFRoute, description, reas
 }
 
 func (r *CFRouteReconciler) setRouteErrorStatusAndReturn(ctx context.Context, cfRoute *korifiv1alpha1.CFRoute, err error, description, reason string) (ctrl.Result, error) {
```
georgethebeatle, line 154: This function is a bit weird, since it takes an error argument simply to return it. Also it does not seem to use the context argument. To me it would look better if it was called `setRouteErrorStatus` and did not return an error. Then its usage will be something like:

```
if apierrors.IsNotFound(err) {
	r.setRouteErrorStatus(cfRoute, "CFDomain not found", "InvalidDomainRef")
        return err
}
```

controllers/controllers/workloads/cfspace_controller.go
```diff
@@ -127,28 +110,24 @@ func (r *CFSpaceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ct
...
 		return ctrl.Result{}, err
 	}
 
 	cfSpace.Status.GUID = namespace.Name
-	err = updateStatus(ctx, r.client, cfSpace, metav1.ConditionTrue)
-	if err != nil {
-		r.log.Error(err, fmt.Sprintf("Error updating status on CFSpace %s/%s", req.Namespace, req.Name))
-		return ctrl.Result{}, err
-	}
+	setStatusCondition(&cfSpace.Status.Conditions, StatusConditionReady, metav1.ConditionTrue, StatusConditionReady, "")
```
georgethebeatle, line 130: It is arguable if this helper func is more readable than the `meta.SetStatusCondition` utility it wraps, where the literal syntax makes the meaning of each string more clear. Maybe consider inlining?

georgethebeatle (approved):
LGTM, there are some minor comments inline

controllers/controllers/workloads/cfspace_controller.go
danail-branekov: Yeah, I was thinking about the same. I tried replacing the function with `meta.SetStatusCondition` on one place and I thought that looks a bit more unreadable. Can't really make up my mind, but as you raised the question as well, I guess it is worth doing.

controllers/controllers/networking/cfroute_controller.go
danail-branekov: Sounds reasonable
+1

PR #1765, opened 2022-10-10
Use gsed on macOS and sed on linux

matt-royal:
No

## What is this change about?
Make `scripts/deploy-on-kind.sh` work on macOS again

No, it fixes a breaking change

## Acceptance Steps
Run `scripts/deploy-on-kind.sh` in both a linux and macOS environment and confirm that it results in a healthy korifi installation

 @clintyoshimura

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,13 @@ else
 GOBIN=$(shell go env GOBIN)
 endif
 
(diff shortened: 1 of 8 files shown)


clintyoshimura (approved):
LGTM

PR #1774, opened 2022-10-11
`containerdConfigPatches` should be a list

gcapizzi:

No.

## What is this change about?

This turns `containerdConfigPatches` into a YAML list. It looks like older versions of Kind tolerated a single value, but newer ones (I'm running `v0.16.0`) wants a list.

I believe this change should still work on older versions, although I haven't tested it.


kieron-dev:
This was a mistake fixed already by #1768

kieron-dev:
Closing as there is zero diff!

PR #1788, opened 2022-10-13
Implement logging proposal for korifi-controllers

acosta11:
#1180 

## What is this change about?
Configure the korifi-controllers logger according to the proposal in the linked issue

No

## Acceptance Steps
Deploy korifi as you normally would and check the logs with k9s or kubectl directly. See that the log format for controller-manager logs matches the proposal.

paired w/ @clinty

## Things to remember
Note out comment on the issue that we don't think it's currently possible to have controller-runtime pass the configured logger through to the underlying client-go klog logging that generates the leadership election logs.

### DIFF
--- a/controllers/main.go
+++ b/controllers/main.go
@@ -22,6 +22,8 @@ import (
 	"os"
 	"time"
 
+	"go.uber.org/zap/zapcore"
+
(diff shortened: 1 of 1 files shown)


clintyoshimura (approved):
LGTM

Issue #1802, opened 2022-10-17
[Bug]: Changes to CFSpace create new service account token secrets on k8s v1.23


matt-royal:
### What happened?

kubernetes v1.23 clusters that have korifi installed are showing many service account token secrets in each CFSpace's namespace. There should be exactly one of these per service account per namespace, but we're seeing many versions.

### What you expected to happen

There should be exactly one of these per service account per namespace.

### Acceptance Criteria

GIVEN korifi deployed to a kubernetes *v1.23* cluster
WHEN I create a space
AND then I update that CFSpace resource (for example, by adding a label)
THEN I should see only one secret in the corresponding namespace for each of the propagated service account's tokens (see example of bad output below)

```
$ k get secret -n cf-space-36854d47-af45-44cb-9df3-fdf73dc4d8fc
NAME                                                 TYPE                                  DATA   AGE
60ab2648-36aa-49cb-8e47-b2a2887d2849-env             Opaque                                0      85s
60ab2648-36aa-49cb-8e47-b2a2887d2849-vcap-services   Opaque                                1      85s
default-token-m9kjh                                  kubernetes.io/service-account-token   3      85s
image-registry-credentials                           kubernetes.io/dockerconfigjson        1      85s
korifi-app-token-m5jpv                               kubernetes.io/service-account-token   3      85s
korifi-app-token-sd56j                               kubernetes.io/service-account-token   3      85s
korifi-app-token-w2pkz                               kubernetes.io/service-account-token   3      85s
korifi-task-token-b6vbl                              kubernetes.io/service-account-token   3      85s
korifi-task-token-qfw7c                              kubernetes.io/service-account-token   3      85s
korifi-task-token-r6h5l                              kubernetes.io/service-account-token   3      85s
kpack-service-account-token-6d2wr                    kubernetes.io/service-account-token   3      85s
kpack-service-account-token-d9jbw                    kubernetes.io/service-account-token   3      85s
kpack-service-account-token-v699l                    kubernetes.io/service-account-token   3      85s
```

### How to reproduce it (as minimally and precisely as possible)

Follow acceptance criteria after changing the kind cluster image to v1.23.4. For example, you can change scripts/deploy-on-kind.sh to have this line
```
kind create cluster --name "${cluster}" --wait 5m --image=kindest/node:v1.23.4 --config="$SCRIPT_DIR/assets/kind-config.yaml"
```

### Environment

[...]

PR #1803, opened 2022-10-17
Don't clobber secrets on propagated service accounts

matt-royal:

[#1802]

## What is this change about?

In kubernetes v1.23 and before, secrets were automatically created for service accounts' tokens. These secrets were automatically added to the service account resource. Our code was ignoring this behavior and rewrote the serviceaccount.secrets on every update. This led to a recreation of the service account's token secret any time the CFSpace reconciler ran.

 No

## Acceptance Steps

[#1802]

 @akrishna90


akrishna90 (approved):
LGTM

Issue #1811, opened 2022-10-18
DRAFT [Feature]: Operator can modify the secrets on propagated root-namespace service accounts


matt-royal:
### Blockers/Dependencies

#1802

### Background

**As** an operator
**I want** to be able to modify the secrets on the service accounts in my CF root namespace and see the changes propagate into CFSpace namespaces
**So that** I can rotate credentials

### Acceptance Criteria

**GIVEN** a korifi deployment on kubernetes v1.23
**AND** a CFSpace referred to below as "the space"
**WHEN** I create a new secret named "the-new-secret" in the `cf` namespace
**AND** I redeploy korifi with the `packageRegistrySecretName` config set to "the-new-secret"
**AND** I update the `kpack-service-account` ServiceAccount in the `cf` namespace by replacing mentions of `image-registry-credentials` in `.secrets` and `.imagePullSecrets` with `the-new-secret`
**THEN** I should see that a secret named `the-new-secret` exists in the space's namespace and that it has the same credentials as the version in the `cf` namespace
**AND** I should see that the `image-registry-credentials` secret does not exist in the space's namespace (it was deleted)
**AND** I should see that the copy of the `kpack-service-account` ServiceAccount in the space's namespace has `.secrets` and `.imagePullSecrets` that contain `the-new-secret` and do not contain `image-registry-credentials`
**AND** I should see that there is exactly 1 secret in the space's namespace that has a name starting with `kpack-service-account-token-`. If more than 1 exists, this is a reversion of the fix in #1802 

### Dev Notes

This is complicated and includes:
- Propagating the new package registry credential secret into space namespaces when the `packageRegistrySecretName` config changes
- Deleting the old package registry credential secret from those same spaces
- Updating the `.secrets` and `.imagePullSecrets` fields on propagated service accounts to include the new package registry credential secret and exclude the old one WITHOUT clearing the secrets that represent service account tokens

matt-royal:
This whole flow is confusing and will be complicated to implement. I suggest that we document the limitation (service account secret refs aren't updated after initial propagation) and leave it at that. If an operator ever runs into this limitation, then we can get details about their specific needs to help us implement a feature that meets their needs.

PR #1813, opened 2022-10-18
Update secrets on propagated service accounts on all updates

matt-royal:

[#1811]

## What is this change about?

Before this change, the Secrets and ImagePullSecrets were set on creation and never updated

No

## Acceptance Steps
See [#1811]

@julian-hj

### DIFF
--- a/controllers/controllers/workloads/integration/cfspace_controller_integration_test.go
+++ b/controllers/controllers/workloads/integration/cfspace_controller_integration_test.go
@@ -161,8 +161,7 @@ var _ = Describe("CFSpaceReconciler Integration Tests", func() {
 						"ObjectMeta": MatchFields(IgnoreExtras, Fields{
 							"Name": Equal(serviceAccount.Name),
 						}),
-						"ImagePullSecrets": ConsistOf(MatchFields(IgnoreExtras, Fields{"Name": Equal(packageRegistrySecretName)})),
-						"Secrets":          ConsistOf(MatchFields(IgnoreExtras, Fields{"Name": Equal(packageRegistrySecretName)})),
+						"Secrets": ConsistOf(MatchFields(IgnoreExtras, Fields{"Name": Equal(packageRegistrySecretName)})),
 					}),
 				))
 			}).Should(Succeed())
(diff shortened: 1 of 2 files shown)


controllers/controllers/workloads/integration/cfspace_controller_integration_test.go
```diff
@@ -295,25 +296,127 @@ var _ = Describe("CFSpaceReconciler Integration Tests", func() {
...
+		})
+	})
+
+	When("the package registry secret is added to the root service account", func() {
+		var (
+			rootServiceAccount       *corev1.ServiceAccount
+			propagatedServiceAccount corev1.ServiceAccount
+			tokenSecretName          string
+		)
+		BeforeEach(func() {
```
danail-branekov, line 332: There is quite a substantial repetition in the `BeforeEach` setup of the `the package registry secret is added to the root service account` and `the image pull secrets are updated on the root service account`. Would it make sense to group those two under a `Describe("Updating the service account secrets"...)` so that the two contexts just configure the service account in their `BeforeEach`?

controllers/controllers/workloads/integration/cfspace_controller_integration_test.go
```diff
@@ -295,25 +296,127 @@ var _ = Describe("CFSpaceReconciler Integration Tests", func() {
...
+			// Simulate k8s adding a token secret to the propagated service account
+			origPropagatedServiceAccount := propagatedServiceAccount.DeepCopy()
+			tokenSecretName = rootServiceAccount.Name + "-token-XYZABC"
+			propagatedServiceAccount.Secrets = []corev1.ObjectReference{{Name: tokenSecretName}}
+			Expect(k8sClient.Patch(ctx, &propagatedServiceAccount, client.MergeFrom(origPropagatedServiceAccount))).To(Succeed())
+
+			// Add the package registry secret to the root service account
+			origRootServiceAccount := rootServiceAccount.DeepCopy()
+			rootServiceAccount.Secrets = []corev1.ObjectReference{{Name: packageRegistrySecretName}}
+			Expect(k8sClient.Patch(ctx, rootServiceAccount, client.MergeFrom(origRootServiceAccount))).To(Succeed())
```
danail-branekov, line 363: We have recently introduced [`k8s.PatchResource`](https://github.com/cloudfoundry/korifi/blob/72cc20246f06fe67741da88d24364ac6f5fb6cec/tools/k8s/patch.go#L80) (or `Patch` if you want to also patch status) to patch k8s objects so that you do not have to do the `deep-copy` - `client.MergeFrom` dance yourself. We spent quite some time on using these utilities all across the codebase (we might have missed some places though) to simplify the code a bit. I would suggest using them here as well.

danail-branekov:
The pull request makes controllers tests fail (I git the same failure as the one in actions locally).

Also, in #1811 Matt wrote 

```
This whole flow is confusing and will be complicated to implement. I suggest that we document the limitation (service account secret refs aren't updated after initial propagation) 
```

The limitation is not documented, maybe that should be part of the PR as well?

julian-hj:
> The limitation is not documented, maybe that should be part of the PR as well?

This latest PR actually removes that limitation, which turned out to be not that hard to fix, so we don't think there is a limitation to document anymore.

controllers/controllers/workloads/integration/cfspace_controller_integration_test.go
julian-hj: We thought about that, but each of the tests differs in subtle ways from the others, so it felt like it wasn't worth the complexity to try to refactor it while accommodating all of the slightly different behaviors we still need.

julian-hj (approved):

Issue #1731, opened 2022-09-30
[Feature]: Operator can see that no statefulsets are created for 0 instance processes


matt-royal:
### Background

**As an** operator
**I want** to see that there are no unnecessary resources are created for processes that aren't running
**So that** I don't have unnecessary resource bloat on my clusters

### Acceptance Criteria

**GIVEN** I push a multi-process app with only the "web" process scaled to >=1 instances (pushing any go app will do this)
**WHEN I** examine the resources in the app's space's namespace
**THEN I** see that there is 1 AppWorkload and 1 statefulset for the app, and they represent the "web" process

---

**WHEN I** scale the "web" process to 0 instances and scale the other process to >= 1
**THEN I** should see that there is 1 AppWorkload and 1 statefulset for the app, and these resources represent the process that has >= 1 instances (i.e not web, the other one)

---

**WHEN I** scale all processes on the app to 0 instances but leave the app in the "started" state
**THEN I** should see that there are no AppWorkload or statefulset resources for the app

### Dev Notes

We currently create an AppWorkload for every process on a started app, regardless of how many instances are requested for that process. Instead, we should only create AppWorkloads when there are >= 1 instances. We also need to ensure that we handle the case when a process is scaled down to 0 instances (i.e. delete the AppWorkload) or scaled up from 0 instances (i.e. create the AppWorkload)

matt-royal:
Confirmed on main. Thanks!

PR #1814, opened 2022-10-19
Delete AppWorkloads for processes scaled down to 0

danail-branekov:

https://github.com/cloudfoundry/korifi/issues/1731

## What is this change about?

Delete `AppWorkload`s for `CFProcess`es that have been scaled down to 0 instances. This has the side effect that `Statefulset`s for that app workload are also deleted as they are owned by `AppWorkload`s

As `CFProcess.Spec.DesiredInstances` is a pointer (i.e. can be unset), this PR treats the unset instances pointer as zero instances, therefore when `DesiredInstances` is unset on the spec, the `AppWorkload` gets deleted.

No

## Acceptance Steps

See #1731

### DIFF
--- a/controllers/controllers/workloads/cfprocess_controller.go
+++ b/controllers/controllers/workloads/cfprocess_controller.go
@@ -94,7 +94,7 @@ func (r *CFProcessReconciler) ReconcileResource(ctx context.Context, cfProcess *
 		cfAppRev = foundValue
 	}
 
-	if cfApp.Spec.DesiredState == korifiv1alpha1.StartedState {
+	if needsAppWorkload(cfApp, cfProcess) {
 		err = r.createOrPatchAppWorkload(ctx, cfApp, cfProcess, cfAppRev)
 		if err != nil {
 			return ctrl.Result{}, err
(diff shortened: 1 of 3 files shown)


controllers/controllers/workloads/cfprocess_controller.go
```diff
@@ -165,7 +180,9 @@ func (r *CFProcessReconciler) cleanUpAppWorkloads(ctx context.Context, cfProcess
 	}
 
 	for i, currentAppWorkload := range appWorkloadsForProcess {
-		if desiredState == korifiv1alpha1.StoppedState || currentAppWorkload.Labels[korifiv1alpha1.CFAppRevisionKey] != cfAppRev {
+		if desiredState == korifiv1alpha1.StoppedState ||
+			desiredProcessInstances(cfProcess) == 0 ||
+			currentAppWorkload.Labels[korifiv1alpha1.CFAppRevisionKey] != cfAppRev {
```
georgethebeatle, line 185: Maybe these conditions can be extracted in a predicate function `needsCleanup`, so that it is symmetrical with the creation case?

controllers/controllers/workloads/cfprocess_controller.go
```diff
@@ -109,6 +109,21 @@ func (r *CFProcessReconciler) ReconcileResource(ctx context.Context, cfProcess *
...
+func needsAppWorkload(cfApp *korifiv1alpha1.CFApp, cfProcess *korifiv1alpha1.CFProcess) bool {
+	if cfApp.Spec.DesiredState != korifiv1alpha1.StartedState {
+		return false
+	}
+	return desiredProcessInstances(cfProcess) > 0
+}
+
+func desiredProcessInstances(cfProcess *korifiv1alpha1.CFProcess) int {
+	if cfProcess.Spec.DesiredInstances == nil {
+		return 0
```
georgethebeatle, line 121: I wonder if  `0` is a reasonable default, since the app does not do anything when there are 0 instances. If a kubectl user wants to push an app without specifying too much stuff wouldn't `1` be a better default? Scaling down to zero sounds like a special case rather than a default.

Issue #1786, opened 2022-10-13
[Bug]: Korifi API doesn't notice when new ingress certs are mounted


julian-hj:
### What happened?

If you install korifi and wait 60 days, cert manager will generate a new cert.

Korifi should notice this cert, but it doesn't pick it up until it is restarted.

### What you expected to happen

Korifi should notice changes to the tls cert and start serving with the new cert. cert-manager should be updating the Secret containing this cert once it rotates it which will update the secret files that are volume mounted to our deployment. 

### Acceptance Criteria

GIVEN a Korifi deployment
WHEN I wait 60 days
THEN I see that Korifi is running with a new cert and ingress still works.

### How to reproduce it (as minimally and precisely as possible)

- deploy korifi
- get cert manager to replace the cert somehow.

### Anything else we need to know?

We can probably use the same package in controller-runtime that our webhooks are using to watch and reload the cert as it changes on disk:
https://github.com/kubernetes-sigs/controller-runtime/pull/421 

### Environment

Revision of codebase:
Kubernetes version (use `kubectl version`):
Cloud provider or hardware configuration:
Others:

julian-hj:
I was able to reproduce this by
- running deploy-on-kind
- deleting the korifi-api-internal secret and certificate
- running deploy-on-kind again

This procedure makes a new ingress cert without restarting the pod.

julian-hj:
After some further testing, we figured out that we could install the cert in such a way that it rolls every 3 minutes:

In this case, the cert is good for 90 minutes, but renews 87 minutes before it expires, or once every 3 minutes.
```
apiVersion: cert-manager.io/v1
kind: Certificate
metadata:
  name: korifi-api-internal-cert
  namespace: {{ .Release.Namespace }}
spec:
  duration: 90m
  renewBefore: 87m
  commonName: korifi-api-svc.{{ .Release.Namespace }}.svc.cluster.local
  dnsNames:
  - korifi-api-svc.{{ .Release.Namespace }}.svc
  - korifi-api-svc.{{ .Release.Namespace }}.svc.cluster.local
  issuerRef:
    kind: Issuer
    name: selfsigned-issuer
  secretName: korifi-api-internal-cert
  subject:
    organizations:
    - korifi
  usages:
  - server auth
 ```
After applying this, we noticed that Contour takes a minute or so to notice the new cert, so even after our fix, there is a 1 minute period where Contour can't talk to the API.

PR #1826, opened 2022-10-21
Fix gosec vuln in api

julian-hj:
[#1786]

## What is this change about?
gosec flags a vulnerability about TLS version

No

## Acceptance Steps
`gosec api/...` should not report G402 (there are some other false positives still)

@matt-royal


acosta11 (approved):
LGTM

PR #1832, opened 2022-10-24
Remove the pre-install hook that created the root namespace

gcapizzi:
This allows users to create their container registry secret before they install Korifi. This will make sure that their Kpack `ClusterBuilder` will be immediately ready after the installation.

Creating the secret after the installation sometimes caused the `ClusterBuilder` to get stuck in a `Ready: False` state.


INSTALL.md
```diff
@@ -125,6 +114,36 @@ Make sure the value of `--docker-server` is a valid [URI authority](https://data
...
 
+## Install Korifi
+
+Korifi is distributed as a [Helm chart](https://helm.sh). See [_Customizing the Chart Before Installing_](https://helm.sh/docs/intro/using_helm/#customizing-the-chart-before-installing) for details on how to specify values when installing a Helm chart.
+
+For example:
+
+```sh
+helm install korifi https://github.com/cloudfoundry/korifi/releases/download/v<VERSION>/korifi-<VERSION>.tgz \
+    --namespace="$KORIFI_NAMESPACE" \
```
georgethebeatle, line 125: Do we also need to pass `--create-namespace` so that helm creates the korifi namespace?

gcapizzi: people _could_ do that if they want, but I don't think it should be in the docs - we ask for a namespace, it's up to them how they get one

scripts/deploy-on-kind.sh
```diff
@@ -138,21 +138,10 @@ function deploy_korifi() {
...
-kind: Namespace
-metadata:
-  labels:
-    pod-security.kubernetes.io/enforce: $secLevel
-  name: korifi
-EOF
-
     if [[ -n "$use_custom_registry" ]]; then
       helm upgrade --install korifi helm/korifi \
         --namespace korifi \
```
georgethebeatle, line 147: We need to add the `--create-namespace` flag if we will no longer be creating the korifi namespace as part of the deploy script. In this case we probably need to patch the created namespace with the pod security standards labels, because helm won't do it. Alternatively this can become a manual step as well.

gcapizzi: oh yeah my mistake - I thought I was just moving the root namespace creation, but this is actually the Korifi namespace 😓 `--create-namespace` should be enough, I don't think we need to add the annotations

georgethebeatle: Sounds good

georgethebeatle (approved):

PR #1851, opened 2022-11-01
Fix service account image pull secret propagation

clintyoshimura:
no

## What is this change about?
Fix service account image pull secret propagation when copying service accounts to spaces. Some distributions of k8s create a dockercfg auto generated secret and adds it to service accounts. Copying service accounts with these secrets present in the secret or image pull secret fields causes a not found error because those secrets will not be copied into the space namespace.

no

## Acceptance Steps
All unit, integration, and e2e tests pass.

@davewalter @matt-royal @julian-hj


davewalter (approved):
Paired with Clint.

PR #1864, opened 2022-11-03
Format logs from low level loggers

julian-hj:
[#1179]

## What is this change about?
We found some additional cases where low level loggers from klog and the http server's log.Logger (the golang standard library logger) were getting into the log stream unformatted. This PR redirects those log streams.

- set klog.Log to use the zapr logger
- create an io.Writer shim to collect error messages from the http server ErrorLogger and redirect them to our zapr logger's Error() method.

No

## Acceptance Steps
some details are in the comments of [#1179]

@clintyoshimura

### DIFF
--- a/api/main.go
+++ b/api/main.go
@@ -3,7 +3,9 @@ package main
 import (
 	"context"
 	"crypto/tls"
+	"errors"
 	"fmt"
+	"log"
 	"net/http"
 	"net/url"
 	"os"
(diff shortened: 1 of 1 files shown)


clintyoshimura (approved):
LGTM

kieron-dev (approved):
I can't think of a better way of doing this as we have to use the Logger struct.

PR #1877, opened 2022-11-08
Remove the finaizer from CFServiceInstance

danail-branekov:
https://github.com/cloudfoundry/korifi/issues/1876

## What is this change about?
The finalizer has been replaced by setting owner reference from the service to the
service binding in the service binding repository

No

## Acceptance Steps
N/A

@kieron-dev


tools/k8s/owner_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+	"code.cloudfoundry.org/korifi/tools/k8s"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	. "github.com/onsi/gomega/gstruct"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = FDescribe("Owner", func() {
```
georgethebeatle, line 14: Opps :)

kieron-dev (changes requested):

danail-branekov:
We have realised that with this change a service binding has two owners - the app and the service. If one of them is deleted, the binding remains behind. In order to be deleted automatically, both the app AND the service have to be deleted. This is something we don't really want

danail-branekov:
This PR is no longer needed. The issue addressed with https://github.com/cloudfoundry/korifi/pull/1882

PR #1882, opened 2022-11-10
Simplify finalizers

danail-branekov:
https://github.com/cloudfoundry/korifi/issues/1876

## What is this change about?
* Use owner references to clean up dependent resources instead of doing that in finalizers wherever possible
* The app finalizer makes best effort to delete related service bindings
* Favour setting owner references in controllers' reconcile log
* ADR to record our decision about managing owner references

No

## Acceptance Steps
N/A

@georgethebeatle


clintyoshimura:
FYI In OpenShift we need finalizer permissions on the owning object to set owner references

Issue #1869, opened 2022-11-07
[Bug]: switching buildpacks with subsequent cf pushes fails to update the start command


julian-hj:
### What happened?

If you push a node app and then push a java app with the same name, it fails to start up.

### What you expected to happen

It should start up.

### Acceptance Criteria

GIVEN I have a running app using the node buildpack
WHEN I cf push a java app with the same name
THEN I see that my java app eventually starts up.

GIVEN I have a running app using the node buildpack
WHEN I change the name of server.js to index.js (or some similar change) and re-push
THEN I see that my node app eventually starts back up.

### How to reproduce it (as minimally and precisely as possible)

See A/C

### Anything else we need to know?

It appears that the command on the "web" process doesn't get updated on the re-push.

(maybe we don't actually need to set the command most of the time. The container entry point should already be correct)

```
$ cf curl /v3/processes/cf-proc-27f0c6c2-120d-4377-88c5-5cb835381c17-web

[...]

PR #1887, opened 2022-11-10
Handle changing process commands across builds

julian-hj:
[#1869]

## What is this change about?
- Prior to this change, when an app was re-pushed with source that caused the `command` to change for a CFProcess, we'd ignore the change.
- For example pushing a node app, and then pushing java code to the same app name would leave the `node` start command, which would then fail.
- This commit adds a separate `dropletCommand` field to store the command from the built image. If no `command` is explicitly set on the process, the `dropletCommand` is used.
- For apps that explictly set a `command` for the process (for example in a manifest) the `command` overrides the droplet command.

image

No

## Acceptance Steps
[#1869]

@tcdowney PTAL. I'm especially open to different suggestions for the field name `dropletCommand`

### DIFF
--- a/controllers/api/v1alpha1/cfprocess_types.go
+++ b/controllers/api/v1alpha1/cfprocess_types.go
@@ -40,7 +40,7 @@ type CFProcessSpec struct {
 	Command string `json:"command,omitempty"`
 
 	// The default command for this process as defined by the build. This field is ignored when the Command field is set
-	DropletCommand string `json:"dropletCommand,omitempty"`
+	DetectedCommand string `json:"detectedCommand,omitempty"`
 
 	// Used to build the Liveness and Readiness Probes for the process' AppWorkload.
 	HealthCheck HealthCheck `json:"healthCheck"`
--- a/controllers/api/v1alpha1/cfprocess_types.go
+++ b/controllers/api/v1alpha1/cfprocess_types.go
@@ -39,6 +39,9 @@ type CFProcessSpec struct {
 	// Command string used to run this process on the app image. This is analogous to command in k8s and ENTRYPOINT in Docker
 	Command string `json:"command,omitempty"`
 
+	// The default command for this process as defined by the build. This field is ignored when the Command field is set
+	DropletCommand string `json:"dropletCommand,omitempty"`
+
 	// Used to build the Liveness and Readiness Probes for the process' AppWorkload.
 	HealthCheck HealthCheck `json:"healthCheck"`
 
--- a/controllers/controllers/workloads/integration/cfprocess_controller_integration_test.go
+++ b/controllers/controllers/workloads/integration/cfprocess_controller_integration_test.go
@@ -46,6 +46,8 @@ var _ = Describe("CFProcessReconciler Integration Tests", func() {
 
 		processTypeWeb           = "web"
 		processTypeWebCommand    = "bundle exec rackup config.ru -p $PORT -o 0.0.0.0"
+		oldDropletCommand        = "old-droplet-command"
+		newDropletCommand        = "new-droplet-command"
 		processTypeWorker        = "worker"
 		processTypeWorkerCommand = "bundle exec rackup config.ru"
 		port8080                 = 8080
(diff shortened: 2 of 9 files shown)


controllers/api/v1alpha1/cfprocess_types.go
```diff
@@ -39,6 +39,9 @@ type CFProcessSpec struct {
 	// Command string used to run this process on the app image. This is analogous to command in k8s and ENTRYPOINT in Docker
 	Command string `json:"command,omitempty"`
 
+	// The default command for this process as defined by the build. This field is ignored when the Command field is set
+	DropletCommand string `json:"dropletCommand,omitempty"`
```
Birdrock, line 43: We're discussing the naming of this field. It gets pulled from the OCI image config labels, so we're not quite sure what to name it. In a sense, it is the default command, but default is a term we want to be pretty careful about using. Anyone have additional thoughts?

controllers/controllers/workloads/integration/cfprocess_controller_integration_test.go
```diff
@@ -179,6 +183,33 @@ var _ = Describe("CFProcessReconciler Integration Tests", func() {
...
+					g.Expect(appWorkload.Spec.GUID).To(Equal(cfProcess.Name))
+					g.Expect(appWorkload.Spec.Version).To(Equal(cfApp.Annotations[cfAppRevisionKey]))
+					g.Expect(appWorkload.Spec.Image).To(Equal(cfBuild.Status.Droplet.Registry.Image))
+					g.Expect(appWorkload.Spec.ImagePullSecrets).To(Equal(cfBuild.Status.Droplet.Registry.ImagePullSecrets))
+					g.Expect(appWorkload.Spec.ProcessType).To(Equal(processTypeWeb))
+					g.Expect(appWorkload.Spec.AppGUID).To(Equal(cfApp.Name))
+					g.Expect(appWorkload.Spec.Ports).To(Equal(cfProcess.Spec.Ports))
+					g.Expect(appWorkload.Spec.Instances).To(Equal(int32(*cfProcess.Spec.DesiredInstances)))
+
+					g.Expect(appWorkload.Spec.Command).To(ConsistOf("/cnb/lifecycle/launcher", newDropletCommand))
```
danail-branekov, line 208: I think the important bit in this `It` is to verify the workload `Spec.Command`, the rest of the checks repeat the ones on the "happy path".
I would therefore suggest that we only leave this `Exect` in this test, this would make the test easier to reason about how the context affects the result.

controllers/api/v1alpha1/cfprocess_types.go
danail-branekov: How about we call this field `Command` and instead rename the old `Command` to `CustomCommand`, or `UserSpecifiedCommand` or something like this? I guess that such a change would be even transparent to currently running workloads, i.e. hopefully there would be no upgrade issues.

danail-branekov (approved):
I left few minor comments, but overall looks good to me.

controllers/api/v1alpha1/cfprocess_types.go
tcdowney: I think it's useful to leave the user-specific command field as just `command` since that aligns with the existing CF APIs.

https://v3-apidocs.cloudfoundry.org/version/3.128.0/index.html#the-process-object

Screen Shot 2022-11-14 at 10 14 38 AM

I'm not sure if it's publicly referred to as this often, we've referred to the "droplet command" as 
the "detected command" because the buildpacks use that that terminology a lot (I think in CF for VMs it comes from a callback object called the "buildpack detection receipt" for example).

Here's some various naming there.
https://github.com/cloudfoundry/cloud_controller_ng/blob/3dcc660170c86c581b0d26ee4290028f38cbbd75/app/models/runtime/process_model.rb#L410-L412

So I have a slight preference to `DetectedCommand` or `ImageCommand` or something. `dropletCommand` is fine too, I'm just not sure how much life the "droplet" metaphor has left in it on Korifi.

I'm also not entirely opposed to `UserSpecifiedCommand` like @danail-branekov suggested since it's clear, but IMO there is value to being able to say this `CFProcess` looks very close to "V3 CF API Process" and that `Command` means the same thing in both contexts.

julian-hj: Thanks all, we went with `DetectedCommand` just to expunge the droplet metaphor

tcdowney (approved):

PR #1908, opened 2022-11-16
Make sure a CFTask SequenceID is only set once

georgethebeatle:
Issue: https://github.com/cloudfoundry/korifi/issues/1905

## What is this change about?
Use an explicit status patch with test of previous value, to make sure
we do not accidentally reinitialize the status if the client cache is stale.

No

## Acceptance Steps
Green tests

@eirini


controllers/controllers/workloads/cftask_controller.go
```diff
@@ -299,25 +299,55 @@ func calculateDefaultCPURequestMillicores(memoryMiB int64) int64 {
...
+	// Do not modify the original cfTask, otherwise the patching reconciler
+	// will apply the changes.
+	if err := r.k8sClient.Status().Patch(ctx, cfTask.DeepCopy(), client.RawPatch(types.JSONPatchType, []byte(jsonPatch))); err != nil {
+		r.logger.Info("error-patching-sequence-id", "error", err)
+		return err
 	}
 
+	// Now that we managed to set the seq id, modify the original cfTask,
+	// otherwise the patching reconciler will revert it back to the old value.
+	cfTask.Status.SequenceID = sequenceId
```
gcapizzi, line 350: it feels wrong to work around the patching reconciler - maybe we shouldn't use it here? or extend it so that tests can be attached to some patching operations?

danail-branekov: > or extend it so that tests can be attached to some patching operations?

How about we just return a special kind of error upon which the patching reconciler does not try to patch at all? Thus actual reconcilers would be able to communicate that they want the reconcile process to be restarted.

gcapizzi: That could be a way - alternatively we could return the `client.Patch` to be applied? Maybe two, one for the resource and the other for the status? This way how we create the patch and its contents becomes abstracted - wether it's a "normal" patch just setting a bunch of fiends or a "raw" one using tests and stuff.

gcapizzi: @danail-branekov @georgethebeatle we were thinking: doesn't this actually belong to a defaulting webhook? That would fix all our concerns we think. Same with all the other defaulting happening in the controller.

danail-branekov: The PR is now updated with setting the sequence ID with a defaulting webhook. Also, the validating webhook would reject modifying the sequence id

danail-branekov (approved):

Issue #1930, opened 2022-11-22
[Chore]: CFOrg/Space namespace deletion race


kieron-dev:
Fix these flakes:

| Ended | Job                              | Url
|-------|----------------------------------|-----------------------------------------------------|
| 1d    | main/run-tests-periodic          | https://ci.korifi.cf-app.com//teams/main/pipelines/main/jobs/run-tests-periodic/builds/4309|
| 3d    | main/run-tests-periodic          | https://ci.korifi.cf-app.com//teams/main/pipelines/main/jobs/run-tests-periodic/builds/4213|
| 4d    | main/run-tests-periodic          | https://ci.korifi.cf-app.com//teams/main/pipelines/main/jobs/run-tests-periodic/builds/4193|
| 8d    | main/run-tests-periodic          | https://ci.korifi.cf-app.com//teams/main/pipelines/main/jobs/run-tests-periodic/builds/4128|
| 14d   | main/run-tests-periodic          | https://ci.korifi.cf-app.com//teams/main/pipelines/main/jobs/run-tests-periodic/builds/3948|
| 23d   | main/run-tests-periodic          | https://ci.korifi.cf-app.com//teams/main/pipelines/main/jobs/run-tests-periodic/builds/3688|
| 24d   | main/run-tests-periodic          | https://ci.korifi.cf-app.com//teams/main/pipelines/main/jobs/run-tests-periodic/builds/3631|
| 28d   | main/run-tests-periodic          | https://ci.korifi.cf-app.com//teams/main/pipelines/main/jobs/run-tests-periodic/builds/3575|
| 31d   | main/run-tests-periodic          | https://ci.korifi.cf-app.com//teams/main/pipelines/main/jobs/run-tests-periodic/builds/3444|
| 31d   | main/run-tests-main              | https://ci.korifi.cf-app.com//teams/main/pipelines/main/jobs/run-tests-main/builds/423|
| 42d   | main/run-tests-main              | https://ci.korifi.cf-app.com//teams/main/pipelines/main/jobs/run-tests-main/builds/383|
| 45d   | main/run-tests-periodic          | https://ci.korifi.cf-app.com//teams/main/pipelines/main/jobs/run-tests-periodic/builds/3076|
| 46d   | main/run-tests-periodic          | https://ci.korifi.cf-app.com//teams/main/pipelines/main/jobs/run-tests-periodic/builds/3043|
[...]

PR #1931, opened 2022-11-22
Fix finalizer persistence race

danail-branekov:

#1930

## What is this change about?

Commit immediately after setting finalizer in controllers

There is a potential race when deleting a resource with a finalizer
when the finalizer is not immediately persisted. E.g. for a CFSpace:

1. Create the CFSpace
2. Wait for namespace to appear
3. Delete the CFSpace.

However, the reconciler could well be in the code between creating that
namespace and propagating service accounts / secrets, and nothing has
yet been saved.

When the big 'patch' call finally occurs, the CFSpace has already been
deleted, since it didn't actually have a saved finalizer on it.

All resources created by the reconciliation are now orphaned and will
not be cleaned up.

(We also tidied up logs in the touched files)

No

## Acceptance Steps

We stop seeing flakes saying 'Eventually deletes the namespace'

@kieron-dev

### DIFF
--- a/controllers/controllers/networking/cfroute_controller.go
+++ b/controllers/controllers/networking/cfroute_controller.go
@@ -18,6 +18,7 @@ package networking
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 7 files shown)


kieron-dev (approved):
paired on this

PR #1936, opened 2022-11-23
Use configured dropletRepository as given

kieron-dev:
#1914

## What is this change about?
Do not append a suffix, as this means we cannot support registries which disallow dynamic repository creation on push (e.g. ECR).

Rename some config values to make this clear to operators:

- Helm value in kpack-image-builder:
  - `dropletRepositoryPrefix` --> `dropletRepository`
- kpack-image-builder config map property
  - `kpackImageTag` --> `dropletRepository`

Finally, update docs.

Yes. Korifi packagers and users should note the two renames above.

## Acceptance Steps
1. Deploy korifi successfully using the new helm value in the kpack-image-builder chart.
2. Successfully push an app.

### DIFF
--- a/INSTALL.kind.md
+++ b/INSTALL.kind.md
@@ -56,7 +56,7 @@ No changes here. If using DockerHub as recommended above, set the following valu
 
 -   `api.packageRepository`: `index.docker.io/<username>/packages`;
 -   `kpack-image-builder.builderRepository`: `index.docker.io/<username>/kpack-builder`;
--   `kpack-image-builder.dropletRepositoryPrefix`: `index.docker.io/<username>`.
+-   `kpack-image-builder.dropletRepository`: `index.docker.io/<username>/droplets`.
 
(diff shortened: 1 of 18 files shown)


helm/kpack-image-builder/templates/cluster-builder.yaml
```diff
@@ -3,6 +3,8 @@ apiVersion: kpack.io/v1alpha2
 kind: ClusterStore
 metadata:
   name: cf-default-buildpacks
+  annotations:
+    korifi.cloudfoundry.org/helm-release-version: {{ .Release.Revision | quote }}
```
danail-branekov, line 7: Is that annotation related?

kieron-dev: Ah no. That was me trying to fix yesterday's kpack problems. I'll remove it.

danail-branekov (approved):

Issue #1922, opened 2022-11-21
[Feature]: Operator can install Korifi configuring exclusively a service account for container registry access


gcapizzi:
### Background

**As an** operator
**I want** to use a service account for container registry access
**So that** I can use Korifi with container registries that don't work well with regular secrets (e.g. ECR)

### Acceptance Criteria

**GIVEN** I have installed Korifi leaving `global.packageRegistrySecret` empty and setting `global.containerRegistryServiceAccount` to a properly authorised service account
**AND** my cluster nodes are authorised to pull images from my registry;
**WHEN I** try to push apps
**THEN I** I see everything works as expected

### Dev Notes

A (potentially incomplete) list of changes needed to achieve this:
* `global.packageRegistrySecret` and `global.containerRegistryServiceAccount` should be validated to be mutually exclusive.
* The `api` should be able to push packages using a service account instead of a secret. This could be achieved by invoking different constructors from `go-containerregistry` based on which information has been configured, or by creating a service account ourselves in case we're provided with a secret: the library should be able to retrieve the secret from the service account in that case. It shouldn't be necessary to store a secret reference in the `CFPackage`, as `global.containerRegistryServiceAccount` might not have any secrets attached.
* The service account used in our `ClusterBuilder` (which is currently hardcoded to `kpack-service-account`) needs to become `global.containerRegistryServiceAccount`.
* `controllers` will need to propagate `global.containerRegistryServiceAccount` just like they propagate `global.packageRegistrySecret`, for the same reason. They will also, obviously, need to tolerate either the secret or the service account not existing.
* The service account used in `kpack-image-builder` (which is currently hardcoded to `kpack-service-account`) needs to become `global.containerRegistryServiceAccount`. necessary to store a secret reference in the `CFBuild` (droplet), as `global.containerRegistryServiceAccount` might not have any secrets attached.
* Any `imagePullSecrets` we set to the value of `global.packageRegistrySecret` when scheduling pods (e.g. in `StatefulSet`s, `Job`s, etc) should be omitted when `global.containerRegistryServiceAccount` is provided instead. Users will have to guarantee the necessary permissions to their nodes in that case.

PR #1961, opened 2022-12-01
Make registry secrets optional

danail-branekov:
#1922

## What is this change about?
There is now an option to configure a service account with IAM
permissions to push and pull from the package, droplet and kpack-builder repositories, and to provide
that service account name rather than a secret name.

This commit contains the helm modifications, and any modifications to
make korifi deal with empty / non-existent secrets

No

## Acceptance Steps
At this stage, we want korifi on GKE to continue to work with secrets

Although, it should be possible to manually get this working on EKS/ECR now.

@kieron-dev


kieron-dev (approved):
I paired on this

PR #1988, opened 2022-12-08
Expand on race condition fix when deleting CFSpace

clintyoshimura:
no

## What is this change about?
Expands on the previously added fix for a race condition when Korifi is deployed with custom runner & builder implementations. The previous fix deleted any associated CFApps before deleting the namespace in the CFSpace finalizer since that caused resources to hang forever in their finalizer blocks. Making the owner references in the CFApp ownership chain set BlockOwnerDeletion to true ensures the child resources are fully removed before the CFApp is removed and therefore deletion of the namespace will not be blocked.

no

## Acceptance Steps
All tests pass

@akrishna90 @acosta11

### DIFF
--- a/controllers/controllers/workloads/cfspace_controller.go
+++ b/controllers/controllers/workloads/cfspace_controller.go
@@ -18,6 +18,7 @@ package workloads
 
 import (
 	"context"
+	"fmt"
 	"strings"
 	"time"
 
@@ -324,14 +325,22 @@ func (r *CFSpaceReconciler) finalize(ctx context.Context, log logr.Logger, space
 		return ctrl.Result{}, nil
 	}
 
-	err := r.finalizeCFApp(ctx, log, space.GetName())
-	if err != nil {
-		log.Error(err, "failed to finalize CFApps while deleting CFSpace")
+	duration := time.Since(space.GetDeletionTimestamp().Time)
+	log.Info(fmt.Sprintf("finalizing CFSpace for %fs", duration.Seconds()))
+	if duration < 60.0*time.Second {
+		err := r.finalizeCFApps(ctx, log, space.GetName())
+		if err != nil {
+			log.Info("failed to finalize CFApps while deleting CFSpace")
+			return ctrl.Result{RequeueAfter: 500 * time.Millisecond}, nil
+		}
+	} else {
+		log.Info("timed out finalizing CFApps while deleting CFSpace")
 	}
 
-	err = r.client.Delete(ctx, &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: space.GetName()}})
+	log.Info("deleting namespace while finalizing CFSpace")
+	err := r.client.Delete(ctx, &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: space.GetName()}})
 	if err != nil {
-		log.Error(err, "Failed to delete namespace")
+		log.Error(err, "failed to delete namespace")
 		return ctrl.Result{}, err
 	}
 
@@ -342,7 +351,7 @@ func (r *CFSpaceReconciler) finalize(ctx context.Context, log logr.Logger, space
 	return ctrl.Result{}, nil
 }
 
-func (r *CFSpaceReconciler) finalizeCFApp(ctx context.Context, log logr.Logger, namespace string) error {
+func (r *CFSpaceReconciler) finalizeCFApps(ctx context.Context, log logr.Logger, namespace string) error {
 	appList := korifiv1alpha1.CFAppList{}
 	err := r.client.List(ctx, &appList, client.InNamespace(namespace))
 	if err != nil {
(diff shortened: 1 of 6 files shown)


Birdrock (approved):
lgtm

controllers/controllers/workloads/cfspace_controller.go
```diff
@@ -383,7 +386,7 @@ func (r *CFSpaceReconciler) finalizeCFApp(ctx context.Context, log logr.Logger,
 	}
 
 	done := make(chan bool, 1)
-	go waitForCFAppDeletion(ctx, log, namespace, done)
+	waitForCFAppDeletion(ctx, log, namespace, done)
```
acosta11, line 389: Sorry we missed this on the initial implementation but it looks like we should pull the timeout channel time.After out of the for loop or else it just resets the timer each iteration.

https://go.dev/play/p/DMnPRODlXmb is an example of the timeout not triggering.
https://go.dev/play/p/_fofq32LHLH is an example of the timeout working as expected.

clintyoshimura: Good catch. Fixed

julian-hj (changes requested):

controllers/controllers/workloads/cfspace_controller.go
```diff
@@ -350,10 +352,11 @@ func (r *CFSpaceReconciler) finalizeCFApp(ctx context.Context, log logr.Logger,
...
 	}
 
+	log.Info(fmt.Sprintf("Deleting %d CFApps while finalizing CFSpace", len(appList.Items)))
 	for i := range appList.Items {
-		err = r.client.Delete(ctx, &appList.Items[i])
+		err = r.client.Delete(ctx, &appList.Items[i], client.PropagationPolicy(metav1.DeletePropagationForeground))
 		if err != nil {
-			log.Error(err, "failed to delete cfApp", "AppName", appList.Items[i].Name)
+			log.Error(err, "failed to delete CFApp", "AppName", appList.Items[i].Name)
 			return err
```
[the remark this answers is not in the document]
clintyoshimura: Good catch. Left the error message, but removed the return

controllers/controllers/workloads/cfspace_controller.go
```diff
@@ -383,7 +386,7 @@ func (r *CFSpaceReconciler) finalizeCFApp(ctx context.Context, log logr.Logger,
 	}
 
 	done := make(chan bool, 1)
```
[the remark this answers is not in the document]
clintyoshimura: Yeah this wasn't being used so I removed it.

controllers/controllers/workloads/cfspace_controller.go
```diff
@@ -351,39 +360,27 @@ func (r *CFSpaceReconciler) finalizeCFApp(ctx context.Context, log logr.Logger,
...
+	var cfAppList korifiv1alpha1.CFAppList
+	err = r.client.List(ctx, &cfAppList, client.InNamespace(namespace))
+	if err != nil {
+		log.Error(err, "failed to list CFApps while watching CFSpace")
+		return fmt.Errorf("failed to list CFApps while watching CFSpace")
+	}
+
+	if len(cfAppList.Items) > 0 {
+		err = fmt.Errorf("%d CFApps still found", len(cfAppList.Items))
+		log.Error(err, fmt.Sprintf("%d CFApps still found", len(cfAppList.Items)))
```
julian-hj, line 381: This should be a log.Info, since I think this case will occur when there isn't any real error

controllers/controllers/workloads/cfspace_controller.go
```diff
@@ -324,14 +325,22 @@ func (r *CFSpaceReconciler) finalize(ctx context.Context, log logr.Logger, space
...
 
-	err := r.finalizeCFApp(ctx, log, space.GetName())
-	if err != nil {
-		log.Error(err, "failed to finalize CFApps while deleting CFSpace")
+	duration := time.Since(space.GetDeletionTimestamp().Time)
+	log.Info(fmt.Sprintf("finalizing CFSpace for %fs", duration.Seconds()))
+	if duration < 60.0*time.Second {
+		err := r.finalizeCFApps(ctx, log, space.GetName())
+		if err != nil {
+			log.Error(err, "failed to finalize CFApps while deleting CFSpace")
```
julian-hj, line 333: The way the logic now looks, this isn't really failure anymore, right? We just haven't finished finalizing yet. So this should probably also be a log.Info?

controllers/controllers/workloads/cfspace_controller.go
```diff
@@ -324,14 +325,22 @@ func (r *CFSpaceReconciler) finalize(ctx context.Context, log logr.Logger, space
...
+			return ctrl.Result{RequeueAfter: 500 * time.Millisecond}, nil
```
julian-hj, line 334: What's the behavior when deleting a cfapp gets stuck? I think eventually we want to start returning errors here, and *maybe* also go ahead and delete the namespace, since that will clean up a lot of other resources.

julian-hj (changes requested):

controllers/controllers/workloads/cfspace_controller.go
clintyoshimura: Good point. Fixed

controllers/controllers/workloads/cfspace_controller.go
clintyoshimura: That's true. Fixed.

controllers/controllers/workloads/cfspace_controller.go
clintyoshimura: So if a CFApp gets stuck, eventually the time since calculation using the space deletion timestamp will be greater than 60s and it will fall into the else case and continue on to delete the namespace. At that point it will stop trying to finalize the CFApps.

julian-hj (approved):

PR #2024, opened 2022-12-19
Remove integration tests for role api endpoint

benjaminguttmann-avtq:

#936

## What is this change about?

Removing the integration tests for the role api endpoint


gcapizzi:
Hey @benjaminguttmann-avtq, thanks for the PR! It looks like CI is failing, can you take a look?

benjaminguttmann-avtq:
@gcapizzi I think it makes sense to also get rid of 

https://github.com/cloudfoundry/korifi/blob/main/api/handlers/integration/integration_suite_test.go and https://github.com/cloudfoundry/korifi/blob/main/api/handlers/integration/get_app_env_test.go i.e. the complete integration folder as there are no test left in there, what do you think? 

I think that would also solve the CI issue.

gcapizzi (approved):

gcapizzi:
Thank you @benjaminguttmann-avtq!

PR #2034, opened 2022-12-23
Get rid of `registry.ContainerRegistryMeta`

gcapizzi:

No.

## What is this change about?

This moves the logic appending `-packages` and `-droplets` respectively to the `PackageRepository` and the `BuildWorkloadReconciler`, and the logic splitting the registry authority from the path to the `ECRRepositoryCreator`, as it's an ECR-specific recessity.

No.


kieron-dev (approved):
lgtm

PR #2035, opened 2023-01-04
Simplify handler wrapping

gcapizzi:
No

## What is this change about?
Keep the handler method signature simple by removing authInfo and logger from the arguments. They should be fetched from context using the helpers as required. A logger containing the correlation ID is now inserted into the context by the Correlation middleware.

This removes the distinction between authorised and non-authorised handlers, so both can be simply cast into a `routing.Handler` which transforms the method signature to the expected `http.HandlerFunc` using a function method.

Also separate the routing and middleware code into their own respective packages away from the handler code. We took the opportunity to standardize the middleware construction to return functions.

No

## Acceptance Steps
tests work

@kieron-dev


danail-branekov (approved):

api/handlers/build_handler.go
```diff
@@ -54,18 +52,23 @@ func NewBuildHandler(
 	}
 }
 
-func (h *BuildHandler) buildGetHandler(ctx context.Context, logger logr.Logger, authInfo authorization.Info, r *http.Request) (*HandlerResponse, error) {
+func (h *BuildHandler) buildGetHandler(r *http.Request) (*routing.Response, error) {
 	buildGUID := chi.URLParam(r, "guid")
+	authInfo, _ := authorization.InfoFromContext(r.Context())
```
danail-branekov, line 57: Getting the `authInfo` in every handler method from the context sort of defeats the purpose of reducing the number of arguments.... It would be awesome if that did not have to be repeated all over the place.

georgethebeatle (approved):

api/handlers/build_handler.go
gcapizzi: We liked the idea of having handlers just fetch what they need from the context, so that adding new middlewares which add new information to the context didn't translate into changing that signature every time. This way every handler grabs what they need from the context, and the wrapper does only one thing, which is to translate the `routing.Response` or the `error` into calls to `http.ResposeWriter`.

danail-branekov: Yeah, I see, still a bit annoying though.

We had a short talk with @georgethebeatle on that matter. Honestly, the auth info is only used to be passed to the repositories, which use it to create the user client. If we extend this refactoring a bit, we could put the user client into the context, therefore handlers would not have to get it. Instead, repos would just get the user client from the context without creating it.

Issue #1717, opened 2022-09-28
[Bug]: Org and Space display names don't need to be more strict than CAPI


dsboulder:
### What happened?

I'm copying orgs and spaces from CF => Korifi. I understand that my resource names have to be k8s compliant, but I surprised that the copy failed because spaces weren't allowed in displayNames. Why even have display names if they don't match CF?

### What you expected to happen

DisplayNames for CF resources should match the naming restrictions of their CAPI pairs for easy interop

### Acceptance Criteria

GIVEN I have some CAPI orgs and spaces
WHEN I attempt to recreate them as K8s Korifi resources
THEN I can match the displayNames

### How to reproduce it (as minimally and precisely as possible)

Create an org directly using a space in the displayName

### Environment

<some proprietary playing around>

tcdowney:
Here's the CF for VMs Org/Space name regexes for reference:
* https://github.com/cloudfoundry/cloud_controller_ng/blob/a70905ee892592f6d05c5b178abbb1fd36f78b14/app/models/runtime/space.rb#L14
* https://github.com/cloudfoundry/cloud_controller_ng/blob/a70905ee892592f6d05c5b178abbb1fd36f78b14/app/models/runtime/organization.rb#L5

gcapizzi:
This is currently still broken, as we still store org/space display names as labels on namespaces, and valid CF display names are not always valid Kubernetes label values.

It looks like the only reason we still use these labels is quickly filtering namespaces that belong to Korifi in [`NamespacePermissions`](https://github.com/cloudfoundry/korifi/blob/6d4edd4238b6aa02dabe55b586c23a83f5a631ce/api/authorization/namespace_permissions.go#L56). This could be achieved with an empty label, as we don't use the label value at all.

Ideally the label key should change but keeping it would maintain backwards compatibility. We could stop setting it but still use it for filtering for the time being.

Having the org/space name on the namespace is still convenient, so we could move it to an annotation, as annotations have no limitations on their value.

tcdowney:
I was checking out the [K8s docs](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set) to see what the restrictions here are:

> Valid label value:
>
> * must be 63 characters or less (can be empty),
> * unless empty, must begin and end with an alphanumeric character ([a-z0-9A-Z]),
> * could contain dashes (-), underscores (_), dots (.), and alphanumerics between.

[...]

PR #2053, opened 2023-01-10
Change CFOrg and CFSpace display name validation to match CAPI

kieron-dev:
#1717

## What is this change about?
Use the same regexp for validating Org and Space display names as is used in CAPI

No

## Acceptance Steps
E.g. create an org called "My Org". The space in the name is now accepted


tcdowney (approved):
LGTM

PR #2055, opened 2023-01-10
update create-new-user script to work with MacOS Ventura

tcdowney:
Issue: https://github.com/cloudfoundry/korifi/issues/2054

## What is this change about?
MacOS Ventura ships with a different version of the base64 binary that requires the `-i` flag when using an input file. When this flag is omitted it defaults to reading from stdin which should be portable with other versions of base64.

No

## Acceptance Steps
Run the `deploy-on-kind.sh` or `create-new-user.sh` scripts on MacOS Ventura, Linux, and older versions of MacOS.


acosta11 (approved):

davewalter (approved):
Tested that both "base64 <file>" and "cat <file> | base64" work on MacOS Monterey, and give the same result.

scripts/common.sh
```diff
@@ -33,7 +33,7 @@ metadata:
   name: ${csr_name}
 spec:
   signerName: "kubernetes.io/kube-apiserver-client"
-  request: "$(base64 "${csr_file}" | tr -d "\n\r")"
+  request: "$(cat "${csr_file}" | base64 | tr -d "\n\r")"
```
julian-hj, line 36: super minor nit-pick... It's not necessary to use `cat` here
`base64 < "${csr_file}" | tr -d "\n\r"`
is slightly cleaner than
`cat "${csr_file}" | base64 | tr -d "\n\r"`

PR #2079, opened 2023-01-13
Don't attempt to add finalizers to deleted resources

davewalter:
No.

## What is this change about?
While deploying Korifi from HEAD of main, Clint and I noticed errors in the korifi-controllers logs indicating a failure to add finalizers to certain resources because they were being deleted. This PR updates the reconcilers for resources that require a finalizer to check if the resource is being deleted before attempting to add a finalizer.

No.

@clintyoshimura


clintyoshimura (approved):
LGTM

PR #2080, opened 2023-01-14
Don't attempt to add finalizers to deleted orgs

davewalter:
No

## What is this change about?
This is a follow-on to #2079 

No

## Acceptance Steps
See #2079 

@clintyoshimura


kieron-dev (approved):
I wonder if we can extract the common finalization / add finalizer logic into a single place, as it occurs in many reconcilers. Maybe into the PatchingReconciler, or a similar wrapper.

PR #2087, opened 2023-01-19
Do not check whether the object to be created by controllers already exists

danail-branekov:
https://github.com/cloudfoundry/korifi/issues/1721

## What is this change about?
Replace the controllers code that checks whether an object to be created
already exists before actually creating it with `CreateOrPatch`. Doing so
eliminates the need to verify for `AlreadyExists`/`NotExists` errors

No

## Acceptance Steps
N/A


danail-branekov:
Abandoning this, see https://github.com/cloudfoundry/korifi/issues/1721#issuecomment-1398619795

PR #2110, opened 2023-01-24
Await for valid CFRoute status prior domain deletion

danail-branekov:
https://github.com/cloudfoundry/korifi/issues/2109

## What is this change about?
This fixes a flake where the domain routes are not deleted upon deleting
the domain. The reason is that the domain may get deleted before the
controllers k8s client manager notices the CFRoutes that reference the
domain.

By awaiting the valid CFRoute status we ensure that controllers
have reconciled the routes at leas once (i.e. are aware of their
existence)

No

## Acceptance Steps
N/A

### DIFF
--- a/controllers/controllers/networking/cfdomain_controller_test.go
+++ b/controllers/controllers/networking/cfdomain_controller_test.go
@@ -10,6 +10,7 @@ import (
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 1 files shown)


gcapizzi (approved):

gcapizzi:
@danail-branekov is this something that could occur in production somehow?

danail-branekov:
> @danail-branekov is this something that could occur in production somehow?

Probably - the domain finalizer is doing a best effort to delete orphan routes. I think that we already agreed that we would accept potential orphan routes when e.g. an app is deleted.

However, IRL it is not really likely that you create a route and immediately (like in a script) delete its domain

PR #2117, opened 2023-01-27
Implement DELETE /v3/roles/guid

georgethebeatle:
https://github.com/cloudfoundry/korifi/issues/1239

## What is this change about?
Implement DELETE /v3/role/{guid}

Also fix listing functionality to read the role guid that gets returned
to the user from the CFRole metadata, as the CFRole name contains the
sha256 sum of the role name and the user name, so that optimistic
locking will prevent us from creating duplicate entries

No

## Acceptance Steps
See story

@eirini

### DIFF
--- a/api/handlers/fake/cfrole_repository.go
+++ b/api/handlers/fake/cfrole_repository.go
@@ -26,6 +26,34 @@ type CFRoleRepository struct {
 		result1 repositories.RoleRecord
 		result2 error
 	}
+	DeleteRoleStub        func(context.Context, authorization.Info, repositories.DeleteRoleMessage) error
+	deleteRoleMutex       sync.RWMutex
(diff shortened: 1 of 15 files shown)


api/repositories/namespace_retriever.go
```diff
@@ -82,6 +82,12 @@ var (
 		Resource: "cftasks",
 	}
 
+	RoleBindingGVR = schema.GroupVersionResource{
```
danail-branekov, line 85: I wonder whether we need to register role bindings with the namespace retriever. Initially I though we should, but then I think the repository ended up not using it

danail-branekov: yes, it turned out it is not needed

danail-branekov (approved):

Issue #1661, opened 2022-09-19
Support multi-app manifests


danail-branekov:
The current manifest action only considers the first application when applying
the manifest. There is even a
[validation](https://github.com/cloudfoundry/korifi/blob/afcd2ef983426b68739950967da0fe71f6a92199/api/payloads/manifest.go#L13)
that returns an error if the manifest contains more than one application.

On CF on VMs multiple app mnifests work OK. We should therefore remove this
validation and change the manifest action to process all the applications in
the manifest.

PR #2118, opened 2023-01-27
Allow multiple apps in manifests

kieron-dev:
#1661

## What is this change about?
Allow the list of apps in a space manifest to have more than 1 entry.

Also fix the behaviour of creating a web process during manifest application or app creation to match cf on VMs. I.e. the web process is created at the same time as the app.

Note that the web app is created earlier now, rather than when starting the app.

## Acceptance Steps
See story

@gcapizzi

### DIFF
--- a/api/actions/manifest.go
+++ b/api/actions/manifest.go
@@ -45,20 +45,25 @@ func NewManifest(domainRepo shared.CFDomainRepository, defaultDomainName string,
 	}
 }
 
-func (a *Manifest) Apply(ctx context.Context, authInfo authorization.Info, spaceGUID string, manifest payloads.Manifest) error {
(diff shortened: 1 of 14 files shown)


api/actions/manifest.go
```diff
@@ -45,20 +45,25 @@ func NewManifest(domainRepo shared.CFDomainRepository, defaultDomainName string,
 	}
 }
 
-func (a *Manifest) Apply(ctx context.Context, authInfo authorization.Info, spaceGUID string, manifest payloads.Manifest) error {
+func (a *Manifest) Apply(ctx context.Context, authInfo authorization.Info, spaceGUID string, manifesto payloads.Manifest) error {
```
georgethebeatle, line 48: Hmm, was that intentional? Not a big deal, just curious.

georgethebeatle (approved):

PR #2124, opened 2023-01-31
switched to `tr` because MacOSX base64 doesn’t support the` -w 0` option

eghobo:

### DIFF
--- a/scripts/account-creation.sh
+++ b/scripts/account-creation.sh
@@ -12,13 +12,13 @@ if [[ -z "${E2E_USER_NAME:=}" ]]; then
   export E2E_USER_NAME="e2e-cert-user"
   createCert "$E2E_USER_NAME" "$tmp/key.pem" "$tmp/cert.pem"
 
-  export E2E_USER_PEM="$(cat $tmp/cert.pem $tmp/key.pem | base64 -w0)"
+  export E2E_USER_PEM="$(cat $tmp/cert.pem $tmp/key.pem | base64 | tr -d '\n\r')"
 fi
 if [[ -z "${E2E_LONGCERT_USER_NAME:=}" ]]; then
   export E2E_LONGCERT_USER_NAME="e2e-longcert-user"
   createCert "$E2E_LONGCERT_USER_NAME" "$tmp/longkey.pem" "$tmp/longcert.pem" "365"
 
-  export E2E_LONGCERT_USER_PEM="$(cat $tmp/longcert.pem $tmp/longkey.pem | base64 -w0)"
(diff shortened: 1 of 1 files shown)


danail-branekov:
For context, the `-w0` option is supported with `base64` that is coming with [coreutils](https://www.gnu.org/software/coreutils/), `brew install coreutils` if brew is used.

eghobo:
actually, coreutils Homebrew package doesn't include base64, https://github.com/Homebrew/homebrew-core/blob/master/Formula/coreutils.rb#L70

but even if you install base64, separately it still doesn't support `w`

```
which base64

$ /opt/homebrew/bin/base64

cat HACKING.md | /opt/homebrew/bin/base64 -w 0

$ /opt/homebrew/bin/base64: illegal option -- w
$ base64  --  Encode/decode file as base64.  Call:
$            base64 [-e / -d] [options] [infile] [outfile]
$
$Options:
$           --copyright       Print copyright information
$           -d, --decode      Decode base64 encoded file
$          -e, --encode      Encode file into base64
$           -n, --noerrcheck  Ignore errors when decoding
$           -u, --help        Print this message
$          --version         Print version number
$
$by John Walker
$http://www.fourmilab.ch/
```

danail-branekov:
On my mac

```
❯ which base64
/opt/homebrew/opt/coreutils/libexec/gnubin/base64

❯ cat HACKING.md | $(which base64) -w 0
// succeeds and prints some reasonable output
```

Then I tried 

```
❯ ls -la $(which base64)
lrwxr-xr-x 1 i030720 17 Apr 16  2022 /opt/homebrew/opt/coreutils/libexec/gnubin/base64 -> ../../bin/gbase64
```

Do you have the `gbase64` binary next to the `base64` one? Do you have the `/opt/homebrew/opt/coreutils/libexec/gnubin/base64` symlink? 

I can also see that the first entry in my `PATH` is `/opt/homebrew/opt/coreutils/libexec/gnubin`. I wonder whether this is something brew did for me...

georgethebeatle:
@eghobo Korifi is intended to be used with the gnu versions of tools like `sed`, `base64`, `bash` etc.

You can install them quite easily on a mac by running

```
brew install coreutils
```

Once you do that you can do one of two things:
1. Add coreutils to the PATH. In this case you will be using gnu tooling for all your projects
2. Add coreutils to the PATH in the `korifi/.envrc` file. In this case you will be using gnu tooling only for korifi

Note that in the second case you will have to have `direnv` installed, which you can do by running:

```
brew install direnv
```

And you will have to add the following line to your `.bashrc`:

```
eval "$(direnv hook zsh)"
```

eghobo:
@danail-branekov @georgethebeatle yes, it was a PATH issue `PATH="/opt/homebrew/opt/coreutils/libexec/gnubin:$PATH"` fixed the problem. I am closing this PR; sorry for the noise ;)

PR #2137, opened 2023-01-31
Update api.md for "create a role"

julian-hj:
no

## What is this change about?
add `organization_manager` to the list of role types we support in api.md.

no.

## Acceptance Steps
- create an org `o`
- `cf -v set-org-role some-user o OrgManager`
- observe that we are posting to the endpoint with `"type": "organization_manager"`

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -225,7 +225,7 @@ This endpoint is fully supported.
 
 #### Supported parameters:
 
--   `type` (the only supported value is `space_developer`
+-   `type` (the only supported values are `space_developer` and `space_manager`)
 -   `relationships.user`
 -   `relationships.organization`
 -   `relationships.space`


gcapizzi:
I don't think `organization_manager` is a supported role, [permissions](https://github.com/cloudfoundry/korifi/blob/38f10451f4cde1a2361be21873d7c6034aecde72/helm/korifi/controllers/cf_roles/cf_org_manager.yaml) look completely random 😅

julian-hj:
> I don't think `organization_manager` is a supported role, [permissions](https://github.com/cloudfoundry/korifi/blob/38f10451f4cde1a2361be21873d7c6034aecde72/helm/korifi/controllers/cf_roles/cf_org_manager.yaml) look completely random 😅

Oh, right. I was remembering that we had a set of tests where we use the manager role to create a "read-only" user that can pretty much only view apps but cannot modify them, but I misremembered which one it was. I've changed the PR to specify `space_manager` instead, but maybe even that isn't fully supported? We have a [number of tests that use it](https://github.com/cloudfoundry/korifi/search?q=space_manager), so it seems a bit safer...

kieron-dev (commented):
The only valid role name for create role at the moment is space_developer. We have a wider set of roles [declared in the system](https://github.com/cloudfoundry/korifi/blob/main/helm/korifi/controllers/cf_roles), but only admin and space_developer have the correct permissions (and the API doesn't allow creating admin roles, as that was previously done directly in UAA). The others have had random permissions added to make certain tests feel more realistic, but are neither complete, tested, nor necessarily correct. There are also some missing from the full set [described in the API docs](https://v3-apidocs.cloudfoundry.org/version/3.130.0/index.html#valid-role-types), and we're missing the admin_read_only and global_auditor roles [described in other docs](https://docs.cloudfoundry.org/concepts/roles.html).

I think it is a suggested epic to flesh out the other roles, but I don't think we even have stories for that yet. @gcapizzi might know.

So this doc is correct as it is, and shouldn't be changed at the moment. Maybe we should look at enforcing space_developer as the only valid value at the API level to avoid users believing the other roles are valid currently.

PR #2141, opened 2023-02-01
Reflect the cluster builder ready status in the builder info status

kieron-dev:
No

## What is this change about?
The builder info ready status condition should be true iff the cluster builder ready status condition is true

No


davewalter (commented):
LGTM

davewalter (approved):

Issue #2154, opened 2023-02-03
Do not pin go version in github actions or figure out a way to bump the go version there


kieron-dev:
It looks like 'stable' will work: https://github.com/actions/setup-go#using-stableoldstable-aliases

PR #2155, opened 2023-02-03
Use 'stable' for the go-version in github actions

kieron-dev:
#2154

## What is this change about?
That means we don't have to manually bump the go version in .github/workflows/test-pr.yml each time the minor version changes

No


kieron-dev:
If you look at the `actions/setup-go` step of https://github.com/cloudfoundry/korifi/actions/runs/4083019158/jobs/7038047974 for example, you'll see go 1.20 gets used. So this looks fine.

gcapizzi (approved):

PR #2178, opened 2023-02-14
Move masking of the Forbidden error to the handler

danail-branekov:
No

## What is this change about?
Error masking is usually done in the HTTP handlers rather than in the
repositories.

No

## Acceptance Steps
N/A - no functional change

@georgethebeatle


georgethebeatle (approved):

danail-branekov:
It turned out that we cannot merge this change as masking the forbidden error in the repo on get has the purpose of space managers (that are allowed to get service bindings, but not delete them) get forbidden errors rather than not found. This has been expoxed by an [e2e test](https://github.com/cloudfoundry/korifi/blob/e84dc03fa442fd91123312fb38ddd2f72ddb96d0/tests/e2e/service_bindings_test.go#L56) failure

Issue #1925, opened 2022-11-21
[Feature]: Operator can make dynamic changes to log level


julian-hj:
### Blockers/Dependencies

#1836

### Background

**As an** Operator
**I want** to control the verbosity of my component logging
**So that** I can get detailed logging only when I need it

### Acceptance Criteria

**GIVEN** a cluster with all the prerequisites installed
**WHEN I** install Korifi
**THEN I** can set a log level that gets passed to the installed components
**AND**
**WHEN I** edit the log level of a running component by editing its configmap
**THEN I** can see that the log output adjusts without having to restart the deployment.

### Dev Notes

See comments on #1836 for details and a spike branch where this mechanism was implemented.

We don't currently have any debug logging implemented, so it probably makes sense to add at least one debug message at startup in order to know whether debug messages are being filtered.

julian-hj:
I ran through a/c. It works! Very delightful.

PR #2181, opened 2023-02-15
Add config params to configure log level

clintyoshimura:
#1925

## What is this change about?
Adds config params that allows the api and controller log levels to be changed on the fly. The default log levels are info.

no

## Acceptance Steps
See issue

@davewalter


controllers/config/config.go
```diff
@@ -33,6 +35,9 @@ type ControllerConfig struct {
 	BuilderServiceAccount     string `yaml:"builderServiceAccount"`
 	ContainerRepositoryPrefix string `yaml:"containerRepositoryPrefix"`
 	ContainerRegistryType     string `yaml:"containerRegistryType"`
+
+	// log level
```
danail-branekov, line 39: I do not think we need this comment, the code itself is self-explanatory

tools/zap_logger.go
```diff
@@ -0,0 +1,95 @@
...
+	cfg.EncoderConfig.EncodeTime = zapcore.RFC3339NanoTimeEncoder
+	cfg.Level.SetLevel(logLevel)
+	logger, err := cfg.Build()
+	if err != nil {
+		return logr.Logger{}, zap.AtomicLevel{}, err
+	}
+	return zapr.NewLogger(logger), cfg.Level, nil
+}
+
+func WatchLoggingConfig(ctx context.Context, watcher *fsnotify.Watcher, configFilepath string, logger logr.Logger, atomicLevel zap.AtomicLevel, getCfgLogLevelFunc func(string) (zapcore.Level, error)) error {
```
danail-branekov, line 26: This function is quite complex, codeclimate claims that it has 6 arguments and cognitive complexity of 36. Here are some refactoring ideas that might help:

* refactor the function to return a channel that a log level is being sent on an update. This would probably reduce the number of arguments (as `atomicLevel` and `getCfgLogLevelFunc` would be moved into the called).
* extract an utility that takes care about re-registering the watch upon path being deleted so that path handling is transparent to this function. This might be just a function that creates the watcher, setup a go routine to track path being deleted and return the watcher's `Event` channel

danail-branekov:
Besides the notes I left above, I do not see the feature being really tested in an integration manner.I wonder how expensive it would be to have a suite that changes the log level and verifies by looking at the logs whether logs reflect the new value.

davewalter:
Thanks for the thoughts/suggestions Danail. We weren't happy with where we left it last night, but wanted to get feedback from CI to see what failed.

tools/zap_logger.go
```diff
@@ -0,0 +1,95 @@
...
+			case event, ok := <-watcher.Events:
+				// Channel is closed.
+				if !ok {
+					return
+				}
+
+				logger.V(1).Info("logger level config event", "event", event)
+
+				// Only care about events which may modify the contents of the file.
+				if !(isWrite(event) || isRemove(event) || isCreate(event)) {
```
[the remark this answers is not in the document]
clintyoshimura: Fixed

tools/zap_logger.go
clintyoshimura: Refactored this function. Code complexity is better, but still higher than codeclimate would like.

clintyoshimura:
> Besides the notes I left above, I do not see the feature being really tested in an integration manner.I wonder how expensive it would be to have a suite that changes the log level and verifies by looking at the logs whether logs reflect the new value.

We think it would be difficult to add an e2e test since they don't interact directly with k8s. However, we are adding some tests which verify the functionality of the directory watching component. We will also add tests for the functions that change the log level.

georgethebeatle:
I was wondering if you considered using something like this: https://github.com/spf13/viper#watching-and-re-reading-config-files

It uses fsnotify as well, but seems to handler the low level details of event detection and transmission.

akrishna90 (approved):

davewalter:
> I was wondering if you considered using something like this: https://github.com/spf13/viper#watching-and-re-reading-config-files
> 
> It uses fsnotify as well, but seems to handle the low level details of event detection and transmission.

Looks interesting. We've pulled this out into a separate chore.

PR #2199, opened 2023-02-17
fix/adapt scripts to work on mac

wanddynosios:
https://github.com/cloudfoundry/korifi/issues/2108

## What is this change about?
I ran the `deploy-on-kind.sh` on my Mac and ran into a couple of issues that this PR adresses:
* kbld has a [reported bug](https://github.com/carvel-dev/kbld/issues/213) where it does not work properly on machines set to another language. Setting `LC_ALL=en_US.UTF-8` in the script fixes this.
* If `registry_configuration` is an empty array, I get a "unbound variable" so I added a default empty value. As array elements are interpreted as if in single quotes, `helm` thinks it is getting one argument too many. Removing the double quotes fixes this.
* The Mac's default bash version is unfortunately 3.2., so it does not know the `[[ -v VAR_NAME ]]` syntax. Comparing via `== "true"` should work on both newer and older machines.

no

### DIFF
--- a/scripts/deploy-on-kind.sh
+++ b/scripts/deploy-on-kind.sh
@@ -5,6 +5,10 @@ set -euo pipefail
 ROOT_DIR="$(cd "$(dirname "$0")/.." && pwd)"
 SCRIPT_DIR="${ROOT_DIR}/scripts"
 
+# workaround for https://github.com/carvel-dev/kbld/issues/213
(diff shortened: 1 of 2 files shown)


gcapizzi (approved):

gcapizzi:
Thank you @wanddynosios!

PR #2200, opened 2023-02-17
Use non-zipped assets for e2e tests

matt-royal:

This was originally part of #1146, but the meat of that ended up being done in a different PR by @kieron-dev.

## What is this change about?

Rather than committing zip files into our repo, we now commit the contents of those zips and zip them up when the e2e tests run. This allows us to more easily track changes to the test apps, and also to explicitly track the source of these apps (via vendir) where appropriate.

No

## Acceptance Steps

The e2e tests should continue to pass

@kieron-dev


tests/vendir.yml
```diff
@@ -36,13 +36,29 @@ directories:
...
+- path: e2e/assets/dora
+  contents:
+  - path: .
+    git:
+      url: https://github.com/cloudfoundry/cf-acceptance-tests.git
+      depth: 1
+      refSelection:
+        semver:
+          constraints: ">=10.0.0"
     newRootPath: assets/dora
```
tcdowney, line 62: `e2e/assets` contains a mix of non-vendored and vendored apps now, right? I feel like it might get confusing if someone makes a change to `node` or `dora` (e.g. if we add endpoints that look at service binding files) and doesn't know these are actually managed through `vendir`.

tcdowney: I liked that the original path made it was clear it was vendored... I am wondering if we could put this in a subdir e.g. `e2e/assets/vendored/dora` or some other convention involving the directory name or something. I just think the `vendir.yml` is pretty far removed from where these files actually live and it could be confusing.

kieron-dev:
@matt-royal The zip approach looks good to me

tests/vendir.yml
tcdowney: `e2e/assets/vendored/dora` LGTM, thanks!

tcdowney (approved):
LGTM!

PR #2222, opened 2023-02-23
Make procfile app output configurable

davewalter:
#2216 

## What is this change about?
When we restored the ability for users to provide their own app for the majority of the e2e tests, we ran into a problem where the expected output does not always match the output from the default procfile app. This PR introduces a new `APP_BITS_OUTPUT` env var to allow users to specify the expected output for the test that cares about it.

No.

## Acceptance Steps
Run the e2e tests against an environment without a procfile buildpack and see that they pass if `APP_BITS_PATH` is set to another app and `APP_BITS_OUTPUT` is set to the expected output.

@acosta11 @Birdrock

### DIFF
--- a/tests/e2e/routes_test.go
+++ b/tests/e2e/routes_test.go
@@ -4,6 +4,7 @@ import (
 	"crypto/tls"
 	"fmt"
 	"net/http"
+	"os"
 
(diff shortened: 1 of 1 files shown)


acosta11 (approved):
LGTM

Issue #2174, opened 2023-02-09
[Bug]: `cf buildpacks` should return a better error message when builder doesn't exist


akrishna90:
### What happened?

WHEN I change the configmap for the korifi-api to set `builderName` to `no-such-builder`
AND I run cf buildpacks
THEN I see

```
$ cf buildpacks
Getting buildpacks as cf-admin...

Unexpected Response
Response Code: 500
Code: 0, Title: , Detail: {"errors":[{"detail":"An unknown error occurred.","title":"UnknownError","code":10001}]}
```

### What you expected to happen

A better error message explaining that the builder doesn't exist. 

#### Builder doesn't exist
See 3rd acceptance criteria on #1529 

WHEN I change the configmap for the korifi-api to set `builderName` to `no-such-builder`
AND I run cf buildpacks
THEN I should see an error which explains that no valid builder with that name exists

#### Builder not ready

WHEN the builder specified in the korifi-api's configMap
AND I run cf buildpacks
THEN I should see either an empty list of buildpacks or an error that is more meaningful than UnknownError (it is probably already doing this, but let's validate)

If the builder isn't ready, return a more meaningful error message indicating that the builder isn't ready.
If the build is ready but the buildpack list is empty, it's fine to return an empty list.

### How to reproduce it (as minimally and precisely as possible)

See "**What happened?**" section above for steps to reproduce

### Anything else we need to know?

NA

### Environment

Revision of codebase: latest on main
Kubernetes version (use `kubectl version`):
Cloud provider or hardware configuration: Kind
Others:

PR #2250, opened 2023-03-03
Improve "cf buildpacks" error responses

davewalter:
#2174 

## What is this change about?
This PR improves the error responses from the `/v3/buildpacks` endpoint used by the "cf buildpacks" command.

- Wrap the error returned when the BuilderInfo resource cannot be retrieved in a ResourceNotFound apierror.
- Check the status of the BuilderInfo resource, and return the message if it isn't ready.
- Include the ClusterBuilder ready condition message in the BuilderInfo condition message if it is not ready.

No

## Acceptance Steps
See #2174 

@clintyoshimura @tcdowney

### DIFF
--- a/api/errors/errors.go
+++ b/api/errors/errors.go
@@ -298,15 +298,15 @@ func NewBlobstoreUnavailableError(cause error) BlobstoreUnavailableError {
 	}
 }
 
-type CFCreateError struct {
+type ResourceNotReadyError struct {
 	apiError
 }
 
-func NewCFCreateError(cause error) CFCreateError {
-	return CFCreateError{
+func NewResourceNotReadyError(cause error) ResourceNotReadyError {
+	return ResourceNotReadyError{
 		apiError: apiError{
 			cause:      cause,
-			title:      "CF-Create",
+			title:      "CF-ResourceNotReady",
 			detail:     cause.Error(),
 			code:       420000,
 			httpStatus: http.StatusInternalServerError,
--- a/api/repositories/buildpack_repository.go
+++ b/api/repositories/buildpack_repository.go
@@ -8,6 +8,7 @@ import (
 	apierrors "code.cloudfoundry.org/korifi/api/errors"
 	"code.cloudfoundry.org/korifi/controllers/api/v1alpha1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/types"
 )
 
@@ -56,9 +57,25 @@ func (r *BuildpackRepository) ListBuildpacks(ctx context.Context, authInfo autho
 	)
 	if err != nil {
 		if errors.IsNotFound(err) {
-			return nil, fmt.Errorf("no BuilderInfo %q resource found in %q namespace", r.builderName, r.rootNamespace)
+			return nil, apierrors.NewResourceNotReadyError(fmt.Errorf("BuilderInfo %q not found in namespace %q", r.builderName, r.rootNamespace))
 		}
+
 		return nil, apierrors.FromK8sError(err, BuildpackResourceType)
+	}
+
+	if !meta.IsStatusConditionTrue(builderInfo.Status.Conditions, StatusConditionReady) {
+		var conditionNotReadyMessage string
+
+		readyCondition := meta.FindStatusCondition(builderInfo.Status.Conditions, StatusConditionReady)
+		if readyCondition != nil {
+			conditionNotReadyMessage = readyCondition.Message
+		}
+
+		if conditionNotReadyMessage == "" {
+			conditionNotReadyMessage = "resource not reconciled"
+		}
+
+		return nil, apierrors.NewResourceNotReadyError(fmt.Errorf("BuilderInfo %q not ready: %s", r.builderName, conditionNotReadyMessage))
 	}
 
 	return builderInfoToBuildpackRecords(builderInfo), nil
--- a/api/repositories/buildpack_repository_test.go
+++ b/api/repositories/buildpack_repository_test.go
@@ -62,7 +62,7 @@ var _ = Describe("BuildpackRepository", func() {
 		When("no build reconcilers exist", func() {
 			It("errors", func() {
 				_, err := buildpackRepo.ListBuildpacks(context.Background(), authInfo)
-				Expect(err).To(MatchError(ContainSubstring(fmt.Sprintf("no BuilderInfo %q resource found in %q namespace", builderName, rootNamespace))))
+				Expect(err).To(MatchError(ContainSubstring(fmt.Sprintf("BuilderInfo %q not found in namespace %q", builderName, rootNamespace))))
 			})
 		})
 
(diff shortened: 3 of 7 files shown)


api/errors/errors.go
```diff
@@ -298,22 +298,38 @@ func NewBlobstoreUnavailableError(cause error) BlobstoreUnavailableError {
...
 }
 
-func NewCFCreateError(cause error) CFCreateError {
-	return CFCreateError{
+func NewResourceNotFoundError(cause error) ResourceNotFoundError {
+	return ResourceNotFoundError{
 		apiError: apiError{
 			cause:      cause,
-			title:      "CF-Create",
+			title:      "ResourceNotFound",
```
clintyoshimura, line 309: Is it ok not to prefix these with "CF-"?

tcdowney, line 309: There is a chance this causes some unexpected behavior since clients might assume `ResourceNotFound` is a `404` (based on its name and [this existing error](https://github.com/cloudfoundry/cloud_controller_ng/blob/main/errors/v2.yml#L78)). I haven't evaluated this code, but a quick search shows the CLI specifically looking for it: https://github.com/search?q=repo%3Acloudfoundry%2Fcli%20ResourceNotFound&type=code

This might be one of the cases where it is worth making a unique error title to be safe.

api/errors/errors.go
```diff
@@ -298,22 +298,38 @@ func NewBlobstoreUnavailableError(cause error) BlobstoreUnavailableError {
...
+	apiError
+}
+
+func NewResourceNotReadyError(cause error) ResourceNotReadyError {
+	return ResourceNotReadyError{
+		apiError: apiError{
+			cause:      cause,
+			title:      "ResourceNotReady",
+			detail:     cause.Error(),
+			code:       420001,
```
clintyoshimura, line 327: Do we need to amend the other PR with this new code as well?

api/errors/errors.go
tcdowney: I think in general CF assumes a unique error `title` per `httpStatus`/`code`.

api/repositories/buildpack_repository.go
```diff
@@ -56,11 +57,22 @@ func (r *BuildpackRepository) ListBuildpacks(ctx context.Context, authInfo autho
...
 		}
+
 		return nil, apierrors.FromK8sError(err, BuildpackResourceType)
 	}
 
+	if !meta.IsStatusConditionTrue(builderInfo.Status.Conditions, StatusConditionReady) {
+		conditionNotReadyMessage := "Resource not reconciled"
+		readyCondition := meta.FindStatusCondition(builderInfo.Status.Conditions, StatusConditionReady)
+		if readyCondition != nil {
+			conditionNotReadyMessage = readyCondition.Message
```
clintyoshimura, line 70: We should probably check if the condition's message is empty before overwriting the default message

api/repositories/buildpack_repository_test.go
```diff
@@ -78,7 +78,29 @@ var _ = Describe("BuildpackRepository", func() {
 
 			It("errors", func() {
 				_, err := buildpackRepo.ListBuildpacks(context.Background(), authInfo)
-				Expect(err).To(MatchError(ContainSubstring(fmt.Sprintf("no BuilderInfo %q resource found in %q namespace", builderName, rootNamespace))))
+				Expect(err).To(MatchError(ContainSubstring(fmt.Sprintf("BuilderInfo %q not found in namespace %q", builderName, rootNamespace))))
+			})
+		})
+
+		When("the build reconciler with the configured BuilderName is not ready", func() {
```
clintyoshimura, line 85: Should we add a when case for the default error message?

api/errors/errors.go
tcdowney: In standup we chatted a bit about this. I don't think we do since it's questionable whether or not `v2.yml` is even meant for that purpose and we've already "reserved" the `420000` block with https://github.com/cloudfoundry/cloud_controller_ng/pull/3205. I think it's worth having that convo there though.

api/errors/errors.go
tcdowney: Ah geez, you're probably right that we want that prefix. I forgot that Cloud Controller prefixes everything with CF- fairly late in the game:

* https://github.com/cloudfoundry/cloud_controller_ng/blob/5614bf7bc92d0e8a91a06fad569b7b738b27e827/app/presenters/v3_error_hasher.rb#L65
* https://github.com/cloudfoundry/cloud_controller_ng/blob/main/app/presenters/v2_error_hasher.rb

clintyoshimura:
Weren't we also saying we have to copy the ready condition message from the ClusterBuilder object when creating a BuilderInfo in the builderinfo_controller.go since otherwise it just has a generic "ClusterBuilder x is not ready" message? Or did we decide not to do that?

api/errors/errors.go
davewalter: Does it make sense to combine this with the `ResourceNotReady` error to avoid any special handling by the CLI? I was also thinking that we don't really need both errors since they are identical in how the handle the underlying error.

tcdowney: Yeah, I think `ResourceNotReady` would work fine. 👍 
There may be a subtle distinction between something that could eventually become ready and something that will never become ready, but for these errors I don't think that matters.

api/errors/errors.go
davewalter: Based on the discussion above, I have consolidated the two errors down to just the `ResourceNotReady` error, with a code of `420000` so I think we are good.

api/errors/errors.go
davewalter: Updated.

davewalter: I've removed the `ResourceNotFound` error and just used the `ResourceNotReady` error everywhere.

api/repositories/buildpack_repository.go
davewalter: Fixed.

api/repositories/buildpack_repository_test.go
davewalter: I've added a new test to cover this.

davewalter:
> Weren't we also saying we have to copy the ready condition message from the ClusterBuilder object when creating a BuilderInfo in the builderinfo_controller.go since otherwise it just has a generic "ClusterBuilder x is not ready" message? Or did we decide not to do that?

The change is included in this PR. I've updated some tests to be more explicit about the status condition messages for the `BuilderInfo` resource.

kpack-image-builder/controllers/builderinfo_controller.go
```diff
@@ -110,7 +110,7 @@ func (r *BuilderInfoReconciler) ReconcileResource(ctx context.Context, info *kor
 			Type:    ReadyConditionType,
 			Status:  metav1.ConditionFalse,
 			Reason:  "ClusterBuilderNotReady",
-			Message: fmt.Sprintf("ClusterBuilder %q is not ready", r.clusterBuilderName),
+			Message: fmt.Sprintf("ClusterBuilder %q is not ready: %s", r.clusterBuilderName, clusterBuilderReadyCondition.Message),
```
clintyoshimura, line 113: Since this statement is in an else clause with clusterBuilderReadyCondition != nil and clusterBuilderReadyCondition.Status == corev1.ConditionTrue, it's possible that clusterBuilderReadyCondition could be nil here right? So we may have to guard against that when using clusterBuilderReadyCondition.Message.

davewalter: Good catch. I've added a test for this and updated the code to detect whether there is a ready condition on the ClusterBuilder.

clintyoshimura (approved):
LGTM

tcdowney (approved):

PR #2252, opened 2023-03-04
Remove Ginkgo from tools/tools.go

davewalter:
No

## What is this change about?
As of [Ginkgo v2.8.2](https://github.com/onsi/ginkgo/releases/tag/v2.8.2), "Ginkgo now includes a tools.go file in the root directory of the ginkgo package. This should allow modules that simply go get github.com/onsi/ginkgo/v2 to also pull in the CLI dependencies. This obviates the need for consumers of Ginkgo to have their own tools.go file and makes it simpler to ensure that the version of the ginkgo CLI being used matches the version of the library. You can simply run go run github.com/onsi/ginkgo/v2/ginkgo to run the version of the cli associated with your package go.mod."

No

## Acceptance Steps
Everything still works.

### DIFF
--- a/tools/tools.go
+++ b/tools/tools.go
@@ -5,7 +5,6 @@ package tools
 
 import (
 	_ "github.com/maxbrunsfeld/counterfeiter/v6"
-	_ "github.com/onsi/ginkgo/v2/ginkgo"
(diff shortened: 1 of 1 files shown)


kieron-dev (approved):
Good spot

Issue #2166, opened 2023-02-07
[Feature]: Developer can see consistent "reason"s on Korifi resources when something bad happens


julian-hj:
### Background

**As a** Developer
**I want** consistent and meaningful "reasons" on k8s resources when there is a failure
**So that** I can debug more easily

### Acceptance Criteria

**GIVEN** it's another day in the office
**WHEN I** inspect the Korifi code
**THEN I** see that we are assigning consistent  and meaningful "reason"s to failure events, in PascalCase

### Dev Notes

Specifically:
* cf_app_controller puts an erroneous "appStaged" reason on an unstaged app: https://github.com/cloudfoundry/korifi/blob/main/controllers/controllers/workloads/cfapp_controller.go#L91
 * cf_app controller puts an erroneous "unimplemented" reason on a condition: https://github.com/cloudfoundry/korifi/blob/main/controllers/controllers/workloads/cfapp_controller.go#L98
 * cfbuild_controller uses a "BuildWorkload" reason and message everywhere which has no meaning. Message is optional and could be left blank. Reason should have a better meaning. Also we should be case consistent. https://github.com/cloudfoundry/korifi/blob/main/controllers/controllers/workloads/cfbuild_controller.go#L119
 * Lots of places have "reason" in lowerCamelCase, but we should standardize on PascalCase (CamelCase with initial letter capitalized).
* [Kubernetes API Conventions doc that discusses Status conditions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties)

matt-royal:
In reviewing `cf_app_controller` we made a few observations:
  - `Staged` is a misleading condition, because it actually indicates that a valid droplet is assigned to the `App`. That droplet might have just been staged, or it might be an old droplet that was reassigned to the App. It would be more accurate to say that the App is `Ready` instead. An app would start as `Ready` false and then become `Ready` true when a valid droplet ref is assigned.
  
  - `Running` is also misleading. The actual place to verify that the App processes are running is to look at the CFProcess objects or the pods that they scheduled. What we're really indicating here is that the user has assigned a droplet and the app has a `DesiredState` of `Started`. It would be clearer if we communicated this by setting an `ObservedGeneration` field on the status. An app *should* be running when it is `Ready`, has its `DesiredState` set to `Running`, and has a `Status.ObservedGeneration` equal to `Metadata.Generation`.

PR #2256, opened 2023-03-07
Improve `reason` on korifi resources

akrishna90:

[#2166]

## What is this change about?
Improved `reason` on korifi resources

- Removed `Running` status condition from `cfApp`. The actual place to verify that the App processes are running is to look at the CFProcess objects or the pods that they scheduled.
- Renamed `Staged` status condition on `cfApp` to `Ready`. `Ready` actually indicates that a valid droplet is assigned to the App. That droplet might have just been staged, or it might be an old droplet that was reassigned to the App. It would be more accurate to say that the App is Ready instead. An app would start as Ready false and then become Ready true when a valid droplet ref is assigned.
- Updated `cfbuild` resource to have `BuildRunning` & `BuildNotRunnning` as reasons for `Staging` condition. `BuildFailed` & `BuildSucceeded` for `Succeeded` status condition
- Updated `cftask` & `TaskWorkload` reasons to match CamelCase
- Removed redundant `Message` attribute wherever applicable.
- When the `job` created by the `taskworkload` fails, the `status.condition.reason` on the taskWorkload is derived from job's container status. We observed that it follows the convention of using CamelCase.

Yes. 
- `reason` on all of `cfapp`, `cfbuild`, `cftask` & `taskWorkload` have been changed.
- `cfapp` now has a new `Ready` status condition and will no longer have `Staging` & `Running` conditions. When existing cfapps reconciles it would get the `Ready` status condition but also keep their `Staging` & `Running` conditions with the same status, reason and message. 

## Acceptance Steps

See #2166

@matt-royal

### DIFF
--- a/api/repositories/app_repository.go
+++ b/api/repositories/app_repository.go
@@ -499,7 +499,7 @@ func (f *AppRepo) SetCurrentDroplet(ctx context.Context, authInfo authorization.
 		return CurrentDropletRecord{}, fmt.Errorf("failed to set app droplet: %w", apierrors.FromK8sError(err, AppResourceType))
 	}
 
-	_, err = f.appConditionAwaiter.AwaitCondition(ctx, userClient, cfApp, workloads.StatusConditionStaged)
+	_, err = f.appConditionAwaiter.AwaitCondition(ctx, userClient, cfApp, workloads.StatusConditionReady)
 	if err != nil {
 		return CurrentDropletRecord{}, fmt.Errorf("failed to await the app staged condition: %w", apierrors.FromK8sError(err, AppResourceType))
 	}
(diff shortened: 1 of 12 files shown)


matt-royal:
Discussion topic: Should we add code to the cfapp reconciler to explicitly remove the `Staging` and `Running` conditions from existing cfapps? Without this addition, old cfapps will retain their last `Staging` and `Running` statuses forever, no matter how the app changes.

gcapizzi (approved):

kieron-dev:
> Discussion topic: Should we add code to the cfapp reconciler to explicitly remove the `Staging` and `Running` conditions from existing cfapps? Without this addition, old cfapps will retain their last `Staging` and `Running` statuses forever, no matter how the app changes.

First question: do we have anybody running this for real? If not, pragmatically there's no migration required.

If we do, then I'm reluctant about loading migration code into reconcilers where it will remain running forever (and will build up over time). We could provide a script to do it less intrusively, and tell operators to run that after this particular upgrade. Or we could begin a migration framework to automatically handle this sort of thing which could run as a k8s job.

kieron-dev (approved):
Happy with these

matt-royal:
@kieron-dev I agree. Let's make the manual script if and when users mention that this is a problem and leave it as is for now. Merging.

Issue #2228, opened 2023-02-28
[Feature]: Container images are deleted alongside packages


gcapizzi:
### Background

**As an** operator
**I want** to only store container images for packages currently used by the system
**So that** I don't face increasing costs of storage for running the same amount of apps

### Acceptance Criteria

**GIVEN** I have a package
**AND** I have uploaded sources for it
**WHEN I** delete the package
**THEN I** see the corresponding container image containing the source has been deleted

### Dev Notes

See https://github.com/cloudfoundry/korifi/issues/2112#issuecomment-1440501732 for details.

PR #2257, opened 2023-03-07
Container images are deleted alongside packages

kieron-dev:
#2228

## What is this change about?
- Introduce image.Client
- Delete package images when deleting CFPackage

No

## Acceptance Steps
Push an app.
See its package image exists in a registry
Delete the app.
See the image disappear

@gcapizzi


kieron-dev:
Converting to draft as we seem to have dangling tags

controllers/controllers/workloads/cfpackage_controller_test.go
```diff
@@ -57,4 +60,59 @@ var _ = Describe("CFPackageReconciler Integration Tests", func() {
...
+				}).Should(Succeed())
+			})
+		})
+
+		When("deletion fails", func() {
+			BeforeEach(func() {
+				imageDeleter.DeleteReturns(errors.New("oops"))
+			})
+
+			It("ignores the errors and finishes finalization", func() {
```
tcdowney, line 111: Cool I think this is good. I was worried about an external resource outage blocking finalization. Any ideas around how to make this scenario easier for operators to detect though? It reminds me of the "orphaned blob" problem that CF for VMs found itself dealing with.

tcdowney:
When we tried to do this back in the day on CF for K8s we found a lot of variance between registries around how they handled image deletion. I'm having trouble remembering the specifics, but I'm thinking some had non-standard delete endpoints (probably DockerHub?) while others supported the ability to delete by digest and others did not. Anyways, I'm wondering if y'all ran into similar issues here?

gcapizzi (approved):

PR #2265, opened 2023-03-09
Remove the "[Chore]:" prefix of chore titles

georgethebeatle:
The chore label already makes it clear enough that this is a chore. Clicking on the label filters by it. Looks like the name prefix is redundant.


tcdowney (approved):
Makes sense to me. It's was a pretty meaningless prefix for outside contributors as well, in my opinion.

Issue #1566, opened 2022-08-25
[Chore]:  Create a new e2e test suite for kubectl-based workflows


tcdowney:
### Background

The current e2e tests are oriented around a user interacting with the Korifi implementation of the CF API. We have gotten feedback from some users that they prefer to declaratively apply config with `kubectl` for certain workflows (so far declarative org and space management has been a big one). We'll start with testing `CFOrg`/`CFSpace` workflows, but this test suite could be home to others as well in the future.

This chore is to bootstrap a new test suite in `tests` for more intentional testing of `kubectl`-driven interactions.

---

### Action to take

* Create a new test suite to sit alongside `tests/e2e` and `tests/smoke`  where can start including some `kubectl`-driven end-to-end tests.

---

### Impact

Having this suite available will enable us to more easily test expected `kubectl` interactions and fine pain points/regressions.

---

### Dev Notes

* We should probably bootstrap it with some simple test -- like a user can apply a `CFOrg` and see that the `Namespace` is created eventually. A future story will test the full workflow.

clintyoshimura:
Created a new CRDs smoke test suite that runs tests via kubectl since we thought that would be more representative of what users would do. Added it to the CI pipelines as well.

PR #2268, opened 2023-03-10
Add the crds test suite to test CRD-based workflows

matt-royal:
 #1566

## What is this change about?
 This PR adds a new test suite that uses the kubernetes API to drive korifi (via kubectl). For now it only tests org creation.

 No

## Acceptance Steps
 Run the new tests locally against a kind cluster and verify that they pass

 @clintyoshimura @tcdowney

### DIFF
--- a/tests/crds/README.md
+++ b/tests/crds/README.md
@@ -1,0 +1,22 @@
+# Kubernetes CRD tests
+
+These tests run a "smoke test" of the Korifi system, driven exclusively by the kubernetes API (via our Custom Resources).
+
+## Prerequisites
+
+Before running these tests, you must be targeting a kubernetes cluster that has korifi installed. This can be done via
+the default kubeconfig or by setting the `KUBECONFIG` environment variable. Your currently selected context must be for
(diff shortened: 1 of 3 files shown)


clintyoshimura (approved):
LGTM

gcapizzi:
Have we considered using `client-go`? The CLI seems a bit clunky and is just a wrapper on top of `client-go` in the end, so maybe we could go with that and avoid dealing with stdin/stdout?

matt-royal:
@gcapizzi The AMER team has been discussing these tests as a way to document workflows for users, which is part of why we chose to use the kubectl CLI directly. It lets us do things like call `kubectl wait` like a real user would, rather than use an `Eventually` to poll the CLI, for example. @tcdowney, is this what you had in mind?

tcdowney:
> @tcdowney, is this what you had in mind?

Yeah, I agree that it's a bit clunky, but I do believe that using `kubectl` is more representative of what users will actually be doing here (I've seen people like to store this configuration in git and apply it with tools like `kubectl` and `kapp`). My worry with client-go is that our tests may just end up reimplementing Korifi API logic in a way that's not realistic for end users.

PR #2286, opened 2023-03-15
Set ServiceBinding .spec.name to the CFServiceBinding's displayName

tcdowney:
Issue: https://github.com/cloudfoundry/korifi/issues/2285

## What is this change about?
Some service binding libraries (such as the Oauth2 connector in Spring Cloud Bindings) use a Service Binding's name as a configuration value. This does not always place nicely with the Service Instance GUID that Korifi currently sets for the Service Binding's name.

This change allows users to override the `.spec.name` property of a Service Binding during binding creation using the `cf bind-service --binding-name` parameter. If `--binding-name` is omitted then the ServiceBinding continues to use the Service Instance GUID as its name.

No, it should only impact new service bindings.

## Acceptance Steps
1. Push the golang test app (`tests/e2e/assets/golang`): `cf push test-app`
2. Create an UPSI: `cf cups my-upsi -p '{"secret": "credential"}'`
3. Bind the service and specify a name: `cf bind-service test-app my-upsi --binding-name my-custom-binding`
4. curl the `/servicebindings` or `/servicebindingroot` endpoint to see that the binding name is reflected in the app container

@akrishna90 since this builds on those service binding tests you added :)

### DIFF
--- a/controllers/controllers/services/cfservicebinding_controller.go
+++ b/controllers/controllers/services/cfservicebinding_controller.go
@@ -185,8 +185,14 @@ func generateDesiredServiceBinding(actualServiceBinding *servicebindingv1beta1.S
 			UID:        cfServiceBinding.UID,
 		},
 	}
+
+	bindingName := secret.Name
+	if cfServiceBinding.Spec.DisplayName != nil {
+		bindingName = *cfServiceBinding.Spec.DisplayName
+	}
+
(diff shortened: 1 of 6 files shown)


tcdowney:
Here you go @alexbarbato 🙂

akrishna90 (approved):

matt-royal (approved):

PR #2287, opened 2023-03-15
Prevent k8s admin from creating orgs

Birdrock:
No.

## What is this change about?
The Kubernetes admin can create orgs, but not fetch them, no do anything further. The Kubernetes admin should not be able to create orgs via the CF api.

This fixes unintended functionality. No breaking changes are included from intended functionality.

## Acceptance Steps
Create a Kind cluster and log in with the CF CLI as 'kind-kind'. Try to create an org `cf create-org my-org`. Notice that an error `No provided roles allow creation of Orgs` is returned.

@davewalter @tcdowney

### DIFF
--- a/api/authorization/namespace_permissions.go
+++ b/api/authorization/namespace_permissions.go
@@ -119,15 +119,15 @@ func (o *NamespacePermissions) hasRole(ctx context.Context, info Info, namespace
 	return false, nil
 }
 
-func (o *NamespacePermissions) AuthorizedCreateOrg(ctx context.Context, info Info) (bool, error) {
-	hasRole, err := o.hasRole(ctx, info, types.NamespacedName{Namespace: "cf", Name: "default-admin-binding"})
+func (o *NamespacePermissions) AuthorizedCreateOrg(ctx context.Context, info Info, namespace string) (bool, error) {
+	hasRole, err := o.hasRole(ctx, info, types.NamespacedName{Namespace: namespace, Name: "default-admin-binding"})
 	if hasRole {
 		return true, err
 	}
 	if k8serrors.IsNotFound(err) {
 		return false, nil
 	}
-	hasRole, err = o.hasRole(ctx, info, types.NamespacedName{Namespace: "cf", Name: "korifi-controllers-organization-manager"})
+	hasRole, err = o.hasRole(ctx, info, types.NamespacedName{Namespace: namespace, Name: "korifi-controllers-organization-manager"})
 	if hasRole {
 		return true, err
 	}
--- a/api/authorization/namespace_permissions.go
+++ b/api/authorization/namespace_permissions.go
@@ -9,10 +9,13 @@ import (
 
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
+
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-//+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=rolebindings,verbs=list
+//+kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=rolebindings,verbs=get;list
 //+kubebuilder:rbac:groups="",resources=namespaces,verbs=list
 
 //counterfeiter:generate -o fake -fake-name IdentityProvider . IdentityProvider
@@ -33,8 +36,8 @@ func NewNamespacePermissions(privilegedClient client.Client, identityProvider Id
 	}
 }
 
-func (o *NamespacePermissions) GetAuthorizedOrgNamespaces(ctx context.Context, info Info) (map[string]bool, error) {
-	return o.getAuthorizedNamespaces(ctx, info, korifiv1alpha1.OrgNameKey, "Org")
+func (n *NamespacePermissions) GetAuthorizedOrgNamespaces(ctx context.Context, info Info) (map[string]bool, error) {
+	return n.getAuthorizedNamespaces(ctx, info, korifiv1alpha1.OrgNameKey, "Org")
 }
 
 func (o *NamespacePermissions) GetAuthorizedSpaceNamespaces(ctx context.Context, info Info) (map[string]bool, error) {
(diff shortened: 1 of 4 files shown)


api/authorization/namespace_permissions.go
```diff
@@ -94,3 +97,43 @@ func (o *NamespacePermissions) AuthorizedIn(ctx context.Context, identity Identi
...
+
+func (o *NamespacePermissions) AuthorizedCreateOrg(ctx context.Context, info Info) (bool, error) {
+	hasRole, err := o.hasRole(ctx, info, types.NamespacedName{Namespace: "cf", Name: "default-admin-binding"})
+	if hasRole {
+		return true, err
+	}
+	if k8serrors.IsNotFound(err) {
+		return false, nil
+	}
+	hasRole, err = o.hasRole(ctx, info, types.NamespacedName{Namespace: "cf", Name: "korifi-controllers-organization-manager"})
```
tcdowney, line 130: It feels off to have hardcoded role names and namespaces here. I know `RootNamespace` is definitely a config property and I think this `RoleMappings` thing is something that we could use as well. https://github.com/cloudfoundry/korifi/blob/main/api/config/config.go#L42

Birdrock: Forgot to go back to change that :sweat_smile:. I'll take a loot at the RoleMappings.

tcdowney:
So is the main reason the Kubernetes Admin can't list orgs because they don't explicitly have a rolebinding in any of the org namespaces? Like this logic is preventing them?
https://github.com/cloudfoundry/korifi/blob/1e0a1de844b0fe9390bb5f5a37e33011cf0d91a8/api/authorization/namespace_permissions.go#L68

I feel this change leaks some of the knowledge of what a role can/can't do to the Korifi API which isn't ideal since it reduces the ability for operators to define their own roles in the future. Is there another way to validate that they are a user "known" to Korifi without actually caring about whether or not they should be able to create orgs? (the actual call to create will fail if they don't have permission to)

api/authorization/namespace_permissions.go
tcdowney: What if you just called `AuthorizedIn` for the user on the root namespace? Would that be sufficient to prevent the admin user from getting into this situation?

Birdrock:
> So is the main reason the Kubernetes Admin can't list orgs because they don't explicitly have a rolebinding in any of the org namespaces? Like this logic is preventing them?

The Admin doesn't have any rolebindings, so it is not allowed to view orgs. There was nothing preventing them from creating orgs, however. I'm trying to put a check in to make sure that you have permissions to create orgs via the CF API. Kubernetes Admin may still create resources directly. 

> https://github.com/cloudfoundry/korifi/blob/1e0a1de844b0fe9390bb5f5a37e33011cf0d91a8/api/authorization/namespace_permissions.go#L68
> 
> I feel this change leaks some of the knowledge of what a role can/can't do to the Korifi API which isn't ideal since it reduces the ability for operators to define their own roles in the future. Is there another way to validate that they are a user "known" to Korifi without actually caring about whether or not they should be able to create orgs? (the actual call to create will fail if they don't have permission to)

The indicated code is from January 2022, so functionality is not changing here. If we want to improve upon this, we should probably spin out an issue to address it.

tcdowney:
> > So is the main reason the Kubernetes Admin can't list orgs because they don't explicitly have a rolebinding in any of the org namespaces? Like this logic is preventing them?
> 
> The Admin doesn't have any rolebindings, so it is not allowed to view orgs. There was nothing preventing them from creating orgs, however. I'm trying to put a check in to make sure that you have permissions to create orgs via the CF API. Kubernetes Admin may still create resources directly.
> 
> > https://github.com/cloudfoundry/korifi/blob/1e0a1de844b0fe9390bb5f5a37e33011cf0d91a8/api/authorization/namespace_permissions.go#L68
> > 
> > I feel this change leaks some of the knowledge of what a role can/can't do to the Korifi API which isn't ideal since it reduces the ability for operators to define their own roles in the future. Is there another way to validate that they are a user "known" to Korifi without actually caring about whether or not they should be able to create orgs? (the actual call to create will fail if they don't have permission to)
> 
> The indicated code is from January 2022, so functionality is not changing here. If we want to improve upon this, we should probably spin out an issue to address it.

Hmm, I wasn't saying there is a problem with it. I just wanted to confirm my understanding of why the cluster admin could create, but not view orgs through the API. Sounds like that's the case based on what you just said. Anyways, my main point is I think we could just check to ensure that the user has a RoleBinding _at all_ in the root namespace instead of checking for them to have specific roles as the initial gate.

api/authorization/namespace_permissions.go
Birdrock: We can use RoleMappings to keep names more consistent, but unless we check the actual RBAC perms on a role, we wouldn't know anything about the RBAC associated with a custom role. That means listing all RoleBindings in the root namespace, filtering for the User/ServiceAccount, then fetching all Roles/ClusterRoles, then iterating through all rules to find out if they have CFOrg create. This would be more flexible, but much more heavy.

tcdowney: I don't think it matters if they have `CFOrg` create or not, the call on line 94 to `r.createOrgCR(ctx, info, userClient, &korifiv1alpha1.CFOrg{ ...` will fail if they don't have permission to create orgs. We're effectively doing double authorization with this.

What I'm suggesting is just calling `AuthorizedIn` as is. That would list all the RoleBindings and filter for the User/ServiceAccount sure, but that's still lighter-weight than all of our list endpoints.

Birdrock: > What if you just called `AuthorizedIn` for the user on the root namespace? Would that be sufficient to prevent the admin user from getting into this situation?

We could leverage some of the existing functionality for this purpose, but it would solve the issue by side effect rather than intentionally.

Birdrock:
> Isn't the problem here the fact that the Kubernetes admin user isn't a CF user? We should probably just prevent people from using users that are not Korifi users (alas, they don't have a `RoleBinding` in the root namespace).
> 
> We already have a [middleware](https://github.com/cloudfoundry/korifi/blob/main/api/middleware/cf_user.go) that issues warnings about this, we could make it reject requests instead? I don't think this problem is specific to the Kubernetes admin or creating orgs in particular, any user that isn't bound to one of our roles will probably have issues at some point, as we have designed all our permissions around our roles and the way they propagate.
> 
> Rejecting non-Korifi users also means we might need to put in some extra work to support custom roles, but I don't think that's impossible: we could allow people to create roles without limiting the type to a CF one?

I talked with @tcdowney and this seems to be the cleanest short term fix. It isn't bullet-proof - if you really want to misconfigure your RBAC, a user with a CF Role that *does not* allow Org creation will still be able to create Orgs if they have CFOrg create RBAC from somewhere else. With that caveat, I'm closing this and will proceed in the middleware direction with a 403 forbidden if you have no CF Roles

PR #2296, opened 2023-03-17
Restrict all Korifi API endpoints

Birdrock:
No.

## What is this change about?
Restrict Korifi API (CF API) usage to Users or ServiceAccounts with a CF Role

If one has a User/SA configured with RBAC for CFOrg create *and* a CF Role of any type, it will allow CFOrg creation via the CF API. Previously, a CF Role was not required. This is an extreme edge case, but worth noting.

## Acceptance Steps
1. Deploy
2. Create User/SA
3. Use KubeConfig from User/SA
4. Attempt to `cf login`

@gcapizzi 

## Notes
* This implementation also blocks produces a 403 on `/whoami` which will cause `cf target` to fail.
* Need to gain consensus about whether blocking all CF API endpoints is ok, including innocuous ones.

### DIFF
--- a/api/middleware/cf_user.go
+++ b/api/middleware/cf_user.go
@@ -59,6 +59,9 @@ func (m *cfUser) middleware(next http.Handler) http.Handler {
 
 		if !isCFUser {
 			w.Header().Add("X-Cf-Warnings", fmt.Sprintf("Warning: subject '%s/%s' has no CF roles assigned. This is not supported and may cause unexpected behaviour.", identity.Kind, identity.Name))
+			w.WriteHeader(http.StatusForbidden)
+			next.ServeHTTP(w, r)
+			return
 		}
 
 		next.ServeHTTP(w, r)
--- a/api/middleware/cf_user_test.go
+++ b/api/middleware/cf_user_test.go
@@ -126,7 +126,7 @@ var _ = Describe("CfUserMiddleware", func() {
 		})
 
 		It("delegates to the next middleware", func() {
-			Expect(rr).To(HaveHTTPStatus(http.StatusTeapot))
+			Expect(rr).To(HaveHTTPStatus(http.StatusForbidden))
 		})
 
 		It("sets the X-Cf-Warning header", func() {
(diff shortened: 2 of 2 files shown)


api/middleware/cf_user.go
```diff
@@ -59,6 +59,9 @@ func (m *cfUser) middleware(next http.Handler) http.Handler {
 
 		if !isCFUser {
 			w.Header().Add("X-Cf-Warnings", fmt.Sprintf("Warning: subject '%s/%s' has no CF roles assigned. This is not supported and may cause unexpected behaviour.", identity.Kind, identity.Name))
```
gcapizzi, line 61: Do we still want to print this warning, given we're now returning an error? Maybe we should but the same message inside the error body?

api/middleware/cf_user.go
```diff
@@ -59,6 +59,9 @@ func (m *cfUser) middleware(next http.Handler) http.Handler {
...
+			w.WriteHeader(http.StatusForbidden)
+			next.ServeHTTP(w, r)
```
gcapizzi, line 63: If we want to block these requests, we should interrupt the middleware chain and just return, so this should go.

api/middleware/cf_user_test.go
```diff
@@ -126,7 +126,7 @@ var _ = Describe("CfUserMiddleware", func() {
 		})
 
 		It("delegates to the next middleware", func() {
```
gcapizzi, line 128: We should rephrase these test descriptions as the new purpose of the middleware is to block requests instead of just adding a warning and then delegating to the next middleware.

api/middleware/cf_user_test.go
```diff
@@ -126,7 +126,7 @@ var _ = Describe("CfUserMiddleware", func() {
...
-			Expect(rr).To(HaveHTTPStatus(http.StatusTeapot))
+			Expect(rr).To(HaveHTTPStatus(http.StatusForbidden))
 		})
 
 		It("sets the X-Cf-Warning header", func() {
```
gcapizzi, line 132: See above about wether we want to keep the warning.

gcapizzi (changes requested):

georgethebeatle:
@Birdrock any updates on this?

Birdrock:
We should take a fresh look at this if it is an issue after we have done more work on users/SAs in general. Closing as not needed now.

Issue #1569, opened 2022-08-25
[Feature]: Platform Operator can view docs describing how to declaratively manage orgs/spaces/roles


tcdowney:
### Blockers/Dependencies

#1568 & #2267 

### Background

**As a** Platform Operator
**I want** to be able to view some docs that describe how to declaratively manage the orgs, spaces and rolebindings in my installation
**So that** I know how to use the `CFOrg`,`CFSpace` and `RoleBinding` resources

### Acceptance Criteria

**GIVEN** I have installed Korifi
**WHEN I** read the docs
**THEN I** can use `kubectl` (or `kapp`) to setup the orgs/spaces/roles in my environment

---

### Dev Notes

*  Primarily document what the e2e tests added in #1568 are doing with the resources and any gotchas
* Make sure we document the cluster roles we currently support.
* Describe what the `cloudfoundry.org/propagate-cf-role` annotation does and when to use it.
* Be clear on mutability constraints
    -  `spec.displayName` is mutable on cforg & cfspace
* Mention about constraints in order to be compatible with CF CLI
    - RoleBindings must be 1:1 SA/user
    - Any labels/annotations necessary.

PR #2344, opened 2023-03-22
Add docs for declaratively managing orgs/spaces/roles

akrishna90:

#1569 

## What is this change about?

Added docs for declaratively managing orgs/spaces/roles

No

## Acceptance Steps

See issue #1569


akrishna90:
Its easier to read the docs here - https://github.com/cloudfoundry/korifi/blob/issues/1569/INSTALL.md#declaratively-manage-orgs-spaces-and-roles

gcapizzi:
Also: shall we be calling this "declarative", given an org and a space can't be applied together, due to the hard dependency of the space on the org being reconciled? People won't be able to put all those YAML fragments in the same file and apply them, so maybe this is more about using the CRDs API directly vs the REST API?

akrishna90:
Thanks @gcapizzi for feedback. We ended up pulling the docs to a separate document and presented both `kubectl` & `kapp` ways of creating orgs/spaces and rolebindings.

clintyoshimura (approved):
LGTM

PR #2353, opened 2023-03-24
Use unique sources and builds in e2e tests

kieron-dev:
No

## What is this change about?
Google Artifact Registry has problems when pushing and deleting the same blobs concurrently. This causes e2e flakes.

To work around this, we introduce unique.* files in the various source directories and replace the occurences of the string `UNIQUENESS_TAG` with a GUID when zipping the sources.

Now we zip the sources before each push rather than at the start of the suite.

We have reverted the change which puts the current timestamp on the OCI image. It is probably good to have the same SHA when pushing the same sources, and this prevented that.

@gcapizzi


gcapizzi (approved):

tests/e2e/e2e_suite_test.go
```diff
@@ -915,26 +880,6 @@ func expectUnprocessableEntityError(resp *resty.Response, errResp cfErrs, detail
 	))
 }
 
-func getAppBitsFileFromPath(sharedSetup sharedSetupData) string {
-	val, ok := os.LookupEnv("APP_BITS_PATH")
```
kieron-dev, line 919: @davewalter @akrishna90 - we're a bit confused about this. We haven't found any reference to APP_BITS_PATH elsewhere in the project or korifi-ci. We see it was only committed a month ago and has something to do with routes_test.go. Similarly for APP_BITS_OUTPUT env var which is mentioned in that test. We can't see it being set anywhere.

This commit is removing that functionality. Please shout if we need to do something different.

tcdowney: This is needed for CI environments where the set of installed buildpacks is not under our control (e.g. a different team manages the kpack installation, default `ClusterBuilder` configuration, etc.). This isn't the case for Korifi CI, but is important for validating it elsewhere. I think we need to keep it and add a comment or something to that effect.

tcdowney: Having said that, I'm working on reintroducing it and realize now that this code is kind of strange. 🙃

tcdowney: I added it back in this commit https://github.com/cloudfoundry/korifi/pull/2353/commits/184f4072d0644ffdcd93f59b27489c345bc301eb and created a chore with some refactoring ideas https://github.com/cloudfoundry/korifi/issues/2355.

kieron-dev: Thank you! That makes things clearer

PR #2371, opened 2023-03-29
Add GET apps/packages handler

saffronjam:
Fixes https://github.com/cloudfoundry/korifi/issues/2333

## What is this change about?
Add GET apps/packages handler

No

## Acceptance Steps
GIVEN I have pushed an app
WHEN I perform the following request
```
curl "https://api.example.org/v3/apps/[guid]/packages"
```
THEN I get a 200 OK response with a list of packages

### DIFF
--- a/api/handlers/app.go
+++ b/api/handlers/app.go
@@ -33,6 +33,7 @@ const (
 	AppRestartPath                    = "/v3/apps/{guid}/actions/restart"
 	AppEnvVarsPath                    = "/v3/apps/{guid}/environment_variables"
 	AppEnvPath                        = "/v3/apps/{guid}/env"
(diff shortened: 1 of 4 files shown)


gcapizzi:
Hey @saffronjam! Thank you so much for this! Do you feel confident writing some tests for this too? Ideally we'd expect to have a unit test for the handler (see [here](https://github.com/cloudfoundry/korifi/blob/main/api/handlers/app_test.go) for inspiration) and and end-to-end test too (see [here](https://github.com/cloudfoundry/korifi/blob/main/tests/e2e/apps_test.go)).

Happy to help if you're stuck!

saffronjam:
Hi @gcapizzi, I think this should do it!

gcapizzi:
It's in! Thank you so much @saffronjam! 🎉

PR #2383, opened 2023-03-30
Check service account NS when looking at rolebinding subjects

kieron-dev:
No

## What is this change about?
Add missing namespace check when comparing service accounts with rolebinding subjects.

This is used to filter org listing, and for optimization when listing resources generally.

No

## Acceptance Steps
Fewer e2e flakes

@gcapizzi


tcdowney (approved):
LGTM

PR #2445, opened 2023-04-13
Put the selfsigned-issuer in the release NS

davewalter:
No

## What is this change about?
We noticed that the Cert Manager self-signed issuer in the Helm templates was being created in the default namespace. While this worked (perhaps because Cert Manager falls back to the issuer in the default namespace if it cannot find one in the same namespace as the certificate request), it seems better to put the issuer in the same namespace as the rest of Korifi to indicate that Korifi is the one using it. We also want to avoid polluting the default namespace.

No

## Acceptance Steps
Deploy and run tests.

@clintyoshimura


clintyoshimura (approved):
LGTM

PR #2471, opened 2023-04-21
Embed rbac subject in identity

Birdrock:
No.

## What is this change about?
Embed RBAC v1 Subject in Identity. Identical field names are used, so we can more directly use the provided Kubernetes structs. This also allows us to clean up some of the comparison code.

No.

## Acceptance Steps
Run tests as normal.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,13 +46,16 @@ lint: fmt vet
 	golangci-lint run -v
 
 test: lint
-	@for comp in $(COMPONENTS); do make -C $$comp test; done
+	make test-unit
 	make test-tools
 	make test-e2e
 
 
 test-tools:
 	./scripts/run-tests.sh tools
+
+test-unit:
+	for comp in $(COMPONENTS); do make -C $$comp test; done
 
 test-e2e:
 	./scripts/run-tests.sh tests/e2e
--- a/api/authorization/cert_inspector.go
+++ b/api/authorization/cert_inspector.go
@@ -23,7 +23,7 @@ func NewCertInspector(restConfig *rest.Config) *CertInspector {
 	}
 }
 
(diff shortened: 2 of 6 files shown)


matt-royal:
My suggestion is to change Identity to an alias instead of embedding, since the syntax is cleaner. This was easier to show than to explain, so I pushed a commit to the branch to do it. Feel free to roll it back if you disagree.

matt-royal (dismissed):
My suggestion is to change Identity to an alias instead of embedding, since the syntax is cleaner. This was easier to show than to explain, so I pushed a commit to the branch to do it. Feel free to roll it back if you disagree.

matt-royal:
I actually have one more suggestion: `CreateRoleMessage` has the fields `Kind`, `User` and `ServiceAccountNamespace`. These are direct analogues to Identity's fields, so I suggest you just replace them with an `Identity` field to make that obvious and simpler.

Birdrock:
I can see the benefits of making it an alias so that we don't need to drill down into the Subject embedded in. The drawback is that you lose all of the Subject methods from Kubernetes. I don't know if we'll ever need to use any, but it's a trade-off. I agree on updating `CreateRoleMessage` using a fully packaged Identity or Subject.

Birdrock:
> Another suggestion - how about introducing a constructor for `Identity`. I have always felt a bit uneasy when certain fields of a type _might_ not be initialised...

I'll look at this. A constructor is a bit of extra effort but does ensure type safety.

georgethebeatle:
@Birdrock any updates on this?

Birdrock:
Closing for now. Will reference changes as part of forthcoming stories centered around a greater Identity effort (user/SA).

PR #2506, opened 2023-05-05
Reduce the golang version to 1.19 to support older kpack installs

matt-royal:
 No

## What is this change about?
 Reduce the golang version required by the golang test asset to 1.19. This version is still supported and not all installs have a kpack that's new enough to include golang 1.20.

 No

## Acceptance Steps
 PR checks pass

### DIFF
--- a/tests/e2e/assets/golang/go.mod
+++ b/tests/e2e/assets/golang/go.mod
@@ -1,3 +1,3 @@
 module dorifi
 
-go 1.20
+go 1.19 // Support installations with a slightly older version of kpack


tests/e2e/assets/golang/go.mod
```diff
@@ -1,3 +1,3 @@
 module dorifi
 
-go 1.20
+go 1.19
```
tcdowney, line 3: Should we add a comment here? And how do we know when it is safe to bump it?

I am worried Dependabot (or one of us) will bump it back up without realizing.

tcdowney (approved):

PR #2518, opened 2023-05-13
Expand the smoke tests to validate the creation and binding of a service instance, and running a task

davewalter:
No

## What is this change about?
This PR expands the smoke tests to validate the creation and binding of a service instance, and running a task.

No.

## Acceptance Steps
Deploy and run the tests.

### DIFF
--- a/tests/smoke/smoke_test.go
+++ b/tests/smoke/smoke_test.go
@@ -48,9 +48,11 @@ var _ = Describe("Smoke Tests", func() {
 				apiArguments = append(apiArguments, "--skip-ssl-validation")
 			}
 
+			By("targetting the API")
 			cfAPI := cf.Cf(apiArguments...)
 			Eventually(cfAPI).Should(Exit(0))
 
+			By("logging in")
 			loginAs(GetRequiredEnvVar("SMOKE_TEST_USER"))
 
 			appRouteProtocol = GetDefaultedEnvVar("SMOKE_TEST_APP_ROUTE_PROTOCOL", "https")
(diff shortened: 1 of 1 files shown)


tests/smoke/smoke_test.go
```diff
@@ -77,10 +82,25 @@ var _ = Describe("Smoke Tests", func() {
...
 		It("creates a routable app pod in Kubernetes from a source-based app", func() {
 			appName = generator.PrefixedRandomName(NamePrefix, "app")
+			serviceName := generator.PrefixedRandomName(NamePrefix, "svc")
 
-			cfPush := cf.Cf("push", appName, "-p", "assets/test-node-app")
+			By("pushing an unstarted app")
+			cfPush := cf.Cf("push", appName, "-p", "assets/test-node-app", "--no-start")
 			Eventually(cfPush).Should(Exit(0))
 
+			By("creating a user-provided service instance")
```
tcdowney, line 91: I'm curious what additional validation we're hoping to get out of this test? I feel like unless we actually do the same types of assertions that these e2e tests do this is just testing that the Korifi API works for these endpoints? I think you could have a broken Service Bindings Reconciler and these would still pass.

https://github.com/cloudfoundry/korifi/blob/4679c58b428c69c0f8a8c1000f9856fc2c78d052/tests/e2e/apps_test.go#L714

davewalter: I was really just trying to expand the test to exercise the creation of all Korifi custom resources.

tcdowney (approved):

PR #2525, opened 2023-05-17
Support `--strategy=rolling` for `cf push`, `restart` and `restage`

kieron-dev:
#2497

## What is this change about?
Support `--strategy=rolling` for `cf push`, `restart` and `restage`.

The rolling strategy uses the `/v3/deployments` endpoint, so this PR implements that as a transparent layer over CFApps.

Creating a deployment bumps the `app-rev` annotation on the CFApp and ensures the desired state is `STARTED`. Getting a deployment returns the expected JSON with the status dependent on the Ready status condition of the CFApp. If the app is ready, the deployment is complete. If not, it is active.

The rolling restart functionality is implemented in the controllers layer by ensuring stable naming of app workloads and statefulsets, and ensuring the app-rev is propagated to the statefulset pods.

We introduce a new annotation on CFApp: `last-stop-app-rev`. This is set to the `app-dev` value when the app state goes from `STARTED` to `STOPPED`, which is the same time the `app-rev` is automatically bumped. We change the naming calculation of app workloads and statefulsets to use the `last-stop-app-rev` value instead of the `app-rev` value. This means their names are not changed unless the app is stopped an started.

Bumping the `app-rev` on the app either via 'create deployment' or manually on the resource causes a rolling restart of the statefulset pods.

The case where an app is restarted and the CFProcess Controller does not catch the STOPPED state is handled by the appworkload naming. A new appworkload is created and the old one deleted, which means the existing statefulset is deleted and a new one created, ensuring a proper restart of the running app.

Existing apps will continue running after this korifi code becomes live. Their names do not change. However, rolling restarts will fail until a non-rolling `push`, `restart` or `restage` is performed. This is because the version property needs to be removed from the statefulset pod selector, but the field is immutable. We require a new statefulset to be created via one of these methods first.

No, it adds the strategy=rolling support. However, see note above about existing apps needing their statefulsets recreated.

@gcapizzi


api/repositories/deployment_repository_test.go
```diff
@@ -0,0 +1,215 @@
...
+	"code.cloudfoundry.org/korifi/tools/k8s"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("DomainRepository", func() {
```
tcdowney, line 20: Should be `DeploymentRepository`

tcdowney:
Awesome that y'all found a way to support this without requiring forced-downtime for all apps on upgrade. 👏 I think this looks good overall -- especially for the statefulset-runner -- though I think some of us are going to also look at this from the lens of how this might impact some other runner implementations.

During the working group sync today we also discussed the edge case where someone does a `--strategy=rolling` operation on an app that hasn't been restarted to use the new AppWorkload names. I think we said the app will get into a bad state, but can be recovered by doing an explicit `cf restart` and we were discussing ways to make this type of situation more discoverable or return an explicit error. Do you think we should create a followup story to capture some of that context and look into that?

api/repositories/deployment_repository.go
```diff
@@ -0,0 +1,163 @@
...
+	"k8s.io/apimachinery/pkg/api/meta"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const DeploymentResourceType = "Deployment"
+
+type DeploymentRepo struct {
+	userClientFactory  authorization.UserK8sClientFactory
+	namespaceRetriever NamespaceRetriever
+	rootNamespace      string
```
tcdowney, line 23: What is this used for?

api/repositories/deployment_repository_test.go
georgethebeatle: Fixed, thanks

api/repositories/deployment_repository.go
georgethebeatle: We have removed the rootNamespace at it was unused, thanks for spotting

georgethebeatle:
We have filed an [issue](https://github.com/cloudfoundry/korifi/issues/2537) to address concerns expressed on wg sync

georgethebeatle (approved):

Issue #1935, opened 2022-11-23
[Chore]: Pass logger in the context parameter within the controllers code


kieron-dev:
### Background

It is useful to pass a logger through method calls, decorating it with the current method name or area, and setting attributes common to that method.

For example, in the org reconciler, a logger is initialised with the name 'CFOrg Reconciler' and 'name' and 'namespace' attributes set to the name and namespace or the org being reconciled. When invoking `propagateSecret()` from the reconcile method, this logger is passed into that function. It decorates the name with 'propagateSecret', so now all logs inside the function will have the name 'CFOrg Reconciler.propagateSecret', and will already contain the org name and namespace as attributes.

When continuing in this pattern, it becomes easy to trace the flow of execution through the log messages, and the inherited attributes mean useful information is included in the logs without explicitly including those attributes in each log call in the code.

The downside is that `log logr.Logger` needs to be passed as a parameter in every function and method that needs to log, or calls another method or function that needs to log. This has already prompted Code Climate to complain about the number of parameters in some methods.

A solution is to pass the logger as part of the context object that is passed through all methods already. People have reservations about passing objects central to the business logic around in the context. This makes it hard to trace the flow of those objects throughout the code. However, loggers don't demand such rigorous introspection. We should include one in the context at the code entrypoint, and happily use whatever is in the context when we need it.

### Action to take

The `logr` package already has functions for using the context to pass a logger around:

* `func NewContext(ctx context.Context, logger Logger) context.Context`
* `func FromContext(ctx context.Context) (Logger, error)`
* `func FromContextOrDiscard(ctx context.Context) Logger`

So at the controllers entrypoint, which is the patching reconciler, we can set the required attributes in the logger and then invoke `NewContext` to place that logger in a new context derived from the original context. This new context should be passed to the subsequent method calls.

In the other methods, instead of receiving a logger as a parameter, we should invoke one of the `logr.FromContext*` functions to get the logger. We need to decide whether it is worth returning an error if there is no logger, or if we should just continue with a discarding logger (i.e. one which doesn't actually write out any log lines).

If we use the discarding logger, we would need to have some tests that assert that log content is actually written at various places in the code. If we choose to make the reconciler error when a logger is not available, our existing tests _should_ be sufficient to guarantee we haven't accidentally created a new context without a logger and started using that instead somewhere in the code. Therefore, I would recommend that option.

So we need to:
* Ensure a logger is correctly initialized with name and attributes in the patching reconciler
[...]

PR #2540, opened 2023-05-23
All controllers pass logger in the ctx

davewalter:
#1935 

## What is this change about?
This PR updates the org and space controllers to pass the logger to helper functions via the context.

No

## Acceptance Steps
Deploy and run tests.

@kieron-dev

### DIFF
--- a/controllers/controllers/networking/cfdomain_controller_test.go
+++ b/controllers/controllers/networking/cfdomain_controller_test.go
@@ -9,6 +9,7 @@ import (
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
+	"github.com/onsi/gomega/gbytes"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -96,7 +97,7 @@ var _ = Describe("CFDomainReconciler Integration Tests", func() {
 		})
 	})
 
-	Describe("Delete CFDomain", func() {
+	When("a domain is deleted", func() {
 		JustBeforeEach(func() {
 			Expect(k8sClient.Delete(ctx, cfDomain)).To(Succeed())
 		})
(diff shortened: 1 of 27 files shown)


kieron-dev (commented):
My main concern is about the
```
	log, err := logr.FromContext(ctx)
	if err != nil {
		return err
	}
```
we have to do at the top of each function that extracts the log from the context. If getting the log from the context returns an error, there is no point returning that from the reconcile function, as it will only run again, and fail again.

So we could ignore the error / use the discard logger function. But then we might miss when we've broken the context wiring. I suppose introducing assertions in tests that particular messages are logged would solve that, though I'm not sure people are generally keen about testing log messages.

davewalter:
> So we could ignore the error / use the discard logger function. But then we might miss when we've broken the context wiring. I suppose introducing assertions in tests that particular messages are logged would solve that, though I'm not sure people are generally keen about testing log messages.

@kieron-dev I've switched to using the discard function and have added a basic logging test for each controller. Let me know what you think.

controllers/controllers/networking/cfdomain_controller_test.go
```diff
@@ -113,6 +114,10 @@ var _ = Describe("CFDomainReconciler Integration Tests", func() {
 			}).Should(Succeed())
 		})
 	})
+
```
kieron-dev, line 117: This log message is a result of the logger from `log := shared.ObjectLogger(r.log, cfDomain)` in the `ReconcileResource` method. There's no chance of that method not producing a logger, so I don't think we need to test this log.

The log which could potentially be discarded is in the `finalizeCFDomain()` method. So if you move this test up into the delete 'when' block and check for the 'finalizer removed' message, this would be better.

controllers/controllers/networking/cfroute_controller_test.go
```diff
@@ -151,6 +152,10 @@ var _ = Describe("CFRouteReconciler Integration Tests", func() {
 		}).Should(Succeed())
 	})
 
+	It("writes a log message", func() {
```
kieron-dev, line 155: Again, this logger is passed directly from `shared.ObjectLogger()` to `createValidRouteStatus()`, so the logger can never be a discard logger and it's not worth testing.

Rather check the loggers used in the finalize and createOrPatch* methods.

kieron-dev (commented):
The new pattern looks good. I think the log tests should test one or more of the logs that actually take the logger from the context though. The observed generation log always seems to come from a logger which has no chance of being a discard logger.

davewalter:
> The new pattern looks good. I think the log tests should test one or more of the logs that actually take the logger from the context though. The observed generation log always seems to come from a logger which has no chance of being a discard logger.

Good point. I've updated the tests that can assert on log messages from helper functions to use them. Unfortunately, some of the controllers don't log anything in the helper functions. Should we add a debug log line to them so that we can test that everything is wired up properly?

kieron-dev (approved):
I think it's good like this. I wouldn't add logs just to test them.

PR #2560, opened 2023-05-31
Use SetControllerReference where possible

davewalter:
No.

## What is this change about?
This PR updates the controllers to use `SetControllerReference` to set the owner reference for created resources, instead of `SetOwnerReference`. This has the benefit of setting the `BlockOwnerDeletion` flag in the owner reference to true.

No.

## Acceptance Steps
Deploy and run tests.

@clintyoshimura


clintyoshimura (approved):
LGTM

PR #2585, opened 2023-06-08
Set finalizers on Korifi resources in a webhook

danail-branekov:
## Is there a related GitHub Issue?
#2367

## What is this change about?
By setting the finalizer in a webhook we ensure that the finalizer is
installed as part of the object creation thus avoiding potential race
conditions. Previously, finalizers used to be installed in the
resources' controllers that could happen before the resource has been
deleted.

We have introduced 2 webhooks:
* one for the "core" Korifi resources, such as CFOrg, CFSpace, CFApp,
  etc.
* one for the kpack-image-builder related resources - BuildWorkload and
  korifi kpack builds

The reason to have them split in this manner is that the
kpack-image-builder is an optional component, if disabled, we should not
be installing our finalizers as custom BuildWorkload controllers may not
need them.

Now that we are setting finalizers in webhooks, we had to enabled
webhooks for all integration tests that tested finalization. This
results in massive webhook wiring as all the webhooks listed on the
webhooks manifest.yaml has to be wired. Not doing that causes envtest
complaining that no implementation has been registered per webhook
configuration.

Installing webhooks (both validating and mutating) means that we need to
ensure that tests are creating valid objects, which caused another chunk
of changes in tests' setup.

## Does this PR introduce a breaking change?
No

## Acceptance Steps
Not a functional change, everything should work as before

## Tag your pair, your PM, and/or team
@gcapizzi

PR #2598, opened 2023-06-14
Revert movement of BuildWorkload and Build finalizers to webhook

matt-royal:
 No

## What is this change about?
 This partially reverts commit fbbbd46e403dc91df0b9d7e1322df00a3b26e91e.

The reverted changes make an incorrect assumption: that only one builder will be running on a cluster at a time. There are already real world deployments of korifi (i.e. tas-adapter) that have kpack-image-builder running concurrently with other builders. The reverted code only works if kpack-image-builder is guaranteed to be the only builder running on a cluster.

kpack-image-builder may be able to use a webhook to handle finalizers, but only if it can establish ownership of the resources before adding a finalizer. Otherwise, we end up in a situation where other builders have to remove the kpack-image-builder finalizer from BuildWorkloads and Builds that they own

 No. This reverts a breaking change from #2585

## Acceptance Steps
 Unit and e2e tests pass

 @danail-branekov @kieron-dev


matt-royal:
We could fix this issue in tas-adapter by explicitly disabling kpack-image-builder in some situations. However, this would make it more difficult to support dynamically-chosen builders in the future, which we've discussed. When we add the dynamic feature, we'd end up reverting this change or finding another way to avoid kpack-image-builder misappropriating the Builds and BuildWorkloads for other builders, so IMHO it's better to solve that issue now.

kieron-dev:
Thanks for spotting this @matt-royal.

I believe the kpack-image-builder is the only builder that will ever set the `korifi.cloudfoundry.org/build-workload-name` label. It is a constant defined within the kpack-image-builder package, so although the name could be more specific to kpack-image-builder, it should be sufficient to identify the build as having originated from kpack-image builder.

For buildworkloads, it is sufficient to check `Spec.BuilderName == "kpack-image-builder"`.

So regarding the [finalizer webhook](https://github.com/cloudfoundry/korifi/blob/be582b016216d7a4fa7132ea03043830f6fcba79/kpack-image-builder/controllers/webhooks/finalizer/finalizer_webhook.go#L23-L24), I propose leaving the build SetPolicy predicate as-is, and introducing a new predicate for buildworkload that checks the `BuilderName`. I think that's all that's required, but let me know if I've missed something.

matt-royal:
Closing in favor of #2601

PR #2601, opened 2023-06-14
kpack-image-builder must only add finalizers to its BuildWorkloads

kieron-dev:
See #2598

## What is this change about?
Add a new `SetPolicy` for the BuildWorkload finalizer in kpack-image-builder which checks the BuildName on the spec, just like its reconciler does. This way multiple builders can co-exist in korifi without interfering.

This requires the signature of SetPolicy to be changed to give access to the Spec of the object. We now use an unstructured object which can be easily converted to an actual resource type if required.

No

## Acceptance Steps
Use kpack-image-builder in parallel with another builder. Check BuildWorkloads (and kpack.Builds) associated with the other builder do not get kpack-image-builder finalizers added.

@matt-royal


matt-royal:
I just confirmed that this fixes the issue. Thank you, @kieron-dev!

PR #2612, opened 2023-06-14
Avoid issues uninstalling korifi or deleting the root ns

matt-royal:
 No

## What is this change about?
 This PR makes a number of changes to avoid problems uninstalling a korifi deployment:
- Don't propagate helm or kapp labels and annotations to secrets, role bindings or service accounts that are propagated by korifi. This avoids situations where kapp would attempt to tear down the propagated resources directly instead of letting the deletion of the root namespace trigger the process in a predictable order. We haven't seen similar problems for helm, but it seemed appropriate to also remove its labels/annotations, as the resources in question aren't managed by helm either.
- Add the `cloudfoundry.org/propagate-deletion` annotation for service accounts and role bindings. When this is present and set to `"false"`, then deleting the resource in the root namespace or an org namespace won't cause the propagated copies to be deleted from org and space namespaces. This avoids problems where a propagated service account may be required to properly tear down the result of a `BuildWorkload` or `AppWorkload`. Without this change, the root namespace's deletion could trigger the deletion of the service account and its role bindings before they could be used for teardown.
- Remove an earlier check in the CFSpace finalizer logic that attempted to skip propagating deletions when the root namespace was being deleted. It did not work in practice (possibly due to caching or a race condition)
- Along the way, we refactored the CFSpace finalizer logic.

No

## Acceptance Steps

- Deploy korifi
- Create an org and space
- Wait for the kpack-service-account to be propagated to the space's namespace.
- Confirm that neither of those propagated resources has helm labels or annotations on them
- Delete the `kpack-service-account` from the root namespace
- Confirm that the propagated resource in the space's namespace is not deleted

 @davewalter @acosta11 @clintyoshimura

### DIFF
--- a/controllers/api/v1alpha1/shared_types.go
+++ b/controllers/api/v1alpha1/shared_types.go
@@ -23,6 +23,7 @@ const (
 
 	PropagateRoleBindingAnnotation    = "cloudfoundry.org/propagate-cf-role"
 	PropagateServiceAccountAnnotation = "cloudfoundry.org/propagate-service-account"
+	PropagateDeletionAnnotation       = "cloudfoundry.org/propagate-deletion"
 	PropagatedFromLabel               = "cloudfoundry.org/propagated-from"
 )
 
--- a/controllers/config/config.go
+++ b/controllers/config/config.go
@@ -17,19 +17,20 @@ type ControllerConfig struct {
 	IncludeContourRouter     bool `yaml:"includeContourRouter"`
 
 	// core controllers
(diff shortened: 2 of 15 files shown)


matt-royal:
FYI, we chose to make the annotation `cloudfoundry.org/propagate-deletion="false"` so that it is opt-in. It's possible that some operators are using the propagation features for their own service accounts and role bindings and we didn't want to break their use cases by default.

matt-royal:
I realized that there's no reason to keep the admin user's role binding from propagating deletion, so I removed the annotation and force pushed.

acosta11 (approved):

PR #2622, opened 2023-06-16
Update default stack to Ubuntu Jammy Jellyfish

tcdowney:
- Ubuntu Bionic Beaver 18.04 left general support in April 2023 and mainline CF has switched to cflinuxfs4 which is based on Jammy Jellyfish 22.04 LTS
- See https://www.cloudfoundry.org/blog/jammy-jellyfish-cloud-foundry/ for more details on that.

No

## What is this change about?
This change updates our Helm template to install a stack based off of Ubuntu Jammy Jellyfish since Bionic is now at its end-of-life.

Since this is a change to the base image that apps build and run with there may be subtle difference. However, this should be very similar to the differences that apps might encounter when moving from `cflinuxfs3` to `cflinuxfs4` and there are dedicated Slack channels on CF Slack related to this (e.g. [#bosh-jammy](https://app.slack.com/client/T02FL4A1X/C02M2R39Y8Z)).

I don't expect there to be a major impact to Korifi development.

## Acceptance Steps
1. Deploy with these changes
2. Push an app
3. `kubectl exec` into the app container and run `cat /etc/os-release` to confirm the app is running with Jammy

@julian-hj

### DIFF
--- a/helm/korifi/kpack-image-builder/cluster-builder.yaml
+++ b/helm/korifi/kpack-image-builder/cluster-builder.yaml
@@ -17,7 +17,7 @@ kind: ClusterStack
 metadata:
   name: cf-default-stack
 spec:
-  id: "io.buildpacks.stacks.bionic"
+  id: "io.buildpacks.stacks.jammy"
   buildImage:
     image: {{ .Values.kpackImageBuilder.clusterStackBuildImage | quote }}
   runImage:
--- a/helm/korifi/kpack-image-builder/cluster-builder.yaml
+++ b/helm/korifi/kpack-image-builder/cluster-builder.yaml
@@ -17,7 +17,7 @@ kind: ClusterStack
 metadata:
   name: cf-default-stack
 spec:
-  id: "io.buildpacks.stacks.jammy"
+  id: {{ .Values.kpackImageBuilder.clusterStackID | quote }}
   buildImage:
     image: {{ .Values.kpackImageBuilder.clusterStackBuildImage | quote }}
   runImage:
--- a/helm/korifi/values.schema.json
+++ b/helm/korifi/values.schema.json
@@ -377,12 +377,16 @@
           "description": "The time that the kpack Builder will be waited for if not in ready state, berfore the build workload fails. See [`time.ParseDuration`](https://pkg.go.dev/time#ParseDuration) for details on the format, an additional `d` suffix for days is supported.",
           "type": "string"
         },
+        "clusterStackID": {
+          "description": "The ID of the `ClusterStack`. Used when `clusterBuilderName` is blank.",
+          "type": "string"
+        },
         "clusterStackBuildImage": {
-          "description": "The image to use for building defined in the `ClusterStack`. Used when `kpack-image-builder` is blank.",
+          "description": "The image to use for building defined in the `ClusterStack`. Used when `clusterBuilderName` is blank.",
           "type": "string"
         },
         "clusterStackRunImage": {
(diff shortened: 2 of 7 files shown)


kieron-dev:
We should check korifi-ci too when this goes in. For instance, the acceptance cluster needs to use the full stack otherwise the ruby db libs won't compile.

tcdowney:
> We should check korifi-ci too when this goes in. For instance, the acceptance cluster needs to use the full stack otherwise the ruby db libs won't compile.

Thanks, I didn't realize we also configured these there. Though as far as I can tell those are still configured to use the base builder?

https://github.com/cloudfoundry/korifi-ci/pull/10

I think there is a bit of a chicken-and-egg thing here as well, the e2es are failing on this PR with "BuilderNotReady: Check ClusterBuilder" errors.

helm/korifi/values.schema.json
```diff
@@ -365,6 +365,10 @@
           "description": "The time that the kpack Builder will be waited for if not in ready state, berfore the build workload fails. See [`time.ParseDuration`](https://pkg.go.dev/time#ParseDuration) for details on the format, an additional `d` suffix for days is supported.",
           "type": "string"
         },
+        "clusterStackID": {
+          "description": "The ID of the `ClusterStack`. Used when `kpack-image-builder` is blank.",
```
tcdowney, line 369: @kieron-dev I made `clusterStackID` configurable, but was a bit confused by the `"Used when `kpack-image-builder` is blank."` message we have in the description for the other properties. What does that mean?

tcdowney: We discussed this on Slack, this should actually say `clusterBuilderName` instead.

danail-branekov:
@tcdowney this PR is now more than a month old, should we keep it around?

helm/korifi/kpack-image-builder/cluster-builder.yaml
```diff
@@ -17,7 +17,7 @@ kind: ClusterStack
 metadata:
   name: cf-default-stack
 spec:
-  id: "io.buildpacks.stacks.bionic"
+  id: {{ .Values.kpackImageBuilder.clusterStackID | quote }}
```
danail-branekov, line 20: Do users really need to configure this ID? I.e. is that not just a string? If so, we could use something like `cf.default.stack` or something. Reading kpack docs, I am not sure how `ClusterStack`'s `metadata.name` differs from `spec.id` and why do you need it at all

tcdowney:
> @tcdowney this PR is now more than a month old, should we keep it around?

@danail-branekov I will refresh this PR and get some eyes on it again. We had discussed in a WG sync that we'd prefer to get the 0.8 release cut before attempting this since it may temporarily mess up CI (there is a corresponding change in https://github.com/cloudfoundry/korifi-ci/pull/10). That's done now so we can get this going again. :)

helm/korifi/kpack-image-builder/cluster-builder.yaml
tcdowney: The ["Create a Stack" docs](https://buildpacks.io/docs/operator-guide/create-a-stack/) mention it a bit:

> NOTE: The stack identifier implies compatibility with other stacks of that same identifier. For instance, a custom stack may use io.buildpacks.stacks.jammy as its identifier so long as it will work with buildpacks that declare compatibility with the io.buildpacks.stacks.jammy stack.

danail-branekov: Thanks, that makes sense now

julian-hj (approved):

Issue #1929, opened 2022-11-22
[Chore]:  Do not interpolate values into log messages, instead use named values.


kieron-dev:
### Background

We output structured JSON logs. To extract information from logs with a tool, we should be able to search for a particular message and then grab associated fields from the message.

If the log is something like:
```
log.Info(fmt.Sprintf("failed to create CFSpace %s/%s", space.Namespace, space.Name))
```
we need to search for message with prefix 'failed to create CFSpace', and then extract namespace and name from that message string to obtain the details.

If instead we logged like:
```
log.Info("failed to create CFSpace", "namespace", space.Namespace, "name", space.Name)
```
then our log tool can find the messages exactly and extract the detail from log fields.

### Action to take

Modify all instances of log messages constructed with `fmt.Sprintf()` to use machine friendly fields instead.

### Impact

Our logs are easily parseable by external tools

PR #2630, opened 2023-06-21
Eliminate use of Sprintf in log messages

davewalter:
#1929

## What is this change about?
This PR eliminates the use of Sprintf in log messages. It moves injected values to named values instead.

No.

## Acceptance Steps
Deploy and run tests. Check the logs and confirm that the intended data is still being recorded.

### DIFF
--- a/statefulset-runner/api/v1/pod_webhook.go
+++ b/statefulset-runner/api/v1/pod_webhook.go
@@ -72,7 +72,7 @@ func (r *STSPodDefaulter) Default(ctx context.Context, obj runtime.Object) error
 			cfInstanceVar := corev1.EnvVar{Name: controllers.EnvCFInstanceIndex, Value: index}
 			container.Env = append(container.Env, cfInstanceVar)
 
-			podlog.V(1).Info(fmt.Sprintf("patching-instance-index env-var - %s: %s", controllers.EnvCFInstanceIndex, index))
+			podlog.V(1).Info("patching instance index env var", controllers.EnvCFInstanceIndex, index)
 
 			return nil
 		}
--- a/api/handlers/build.go
+++ b/api/handlers/build.go
@@ -3,7 +3,6 @@ package handlers
 import (
 	"context"
 	"errors"
-	"fmt"
 	"net/http"
 	"net/url"
 
@@ -59,7 +58,7 @@ func (h *Build) get(r *http.Request) (*routing.Response, error) {
 
 	build, err := h.buildRepo.GetBuild(r.Context(), authInfo, buildGUID)
 	if err != nil {
-		return nil, apierrors.LogAndReturn(logger, apierrors.ForbiddenAsNotFound(err), fmt.Sprintf("Failed to fetch %s from Kubernetes", repositories.BuildResourceType), "guid", buildGUID)
+		return nil, apierrors.LogAndReturn(logger, apierrors.ForbiddenAsNotFound(err), "failed to fetch "+repositories.BuildResourceType, "guid", buildGUID)
 	}
 
 	return routing.NewResponse(http.StatusOK).WithBody(presenter.ForBuild(build, h.serverURL)), nil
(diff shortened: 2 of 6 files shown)


statefulset-runner/api/v1/pod_webhook.go
```diff
@@ -72,7 +72,7 @@ func (r *STSPodDefaulter) Default(ctx context.Context, obj runtime.Object) error
 			cfInstanceVar := corev1.EnvVar{Name: controllers.EnvCFInstanceIndex, Value: index}
 			container.Env = append(container.Env, cfInstanceVar)
 
-			podlog.V(1).Info(fmt.Sprintf("patching-instance-index env-var - %s: %s", controllers.EnvCFInstanceIndex, index))
+			podlog.V(1).Info("patching instance index env var", controllers.EnvCFInstanceIndex, index)
```
danail-branekov, line 75: `Info` should get key-value pairs as additional arguments, here it takes just two values, maybe adding value keys would look better in the logs?

danail-branekov:
There are a few more occurrences of `Sprintf` when logging:

```
api/handlers/build.go:          return nil, apierrors.LogAndReturn(logger, apierrors.ForbiddenAsNotFound(err), fmt.Sprintf("Failed to fetch %s from Kubernetes", repositories.BuildResourceType), "guid", buildGUID)
api/handlers/deployment.go:             return nil, apierrors.LogAndReturn(logger, apierrors.NewRollingDeployNotSupportedError(err), fmt.Sprintf("Runner '%s' does not support rolling deploys", h.runnerName))
api/handlers/service_binding.go:                return nil, apierrors.LogAndReturn(logger, err, fmt.Sprintf("failed to get %s", repositories.AppResourceType))
api/handlers/service_binding.go:                return nil, apierrors.LogAndReturn(logger, err, fmt.Sprintf("failed to get %s", repositories.ServiceInstanceResourceType))
api/handlers/service_binding.go:                return nil, apierrors.LogAndReturn(logger, err, fmt.Sprintf("failed to list %s", repositories.ServiceBindingResourceType))
api/handlers/service_binding.go:                        return nil, apierrors.LogAndReturn(logger, err, fmt.Sprintf("failed to list %s", repositories.AppResourceType))
controllers/controllers/workloads/cfspace_controller.go:        log.V(1).Info(fmt.Sprintf("finalizing CFSpace for %fs", duration.Seconds()))
tests/e2e/helpers/fail_handler.go:      logHeader := fmt.Sprintf(
tests/e2e/helpers/fail_handler.go:              logHeader = fmt.Sprintf(
tools/k8s/reconcile.go:         log.Info(fmt.Sprintf("unable to fetch %T", obj), "reason", err)

```

statefulset-runner/api/v1/pod_webhook.go
davewalter: My intention was that "CF_INSTANCE_INDEX" would be the key, and its value was the value. This would result in logs such as:

{ "severity": "INFO", ..., "message": "patching instance index env var", "CF_INSTANCE_INDEX": 1 }

instead of:

{ "severity": "INFO", ..., "message": "patching instance index env var", "envVarName": "CF_INSTANCE_INDEX", "envVarValue": 1 }

davewalter:
Thanks for taking a look at this @danail-branekov.

A lot of the occurrences you mentioned are simply using a constant string value in an error message, which I think is just a case of trying to keep the code dry. I have replaced the use of `fmt.Sprintf` with string concatenation to hopefully make this clearer, but I don't think that they should be separate named values as that would make the log message look something like:

{ "severity": "INFO", ..., "message": "failed to fetch object", "objectType": "App" }

instead of just:

{ "severity": "INFO", ..., "message": "failed to fetch App" }

I deliberately ignored the use of Sprintf in the e2e test fail handler, and I'm not sure what to do about the one in the reconcile helper that fetches the type of an object without resorting to the `reflect` package, which seems worse to me.

statefulset-runner/api/v1/pod_webhook.go
danail-branekov: sounds reasonable

danail-branekov (approved):

PR #2635, opened 2023-06-23
Fix manifest validation for `random-route`

tcdowney:
The validator was always raising the following error when random-route was specified:
`"default-route and random-route may not be used together"`

This change makes it so that it only errors when both default-route and random-route are set to true.

Related Slack discussion: https://cloudfoundry.slack.com/archives/C0297673ASK/p1687556306576749

No

## What is this change about?
Fixes a bug where any manifest containing `random-route: true` would cause an error. Minimal example:

```yaml
--- 
applications: 
- name: dora
  random-route: true
```

```
15:57 $ cf push node
Pushing app node to org o / space s as cf-admin...
Applying manifest file /Users/tdowney/go/src/github.com/cloudfoundry/cf-acceptance-tests/assets/node/manifest.yml...

Updating with these attributes...
  applications:
  - name: node
    random-route: true
applications[0].default-route and random-route may not be used together
```

No

## Acceptance Steps
Push an app with manifests containing `random-route`, `default-route`, and mixtures of both. They should work in isolation and fail when both are set to `true`.

@acosta11

### DIFF
--- a/api/payloads/manifest.go
+++ b/api/payloads/manifest.go
@@ -193,7 +193,7 @@ func (m Manifest) Validate() error {
 func (a ManifestApplication) Validate() error {
 	return validation.ValidateStruct(&a,
 		validation.Field(&a.Name, payload_validation.StrictlyRequired),
-		validation.Field(&a.DefaultRoute, validation.When(a.RandomRoute, validation.Nil.Error("and random-route may not be used together"))),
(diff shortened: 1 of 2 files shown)


acosta11 (approved):
Looks good. See the related korifi-dev slack thread for other equivalent options for the validation code. If other folks have a strong preference we can update afterwards.

Issue #2605, opened 2023-06-14
[Feature]: Developer can wait for the deletion of an org to be complete


gcapizzi:
### Background

**As a** developer
**I want** to know when an org has finished deleting
**So that** I can plan other operations accordingly

Dev Notes
* We may want to add timeout logic to check the deletion timestamp on the object set for deletion.
* Note that other jobs default to a 15 minute timeout, but technically the top level delete jobs do not get marked as failed in CF for VMs.
* We're thinking about adding the failed state here along with a small message about checking for the lingering resources like service binding projections
* https://v3-apidocs.cloudfoundry.org/version/3.139.0/index.html#jobs
* Similar to other jobs we suspect this will not require a new crd to track the job, instead encoding the required information in the url 

### Acceptance Criteria

**GIVEN** I have an org
**WHEN I** [delete the org](https://v3-apidocs.cloudfoundry.org/#delete-an-organization)
**AND** I follow the redirect to the deletion job
**THEN I** see that the job looks like this:
```json
{
  "guid": "org.delete~<the-org-guid>",
  "operation": "org.delete",
  "state": "PROCESSING",
  "links": {
    "self": {
      "href": "https://api.example.org/v3/jobs/org.delete~<the-org-guid>"
    }
  },
  "errors": [],
  "warnings": []
}
```

**GIVEN** I have waited long enough for the org deletion to successfully complete
**WHEN I** get the deletion job
**THEN I** see it looks like this:
```json
{
  "guid": "org.delete~<the-org-guid>",
  "operation": "org.delete",
  "state": "COMPLETE",
  "links": {
    "self": {
      "href": "https://api.example.org/v3/jobs/org.delete~<the-org-guid>"
    }
  },
  "errors": [],
  "warnings": []
}
```

**GIVEN** I have waited long enough for the org deletion to time out
**WHEN I** get the deletion job
**THEN I** see it looks roughly like this:
```json
{
  "guid": "org.delete~<the-org-guid>",
  "operation": "org.delete",
  "state": "FAILED",
  "links": {
    "self": {
      "href": "https://api.example.org/v3/jobs/org.delete~<the-org-guid>"
    }
  },
 "errors": [
    {
      "code": 10008,
      "title": "CF-UnprocessableEntity",
      "detail": "something went wrong, check for lingering resources in cforg"
    }
  ],
  "warnings": [
    {
      "detail": "warning! warning!"
    }
]
}
```

**GIVEN** the org doesn't exist
[...]

PR #2638, opened 2023-06-27
Implement org deletion job endpoint

acosta11:
#2605 

## What is this change about?
Return status of org deletion based on underlying CFOrg resource

No

## Acceptance Steps
See issue.

paired w/ @davewalter

### DIFF
--- a/api/handlers/job_test.go
+++ b/api/handlers/job_test.go
@@ -1,11 +1,17 @@
 package handlers_test
 
 import (
+	"fmt"
 	"net/http"
 	"net/http/httptest"
+	"time"
 
+	apierrors "code.cloudfoundry.org/korifi/api/errors"
 	"code.cloudfoundry.org/korifi/api/handlers"
+	"code.cloudfoundry.org/korifi/api/handlers/fake"
+	"code.cloudfoundry.org/korifi/api/repositories"
 	. "code.cloudfoundry.org/korifi/tests/matchers"
+
 	"github.com/google/uuid"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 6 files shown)


acosta11:
Messed up my git author config;  had to reset-author and force push to fix the CLA issue.

akrishna90 (commented):
LGTM. I am skeptical about setting 120s as timeout. Is that enough? Does this need to be configurable? I guess for now it should be fine until we get some feedback or hear otherwise.

api/handlers/job_test.go
```diff
@@ -81,5 +90,96 @@ var _ = Describe("Job", func() {
...
+			When("org delete is in progress", func() {
+				BeforeEach(func() {
+					orgRepo.GetOrgReturns(repositories.OrgRecord{
+						GUID:      "cf-org-guid",
+						DeletedAt: time.Now().Format(time.RFC3339Nano),
+					}, nil)
+				})
+
+				It("returns a processing status", func() {
+					guid := operation + "~" + resourceGUID
```
danail-branekov, line 114: All that setup could be moved into `BeforeEach` - to initialise the request URL - and `JustBeforeEach` - to execute the request. Thus the setup repetition in `It`s would be removed and the test would be aligned with idiomatic Ginkgo style

acosta11:
> LGTM. I am skeptical about setting 120s as timeout. Is that enough? Does this need to be configurable? I guess for now it should be fine until we get some feedback or hear otherwise.

Yeah Dave and I had the same thought about circling back if we get feedback but think it's a reasonable starting point given we already have a 1 minute timeout on app deletion.

akrishna90 (approved):

Issue #2634, opened 2023-06-23
[Feature]: Platform Operator can globally configure the kpack Image cache size


tcdowney:
### Background

Kpack `Image` resources by default create a PVC for a [2GB volume](https://github.com/pivotal/kpack/blob/d0bc8bd6b5cab0d581240e34784f1887ce01d652/pkg/apis/build/v1alpha2/image_validation.go#L33) that is used to cache build outputs for the app. This can be overridden by setting the `spec.cache.volume.size` attribute of a given `Image`. Korifi does not currently expose this config which is impacting apps that require a larger cache.

Eventually we may want to support setting this on a per-App basis, but we could start with a Korifi-wide configuration for some initial control.

**As a** Platform Operator
**I want** to be able to specify the size of the cache that kpack builds have access to
**So that** I can support apps that have larger build cache requirements

### Acceptance Criteria

**GIVEN** I am installing Korifi
**WHEN I** provide Helm values
**THEN I** see there is an option to set the image cache size under the `kpack-image-builder` Helm values
**AND** I am able to push an app that requires a larger build cache

### Dev Notes

* Kpack Image docs: https://github.com/pivotal/kpack/blob/d0bc8bd6b5cab0d581240e34784f1887ce01d652/docs/image.md#image-resources

PR #2639, opened 2023-06-27
Add `buildCacheMB` helm chart value

julian-hj:
 #2634

## What is this change about?

- This value is used to set the size of the persistent disk we allocate for image caching when building images with kpack.
- it defaults to 2GiB which is similar to the 2GB default size that kpack allocates.
- The new setting allows customers building large applications to take advantage of caching.

Note: It's a little odd that the staging parameters are in the `api` section of the configuration, but if we want to eventually plumb these through the CFBuild object so that different builds can have different sized caches, then it *kind of* makes sense. I decided that it would be better to put the caching value next to the other staging resources than it would be to put  some of the configuration in the api section and other things in the `controllers` section. Ultimately, we might want to move more things into the top level settings where they make more sense.

 

No

## Acceptance Steps
 #2634

### DIFF
--- a/README.helm.md
+++ b/README.helm.md
@@ -42,6 +42,7 @@ Here are all the values that can be set for the chart:
   - `lifecycle`: Default lifecycle for apps.
     - `stack` (_String_): Stack.
     - `stagingRequirements`:
+      - `buildCacheMB` (_Integer_): Persistent disk in MB for caching staging artifacts across builds.
       - `diskMB` (_Integer_): Disk in MB for staging.
       - `memoryMB` (_Integer_): Memory in MB for staging.
     - `type` (_String_): Lifecycle type (only `buildpack` accepted currently).
--- a/helm/korifi/values.schema.json
+++ b/helm/korifi/values.schema.json
@@ -182,9 +182,13 @@
                 "diskMB": {
                   "description": "Disk in MB for staging.",
                   "type": "integer"
+                },
+                "buildCacheMB": {
+                  "description": "Persistent disk in MB for caching staging artifacts across builds.",
+                  "type": "integer"
                 }
               },
-              "required": ["memoryMB", "diskMB"]
+              "required": ["memoryMB", "diskMB", "buildCacheMB"]
             }
           },
           "required": ["type", "stack", "stagingRequirements"]
--- a/api/repositories/task_repository_test.go
+++ b/api/repositories/task_repository_test.go
@@ -6,13 +6,13 @@ import (
 	"sync"
 	"time"
 
-	"code.cloudfoundry.org/korifi/tools/k8s"
 	"code.cloudfoundry.org/korifi/api/authorization"
 	apierrors "code.cloudfoundry.org/korifi/api/errors"
 	"code.cloudfoundry.org/korifi/api/repositories"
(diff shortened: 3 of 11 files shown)


README.helm.md
```diff
@@ -43,6 +43,7 @@ Here are all the values that can be set for the chart:
     - `stack` (_String_): Stack.
     - `stagingRequirements`:
       - `diskMB` (_Integer_): Disk in MB for staging.
+      - `imageCacheMB` (_Integer_): Persistent disk in MB for image caching
```
tcdowney, line 46:
```suggestion
      - `imageCacheMB` (_Integer_): Persistent disk in MB for caching staging artifacts across builds.
```

tcdowney: I'm also wondering if we should name this something else like `buildCacheMB` or `stagingCacheMB`. I get why you called it `imageCacheMB`, but I feel like that's only because the configuration is set on the kpack `Image` resource and doesn't necessarily describe what it is actually for.

tcdowney (commented):
Overall LGTM, had a couple thoughts on the configuration name, though.

README.helm.md
julian-hj: `stagingRequirements.stagingCacheMB` feels redundant to me, so I'm inclined toward `buildCacheMB`.

helm/korifi/values.schema.json
```diff
@@ -182,9 +182,13 @@
...
                   "description": "Disk in MB for staging.",
                   "type": "integer"
+                },
+                "imageCacheMB": {
+                  "description": "Persistent disk in MB for image caching",
+                  "type": "integer"
                 }
               },
-              "required": ["memoryMB", "diskMB"]
+              "required": ["memoryMB", "diskMB", "imageCacheMB"]
```
akrishna90, line 191: Is this supposed to be optional? Its defaulted to 2GB if not set by kpack right? https://github.com/pivotal/kpack/blob/d0bc8bd6b5cab0d581240e34784f1887ce01d652/pkg/apis/build/v1alpha2/image_validation.go#L33

akrishna90: I see that we are setting it to a default value of `2048` in `values.yaml` if user doesn't specify one. This kind of makes it optional for the user. But should we be even setting this values if user doesn't specify it?

README.helm.md
julian-hj: Updated... PTanotherL

helm/korifi/values.schema.json
julian-hj: I was wondering that myself, but we don't appear to have any optional values in the helm chart, so I made it required to follow the conventions we seem to have in place.

tcdowney: I think setting our own default here and requiring it makes it more apparent to the operator what the default actually is. You have to have some pretty deep knowledge of kpack to know it will create a 2GB volume by default.

I also think it's worth calling out that we should treat these as general staging parameters for Korifi since technically you could replace the builder implementation with something other than kpack which may have different defaults or not support build artifact caching at all (this may be an argument for making it optional, but I'd say that's something we could do later).

tcdowney (approved):
lgtm

Issue #2542, opened 2023-05-24
[Feature]: Runner authors can indicate that their runner doesn't support rolling deploys


matt-royal:
### Background

**As a** runner author
**I want** to have the option of disabling rolling deploys
**So that** my users don't experience incorrect behavior

### Acceptance Criteria

**GIVEN** I have a deployment of korifi using a non-default runner
**AND** that runner has indicated that it does not support rolling deploys
**WHEN I** attempt to run `cf push`, `cf restart` or `cf restage` with  `--strategy=rolling`
**THEN I** see an error message explaining that rolling deployments aren't supported

### Dev Notes

One option for implementing this: add a `RunnerInfo` CRD that has a field like `status.capabilities.rollingDeploy` on it. This mirrors the existing `BuilderInfo` custom resource, but scopes it to the runner, since this is a runner concern.

tcdowney:
@Birdrock @matt-royal is this complete?

Birdrock:
Made one more change to align error messages, but once that is through, should be good.

PR #2643, opened 2023-06-27
Change Runner info lookup error

Birdrock:
#2542 

## What is this change about?
Standardize the RunnerInfo lookup error return against current patterns.

No.

## Acceptance Steps
Run tests as normal.

@matt-royal

### DIFF
--- a/api/handlers/deployment.go
+++ b/api/handlers/deployment.go
@@ -2,6 +2,7 @@ package handlers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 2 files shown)


tcdowney (dismissed):
LGTM

What error was returned before? A `500`?

api/handlers/deployment.go
```diff
@@ -70,7 +70,7 @@ func (h *Deployment) create(r *http.Request) (*routing.Response, error) {
 
 	runnerInfo, err := h.runnerInfoRepo.GetRunnerInfo(r.Context(), authInfo, h.runnerName)
 	if err != nil {
-		return nil, apierrors.LogAndReturn(logger, err, "Error getting runner info in repository")
+		return nil, apierrors.LogAndReturn(logger, apierrors.ForbiddenAsNotFound(err), "Error getting runner info in repository")
```
matt-royal, line 73: If I understand correctly, the user will receive a 404 back when they try to use rolling deploys and the runner doesn't support them. I'm not sure that is the correct error code, since it implies that the create endpoint doesn't exist. It would be better to return a 400 with a clear message explaining that rolling deploys aren't supported by this configuration. I'm open to other error codes, but I couldn't find a better fit.

Also, there don't appear to be any tests for this logic branch.

matt-royal (changes requested):

api/handlers/deployment.go
tcdowney: I agree with Matt. I think `404` is better than a `500` (which is what this returned originally right?) and aligns with our other "not implemented" endpoints which just `404`, but the more I think about this it is implemented and in a mixed-runner case it could be supported for some apps but not others.

tcdowney: I think we should definitely keep logging that this was because we couldn't fetch the `RunnerInfo` though.

julian-hj: OK, we think we fixed it.

julian-hj (approved):

api/handlers/deployment_test.go
```diff
@@ -92,6 +92,18 @@ var _ = Describe("Deployment", func() {
 			})
 		})
 
+		When("runner is not capable of rolling deploy", func() {
```
matt-royal, line 95: I think you meant to change this description to be different from the one above it.

matt-royal (approved):
Approved, though you have a duplicate/inaccurate "When" that you should probably change

Issue #2595, opened 2023-06-13
pr-e2e pre-pushing to prevent BLOB_UNKNOWN error can fail with BLOB_UNKNOWN error :)


kieron-dev:
### Background

We introduced a push of all test apps at the beginning of the e2es on GKE to prevent the problem with aggressive garbage collection in GAR. However, since the smoke test and CRD tests run concurrently with e2es, they might also create images in GAR and trigger the error.

We've just seen
```
export Saving europe-docker.pkg.dev/cf-on-k8s-wg/pr-e2e-images/3bcc342a-f090-4548-8228-c9cd3a887356-droplets...
export *** Images (sha256:185f63c95da798094171dee55c0aea68fd6dfc2dff03f0f5c0f7482030aa0707):
export       europe-docker.pkg.dev/cf-on-k8s-wg/pr-e2e-images/3bcc342a-f090-4548-8228-c9cd3a887356-droplets - PUT https://europe-docker.pkg.dev/v2/cf-on-k8s-wg/pr-e2e-images/3bcc342a-f090-4548-8228-c9cd3a887356-droplets/manifests/latest: BLOB_UNKNOWN:
 Manifest references unknown blob(s): sha256:2e85951e2bd88fbf87a675f1226b9f94bacd6779d1a8d0e9410f9b03c8c78248, sha256:d54afb0f06f90dc1c104732ec7410a59c7be3969aa7a78f671ca17e3ab1d646c, sha256:e95b7f1860c5459c13d100ab34a6e9439ec8187a082f1748bf22eda1700c
ae90
export       europe-docker.pkg.dev/cf-on-k8s-wg/pr-e2e-images/3bcc342a-f090-4548-8228-c9cd3a887356-droplets:b1.20230613.101626
export ERROR: failed to export: failed to write image to the following tags: [europe-docker.pkg.dev/cf-on-k8s-wg/pr-e2e-images/3bcc342a-f090-4548-8228-c9cd3a887356-droplets: PUT https://europe-docker.pkg.dev/v2/cf-on-k8s-wg/pr-e2e-images/3bcc342a-f090
[...]

PR #2649, opened 2023-06-29
Do not push test apps from e2e suite setup

danail-branekov:
#2595

## What is this change about?
Do not push test apps from e2e suite setup, concourse CI is taking care of it prior running e2es

No

## Acceptance Steps
e2e tests succeed

### DIFF
--- a/tests/e2e/e2e_suite_test.go
+++ b/tests/e2e/e2e_suite_test.go
@@ -287,30 +287,6 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 	bs, err := json.Marshal(sharedData)
 	Expect(err).NotTo(HaveOccurred())
 
-	SetDefaultEventuallyTimeout(240 * time.Second)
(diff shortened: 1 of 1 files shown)


gcapizzi (dismissed):

tests/e2e/e2e_suite_test.go
```diff
@@ -290,27 +290,6 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 	SetDefaultEventuallyTimeout(240 * time.Second)
```
kieron-dev, line 290: I think these eventually defaults need removing, as they were added just for this pre-push step

kieron-dev (changes requested):
Can you remove those two lines?

kieron-dev (approved):

tests/e2e/e2e_suite_test.go
danail-branekov: Thanks, I did not realised that those are no longer needed.

Issue #2645, opened 2023-06-28
[Bug]: `cf push` not triggering a build 


gcapizzi:
While accepting https://github.com/cloudfoundry/korifi/issues/2474, I have noticed that the following sequence of pushes, using a NodeJS app, would result in `cf push` getting stuck:

1. ```yaml
    ---
    version: 1
    applications:
    - name: app1
      buildpacks:
      - paketo-buildpacks/java
      - paketo-buildpacks/nodejs
    ```
    (succeeds)
2. ```yaml
    ---
    version: 1
    applications:
    - name: app1
      buildpacks:
      - paketo-buildpacks/java
    ```
    (fails, _"No buildpack groups passed detection."_)
3. ```yaml
    ---
    version: 1
    applications:
    - name: app1
      buildpacks:
      - paketo-buildpacks/java
      - paketo-buildpacks/nodejs
    ```
    (gets stuck, no `kpack.Build` gets created)

It looks like a problem with the implementation of https://github.com/cloudfoundry/korifi/issues/2440.

gcapizzi:
@tcdowney I haven't tried anything else, but maybe pushing different sources? That would definitely convince Kpack to generate a new build!

danail-branekov:
I believe the [this commit](https://github.com/cloudfoundry/korifi/commit/961f366d708e144b6b6063377477282474058899) changes the way how ownership to the builder is set. It used to be a simple owner reference, now it is controller reference.

PR #2652, opened 2023-06-29
Build workloads do not own custom kpack builders via ControllerReference

danail-branekov:
#2645

## What is this change about?
Custom kpack builders (the ones that are created on the fly for build
workloads with buildpacks specified) are shared by build workloads
that have the same set of buildpacks. Therefore, setting
ControllerReference from the build workload to the builder is incorrect
as the ControllerReference can only be set once.

This PR sets simple owner references to custom kpack builds instead.

No

## Acceptance Steps
See bug description


davewalter (approved):
Looks good. Thanks for catching/fixing this.

PR #2655, opened 2023-07-03
Support for updating app name, buildpacks and stack

danail-branekov:
https://github.com/cloudfoundry/korifi/issues/2327

## What is this change about?
Support for updating app name, buildpacks and stack

No

## Acceptance Steps
See story

@kieron-dev

### DIFF
--- a/controllers/coordination/integration/integration_suite_test.go
+++ b/controllers/coordination/integration/integration_suite_test.go
@@ -1,12 +1,15 @@
 package integration_test
 
 import (
+	"context"
 	"testing"
 	"time"
 
+	"code.cloudfoundry.org/korifi/tests/helpers"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/envtest"
 	logf "sigs.k8s.io/controller-runtime/pkg/log"
@@ -22,8 +25,10 @@ func TestIntegration(t *testing.T) {
 }
 
 var (
-	testEnv   *envtest.Environment
-	k8sClient client.Client
+	testEnv           *envtest.Environment
+	k8sClient         client.Client
+	controllersClient client.Client
+	cacheStop         context.CancelFunc
 )
 
 var _ = BeforeSuite(func() {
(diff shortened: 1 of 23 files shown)


kieron-dev (dismissed):

tests/helpers/cache_syncing_client.go
```diff
@@ -19,6 +20,13 @@ func NewCacheSyncingClient(client client.Client) *CacheSyncingClient {
 	}
 }
 
+func NewCacheSyncingClientFromConfig(config *rest.Config, options client.Options) *CacheSyncingClient {
+	k8sClient, err := client.New(config, options)
```
kieron-dev, line 24: Are you sure we are going to get a client with cache configured here? We only pass in scheme in the options. I _think_ we need some cache configuration too.

controllers/coordination/integration/integration_suite_test.go
```diff
@@ -22,22 +23,31 @@ func TestIntegration(t *testing.T) {
...
 
-	k8sClient, err = client.New(cfg, client.Options{Scheme: scheme.Scheme})
+	k8sClient, err = client.New(adminConfig, client.Options{Scheme: scheme.Scheme})
+	Expect(err).NotTo(HaveOccurred())
+
+	controllersUser, err := testEnv.ControlPlane.AddUser(envtest.User{Name: "envtest-controller"}, adminConfig)
+	Expect(err).NotTo(HaveOccurred())
+	helpers.BindUserToControllersRole(k8sClient, "envtest-controller")
+
+	controllersClient, err = client.New(
```
kieron-dev, line 46: I think this client won't use a cache.

tests/helpers/cache_syncing_client.go
danail-branekov: Well, as this is a test utility, I don't think we should be really concerned about it. The cache syncing client does not deal with a cache directly, instead it only awaits that the result of an operation is visible by the client (maybe naming is probably not great, suggestion on naming are welcome). It does not really matter whether a cache is actually used or not.

controllers/coordination/integration/integration_suite_test.go
danail-branekov: I do not think cache is required for the name registry integration test. Or am I wrong?

danail-branekov: yeah, that is why we allowed get on leases as well: https://github.com/cloudfoundry/korifi/pull/2655/files#diff-90cb3e4b36d2d633be58cb91c3d2858b713bf7cb42e1f842b791936c4ce43ff4R30

tests/helpers/cache_syncing_client.go
danail-branekov: Oh, I think I see what you mean - this admin client that I mentioned above would need to share controllers cache, so that stuff that has been done with it is immediately visible by the controllers. This would not be the case here as this client shares no cache. Did I get your point?

danail-branekov: I have addressed the above by changing the method to receive the manager cache so that it can be shared.

controllers/coordination/integration/integration_suite_test.go
kieron-dev: I've run this suite without the list and watch permissions on lease - it still passes. So I really think we need to activate the cache on the client here.

kieron-dev (dismissed):
Looks good to me now

kieron-dev:
> Looks good to me now

Though I think the linter wants you to check the return value from starting the cache.

georgethebeatle (approved):

PR #2691, opened 2023-07-14
Support listing service instances by guid

danail-branekov:
#2692

## What is this change about?
This PR makes it possible to filter a list of service instances by their guids

No

## Acceptance Steps

- Create a couple of services
- `cf curl /v3/service_instances?guids=<service-instance-guid1>,<service-instance-guid2>`
- see the result only contains the two service instances with the guids above

@kieron-dev

### DIFF
--- a/tests/e2e/service_instances_test.go
+++ b/tests/e2e/service_instances_test.go
@@ -154,12 +154,9 @@ var _ = Describe("Service Instances", func() {
 				createSpaceRole("space_developer", certUserName, spaceGUID)
 			})
 
-			It("succeeds", func() {
+			It("deletes the service instance", func() {
 				Expect(httpError).NotTo(HaveOccurred())
 				Expect(httpResp).To(HaveRestyStatusCode(http.StatusNoContent))
-			})
-
-			It("deletes the service instance", func() {
 				Expect(listServiceInstances().Resources).NotTo(ContainElement(
 					MatchFields(IgnoreExtras, Fields{
 						"Name": Equal(existingInstanceName),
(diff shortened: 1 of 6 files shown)


kieron-dev (dismissed):

tests/e2e/service_instances_test.go
```diff
@@ -180,4 +177,48 @@ var _ = Describe("Service Instances", func() {
...
+			anotherSpaceGUID = createSpace(generateGUID("space1"), commonTestOrgGUID)
+			anotherInstanceGUID = createServiceInstance(anotherSpaceGUID, generateGUID("service-instance"), nil)
+		})
+
+		JustBeforeEach(func() {
+			serviceInstancesList = resourceList[resource]{}
+			httpResp, httpError = certClient.R().SetResult(&serviceInstancesList).Get("/v3/service_instances")
+		})
+
+		It("returns an empty list", func() {
```
tcdowney, line 198: This test is failing on CI.

tcdowney: I've been playing around with it and I think this test is very sensitive to test pollution from others that are running concurrently (or fail to clean up properly). I wonder if we should make this assertion more like this one that tests app listing:

https://github.com/cloudfoundry/korifi/blob/main/tests/e2e/apps_test.go#L69-L78

Instead of asserting on the absolute contents we could just check that the list contains a few service instances that we know should be in there. And we make a new space that we explicitly don't have access to and verify that the instances in that space aren't present.

danail-branekov: Thanks for spotting this @tcdowney 

I have applied your proposal so the test should be fine now.

gcapizzi (approved):

PR #2632, opened 2023-06-23
Use GinkgoHelper in tests

davewalter:
## Is there a related GitHub Issue?
No.

## What is this change about?
This PR updates test helpers to call GinkgoHelper() as a better way to help Ginkgo compute failure locations. See [the docs](https://onsi.github.io/ginkgo/#mental-model-how-ginkgo-handles-failure) for details.

## Does this PR introduce a breaking change?
No, just test changes.

## Acceptance Steps
Deploy and run the tests.

PR #2698, opened 2023-07-17
Fix line number reporting on an e2e test failure

danail-branekov:
No

## What is this change about?
Fix line number reporting on an e2e test failure

Our fail handler wrapper was not skipping enough stack frames

No

## Acceptance Steps
N/A


davewalter:
It might be easier to use `GinkgoHelper` here. I updated most of the test helpers to use this in #2632, but I must have missed this one.

gcapizzi (approved):

PR #2701, opened 2023-07-17
Revert "Fix GetDeletedAt for orgs"

matt-royal:
Reverts cloudfoundry/korifi#2700 for #2605 

While this PR's change fixes the org deletion job response for unauthorized users, it also prevents the job endpoint from waiting for the org to actually be deleted. Instead, it returns as soon as the role binding in the org's namespace has been deleted.

The acceptance steps I follow to confirm this:

- Create an org `o` as the admin user
- Use kubectl or k9s to add a finalizer called "nosuchfinalizer" to the `CFOrg` resource created above
- Run `cf delete-org -f o`. The expected response was for it to hang for about a minute and then give an error. Instead, it quickly said the deletion was complete. However, the `CFOrg` still existed on the cluster and was in an indefinite terminating state.


matt-royal:
Closing in favor of #2702

PR #2704, opened 2023-07-18
make the image for the jobs configurable

wanddynosios:
## What is this change about?
Currently, Korifi cannot be installed via helm chart in air-gapped environments, as a hard-coded dockerhub  image is used for the helm jobs


georgethebeatle:
Hey @wanddynosios, thanks for the contribution. Just one small thing - can you move the new helm value under a `helm` key rather than in `global` as we are planning on removing all global values soon? We think `.Values.helm.hookImage` is a good name for it as it is the image that powers the helm hook, while it running as a Job is an implementation detail.

wanddynosios:
@georgethebeatle Sure thing. Adapted as key as requested

danail-branekov (approved):

Issue #2705, opened 2023-07-18
[Bug] App deletion is blocked by non-existent domain resources


davewalter:
### What happened?

When an app is deleted, it tries to update its routes to remove itself from the list of destinations. If the domain no longer exists, the route validating webhook rejects the request, which blocks the app from being deleted.

### What you expected to happen

The app should be able to remove itself from all affected route's destinations and finalize its deletion.

### Acceptance Criteria

**Given** that I have pushed an app with a route to an existing domain
**When** I delete the domain
**And** I then attempt to delete the app
**Then** I see that the CFApp resource successfully deletes

### How to reproduce it (as minimally and precisely as possible)

1. Create a domain
2. Push an app with a default route
3. Delete the domain
4. Attempt to delete the app

### Environment

Revision of codebase: HEAD of main
Kubernetes version (use `kubectl version`): N/A
Cloud provider or hardware configuration: N/A
Others: N/A

tcdowney:
## Acceptance

I accepted this story by doing the following:

1. Pushed an app
2. Manually removed the finalizer from the `defauly-domain` `CFDomain` to force it into this state (an existing `CFRoute` that no longer as a domain associated with it)
3. Deleted the app with `cf delete <app-name>`
4. Validated that the `CFApp` was deleted
  ```
  $ k -n cf-space-6907e1ba-4565-4952-b42d-26839fe91f6d get cfapp
  No resources found in cf-space-6907e1ba-4565-4952-b42d-26839fe91f6d namespace.
  ```
5. Validated that the `CFRoute` was updated (app removed from destinations, status now invalid)
  ```yaml
  apiVersion: korifi.cloudfoundry.org/v1alpha1
kind: CFRoute
metadata:
  annotations:
    korifi.cloudfoundry.org/creation-version: v0.8.2-24-g1acec999
  creationTimestamp: "2023-07-31T22:36:44Z"
  finalizers:
  - cfRoute.korifi.cloudfoundry.org
  generation: 3
  labels:
    korifi.cloudfoundry.org/domain-guid: default-domain
    korifi.cloudfoundry.org/route-guid: cf-route-adbfd732-a2d3-47be-a205-4a17f1b514ad
  name: cf-route-adbfd732-a2d3-47be-a205-4a17f1b514ad
  namespace: cf-space-6907e1ba-4565-4952-b42d-26839fe91f6d
  resourceVersion: "3068"
[...]

PR #2707, opened 2023-07-18
Fix app deletion when updating route destinations

davewalter:
#2705

## What is this change about?
This PR updates the route validator to only validate that a domain exists on creation. We only need to validate the domain on route creation as we do not allow users to change the domain reference on an existing route. Validating the existence of the domain on update can block app deletion if the domain has already been deleted.

No.

## Acceptance Steps
See the A/C in the issue.

@matt-royal

### DIFF
--- a/controllers/webhooks/networking/cfroute_validator.go
+++ b/controllers/webhooks/networking/cfroute_validator.go
@@ -128,7 +128,7 @@ func (v *CFRouteValidator) ValidateUpdate(ctx context.Context, oldObj, obj runti
 		return nil, immutableError.ExportJSONError()
 	}
 
-	_, err := v.validateDestinations(ctx, route)
+	err := v.validateDestinations(ctx, route)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


controllers/webhooks/networking/cfroute_validator.go
```diff
@@ -176,12 +181,9 @@ func (v *CFRouteValidator) fetchDomain(ctx context.Context, route *korifiv1alpha
 	return domain, err
 }
 
-func (v *CFRouteValidator) validateDestinations(ctx context.Context, route *korifiv1alpha1.CFRoute) (*korifiv1alpha1.CFDomain, error) {
-	domain, err := v.fetchDomain(ctx, route)
+func (v *CFRouteValidator) validateDestinations(ctx context.Context, route *korifiv1alpha1.CFRoute) error {
+	err := v.checkDestinationsExistInNamespace(ctx, *route)
```
tcdowney, line 185: Can this cause problems? Maybe not since apps remove themselves from the destinations when they're deleted?

tcdowney (approved):

matt-royal (approved):

controllers/webhooks/networking/cfroute_validator.go
davewalter: I think this is fine for the reason you mentioned. At the time of a route being updated, all destinations should be valid as the destination apps will still exist.

PR #2766, opened 2023-08-10
Downgrade controller-runtime to 1.15.0

danail-branekov:
No

## What is this change about?
We suspect that 1.15.1 might cause env tests to get stuck. We have seen
lots of occurrences on CI, for example

https://ci.korifi.cf-app.com/teams/main/pipelines/main/jobs/run-tests-main/builds/966

No

## Acceptance Steps
Tests should not get stuck on CI

@georgethebeatle

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,6 +2,8 @@ module code.cloudfoundry.org/korifi
 
 go 1.19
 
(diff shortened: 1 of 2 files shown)


danail-branekov:
change implemented in https://github.com/cloudfoundry/korifi/pull/2764

Issue #2650, opened 2023-06-29
[Feature]: Operator can understand the structure of Korifi helm install values


julian-hj:
### Blockers/Dependencies

None

### Background

**As a** platform operator
**I want** intuitive install values inputs
**So that** I can get it right the first time

As part of [another story](https://github.com/cloudfoundry/korifi/issues/2634), we noticed that `stagingRequirements` fall under `api` in the helm [chart values schema](https://github.com/cloudfoundry/korifi/blob/main/helm/korifi/values.schema.json). We believe that it is structured this way because the api and controllers components used to be installed separately, and these were values that flowed through the API shim, but this implementation detail is not useful or intuitive to the end user.

We discussed it in the [working group meeting](https://docs.google.com/document/d/1ULNBEjlrNAgn3ko9y8ZJfwI7mw5-oofYdjl-dhkEoDA/edit#heading=h.u6vai5bi4ba3) and agreed that we should probably review all of the install values to make sure that values under `controllers` or `api` are there for reasons that would make sense to a customer. The rest of the values file is also fair game. If we have any deprecated install values, this would be a good opportunity to remove them.

### Acceptance Criteria

We don't have clear a/c for this story. We want to take a pass through the API values to make them more intuitive. This process should generate a PR to Korifi, and we should **leave that PR open for review from the team** for a couple of days before merging it.

### Dev Notes

Please make sure that the references to install values are also updated in:
- ci scripts
- developer scripts
- README files/installation instructions.

We don't have full CI coverage for all the places where install values could be referenced, so it will be important to grep for them to make sure we didn't miss any references.

PR #2767, opened 2023-08-11
Cleanup Helm values

clintyoshimura:
#2650

## What is this change about?
Cleanup the Helm values to make them more understandable for operators. The api.builderName parameter does not seem to be used and the yaml that Helm generated was the same after it was removed. The api.lifecycle.stagingRequirements parameter was moved to global.stagingRequirements because both api and controllers use those parameters.

no

## Acceptance Steps
All tests pass

@acosta11 

## Things to remember
- Include any links to related PRs, issues, stories, slack discussions, etc... that will help establish context.
- Is there anything else of note that the reviewers should know about this change?
- This project follows the Cloud Foundry [Code of Conduct](https://www.cloudfoundry.org/code-of-conduct/)
-->

### DIFF
--- a/INSTALL.EKS.md
+++ b/INSTALL.EKS.md
@@ -256,14 +256,14 @@ Use the following Helm command to install Korifi:
 ```sh
 helm install korifi https://github.com/cloudfoundry/korifi/releases/download/v<VERSION>/korifi-<VERSION>.tgz \
   --namespace="$KORIFI_NAMESPACE" \
-  --set=global.generateIngressCertificates=true \
-  --set=global.rootNamespace="${ROOT_NAMESPACE}" \
+  --set=generateIngressCertificates=true \
+  --set=rootNamespace="${ROOT_NAMESPACE}" \
   --set=adminUserName="${ADMIN_USERNAME}" \
   --set=api.apiServer.url="api.${BASE_DOMAIN}" \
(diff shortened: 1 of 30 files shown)


clintyoshimura:
In addition to the moved parameters, there were a few other issues that were not as significant that could use discussion.
adminUserName - This is a top level parameter. Should it be under global? All the other top level parameters are groups.
api.expose - Expose the API component via Contour. Should Contour be part of the parameter name?

georgethebeatle:
`global` is a special top level group that is visible to all subcharts in case your helm chart has subcharts. Here are the relevant [docs](https://helm.sh/docs/chart_template_guide/subcharts_and_globals/#global-chart-values). We used to have subcharts in the past and global is a historical leftover. It would be better if we put all common values such as `stagingRequirements` to the top level. We should also move the values under `global` a level up so that they are top level values too.

georgethebeatle (changes requested):
See above comment

clintyoshimura:
@georgethebeatle We moved all the global helm values to the top-level and removed the global group. There will be korifi-ci changes required so when the PR is approved, we can push those changes to get the PR checks to pass and merge it that way. Everyone will have to rebase after that.

georgethebeatle (approved):

PR #2812, opened 2023-08-21
Check the value of CSR_SIGNING_DISALLOWED

davewalter:
No, but it is related to #2784.

## What is this change about?
This PR updates the E2E tests to check the value of the `CSR_SIGNING_DISALLOWED` env var. Concourse always creates the env var in the running container, so it will always be found, which will always cause the tests to be skipped.

No.

## Acceptance Steps
Confirm that running the E2E tests with the `CSR_SIGNING_DISALLOWED` env var set to anything except "true" does not skip the certificate authentication tests.

@clintyoshimura

### DIFF
--- a/tests/e2e/e2e_suite_test.go
+++ b/tests/e2e/e2e_suite_test.go
@@ -17,10 +17,10 @@ import (
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


clintyoshimura (approved):
LGTM

Issue #2823, opened 2023-08-24
As a Korifi API user I want to be able to create builds with docker lifecycle so that I am enabled to run workloads backed up by docker images


danail-branekov:
**GIVEN** Korifi is deployed on a k8s cluster

**WHEN** I create a build with docker lifecycle:

```
curl "https://api.example.org/v3/builds" \
  -X POST \
  -H "Authorization: bearer [token]" \
  -H "Content-type: application/json" \
  -d '{
    "package": {
       "guid": "[package-guid]"
    },
    "lifecycle": {
      "type": "docker"
    }
  }'

```

where `package-guid` is the guid of a package for a public docker image

**THEN I** see that the request above retruns a response similar to

```

HTTP/1.1 201 Created
Content-Type: application/json

{
  "guid": "585bc3c1-3743-497d-88b0-403ad6b56d16",
  "state": "STAGED",
  "lifecycle": {
    "type": "docker",
    "data": {}
  },
  "package": {
    "guid": "8e4da443-f255-499c-8b47-b3729b5b7432"
  },
  "droplet": {
    "guid": "the-droplet-guid"
  },
  "relationships": {
    "app": {
      "data": {
        "guid": "7b34f1cf-7e73-428a-bb5a-8a17a8058396"
      }
    }
  },
  "links": {
    "self": {
      "href": "https://api.example.org/v3/builds/585bc3c1-3743-497d-88b0-403ad6b56d16"
    },
    "app": {
      "href": "https://api.example.org/v3/apps/7b34f1cf-7e73-428a-bb5a-8a17a8058396"
    }
  }
}

```

**THEN I* see that the related droplet references the public docker image:

```
curl "https://api.example.org/v3/droplets/[guid]" \
  -X GET \
  -H "Authorization: bearer [token]"

HTTP/1.1 200 OK
Content-Type: application/json

{
  "guid": "585bc3c1-3743-497d-88b0-403ad6b56d16",
  "state": "STAGED",
  "error": null,
  "lifecycle": {
    "type": "docker",
    "data": {
      "image": "the-image-ref"
    }
  },
  "relationships": {
    "app": {
      "data": {
        "guid": "7b34f1cf-7e73-428a-bb5a-8a17a8058396"
      }
    }
  },
  "links": {
    "self": {
      "href": "https://api.example.org/v3/droplets/585bc3c1-3743-497d-88b0-403ad6b56d16"
    },
    "package": {
[...]

PR #2829, opened 2023-08-25
Support cf docker builds

danail-branekov:
#2823

## What is this change about?
- Support CFBuilds with docker lifecycle
- Remove logging tests

No

## Acceptance Steps
See #2823

@danail-branekov


davewalter:
@danail-branekov @georgethebeatle The logging tests were introduced as part of #2540. The helper functions in the controllers use `logr.FromContextOrDiscard()` to extract the logger from the context, and @kieron-dev and I felt that it would be too easy to break logging by not attaching a logger to the context in the controllers' `ReconcileResource()` functions so we decided to add these tests to confirm that everything is wired together correctly.

PR #2830, opened 2023-08-28
Setup logger in context in the patching reconciler

danail-branekov:
No

## What is this change about?
Setup a logger in the patching reconciler and inject it into the context that is passed to concrete object reconcilers. This way object reconcilers can just use that logger and not bother setting it up themselves.

This PR is a response to @davewalter's [comment](https://github.com/cloudfoundry/korifi/pull/2829#issuecomment-1693866828). Dave, could you please review this one?

No

## Acceptance Steps
Controllers logs appear

@davewalter


davewalter (approved):
Looks good.

Issue #2797, opened 2023-08-17
As a Korifi user I want to see an error when I try to create a package with lifecycle type mismatching the one in the app


danail-branekov:
As a Korifi user I want to see an error when I try to create a package with lifecycle type mismatching the one in the app

**GIVEN** I have pushed a docker app to korifi
**WHEN I** create a package for it with lifecycle type buildpack
**THEN I** see an error

PR #2857, opened 2023-09-11
Package type validation

danail-branekov:
#2797

## What is this change about?
Validate CFPackage type to match the CFApp lifecycle type by introducing a validating webhook that:
* Validates that the package type matches the referenced app lifecycle when the CFPackage is created
* Denies CFPackage type modification

Not breaking, but with this PR in order to create a CFPackage, the referenced CFApp must exists. While this is not a problem for API users, it _might_ be a bit awkward for kubectl users who would like to apply an yaml file that describes a CFApp and a CFPAckage. In order to break this dependency we should probably reconsider all Korifi validation by e.g. introducing a `Valid` status condition for all of them and assert on that condition being set to `true`. Sounds related to https://github.com/cloudfoundry/korifi/issues/2098

## Acceptance Steps
See story

### DIFF
--- a/controllers/api/v1alpha1/cfpackage_webhook_test.go
+++ b/controllers/api/v1alpha1/cfpackage_webhook_test.go
@@ -1,9 +1,13 @@
 package v1alpha1_test
 
 import (
+	"context"
+
 	korifiv1alpha1 "code.cloudfoundry.org/korifi/controllers/api/v1alpha1"
 	. "code.cloudfoundry.org/korifi/controllers/controllers/workloads/testutils"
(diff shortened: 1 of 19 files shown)


georgethebeatle (approved):

danail-branekov:
Converting this PR to a draft one until we sort out the concerns raised in [this slack thread](https://cloudfoundry.slack.com/archives/C0297673ASK/p1694447409279149)

danail-branekov:
Discarding in favour of https://github.com/cloudfoundry/korifi/pull/2875

PR #2863, opened 2023-09-13
Error when uploading package bits to non-bits packages

danail-branekov:
https://github.com/cloudfoundry/korifi/issues/2821

## What is this change about?
Error when uploading package bits to non-bits packages

No

## Acceptance Steps
See story


julian-hj (approved):
LGTM

PR #2876, opened 2023-09-18
replace dummy with example

julian-hj:
to avoid offending anyone.

No

## What is this change about?
This change replaces the term "dummy" with "example" to avoid using terms people could find offensive or triggering.

No

## Acceptance Steps

Normal tests should cover it. The change only affects a single test file.


davewalter (approved):

danail-branekov:
:sigh: 
the list of offensive works keeps on growing...

PR #2910, opened 2023-09-29
Korifi installer for kind clusters

danail-branekov:
https://github.com/cloudfoundry/korifi/issues/2895

## What is this change about?
Introduce a k8s job definition that when run on a kind cluster would install Korifi dependencies and latest Korifi release. This also involves a Dockerfile for the job image.

No

## Acceptance Steps
Install Korifi on kind following INSTALL.kind.md

@cloudfoundry/wg-cf-on-k8s

### DIFF
--- a/scripts/installer/install-korifi-kind.yaml
+++ b/scripts/installer/install-korifi-kind.yaml
@@ -1,0 +1,113 @@
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: korifi-installer
+
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: cf
+  labels:
+    pod-security.kubernetes.io/audit: restricted
+    pod-security.kubernetes.io/enforce: restricted
+
+---
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: korifi
+  labels:
+    pod-security.kubernetes.io/audit: restricted
+    pod-security.kubernetes.io/enforce: restricted
+
+---
+apiVersion: v1
+data:
+  .dockerconfigjson: eyJhdXRocyI6eyJsb2NhbHJlZ2lzdHJ5LWRvY2tlci1yZWdpc3RyeS5kZWZhdWx0LnN2Yy5jbHVzdGVyLmxvY2FsOjMwMDUwIjp7InVzZXJuYW1lIjoidXNlciIsInBhc3N3b3JkIjoicGFzc3dvcmQiLCJhdXRoIjoiZFhObGNqcHdZWE56ZDI5eVpBPT0ifX19
+kind: Secret
+metadata:
+  name: image-registry-credentials
+  namespace: cf
+type: kubernetes.io/dockerconfigjson
+
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: korifi-installer
+  namespace: korifi-installer
+
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: korifi-installer
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-admin
+subjects:
+- kind: ServiceAccount
+  name: korifi-installer
+  namespace: korifi-installer
+
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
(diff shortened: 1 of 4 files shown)


beyhan:
It will be really useful to have also a concrete none docker image test app reference like the docker one you are adding with this pr `cf push nginx --docker-image nginxinc/nginx-unprivileged`.

georgethebeatle:
@beyhan We have added an example for pushing buildpack apps as well

beyhan:
Thanks for adding the buildpack example! On my machine it fails with:
```
...
   Loading registry credentials from service account secrets
   Loading secret for "localregistry-docker-registry.default.svc.cluster.local:30050" from secret "image-registry-credentials" at location "/var/build-secrets/image-registry-credentials"
   Loading cluster credential helpers
   Pulling localregistry-docker-registry.default.svc.cluster.local:30050/7d908585-54d2-4f0d-8135-bab2d304eb40-packages@<some-sha>...
   Successfully pulled localregistry-docker-registry.default.svc.cluster.local:30050/7d908585-54d2-4f0d-8135-bab2d304eb40-packages@<some-sha> in path "/workspace"
Error staging application dorifi: timed out after 15 minute(s)
FAILED
```   

Additionally, I think that having teardown step for the Korifi kind cluster would be also useful.

beyhan:
Additional point here. I found the root cause for https://github.com/cloudfoundry/korifi/issues/2912. If you try to push the version `1.23.2` of the `nginxinc/nginx-unprivileged` to CF with `cf push nginx-unprivileged --docker-image nginxinc/nginx-unprivileged:1.23.2` it will work. Pushing the next version `1.23.3` fails with the error described in https://github.com/cloudfoundry/korifi/issues/2912. Looking into the image manifest with `docker manifest inspect --verbose nginxinc/nginx-unprivileged:1.23.3` it is the first version where they attach image [attestation information to the manifest](https://docs.docker.com/build/attestations/attestation-storage/). It looks like Diego can’t deal with this. This is definitely something which has to be addressed in Diego but I would suggest to select a docker based example which could be ran on CF also.

beyhan:
@georgethebeatle I was able to execute the installation and push the sample apps without any issue.

scripts/installer/install-korifi-kind.yaml
```diff
@@ -0,0 +1,106 @@
...
+          # e.g. `docker run --entrypoint htpasswd httpd:2 -Bbn user password`
+          #
+          helm upgrade --install localregistry twuni/docker-registry \
+            --namespace default \
+            --set service.type=NodePort,service.nodePort=30050,service.port=30050 \
+            --set persistence.enabled=true \
+            --set persistence.deleteEnabled=true \
+            --set secrets.htpasswd='user:$2y$05$Ue5dboOfmqk6Say31Sin9uVbHWTl8J1Sgq9QyAEmFQRnq1TPfP1n2'
+
+          helm upgrade --install korifi https://github.com/cloudfoundry/korifi/releases/download/v0.9.0/korifi-0.9.0.tgz \
```
davewalter, line 89: Does it make sense for the script to look up the latest Korifi release version instead of hard-coding it?

scripts/installer/install-korifi-kind.yaml
```diff
@@ -0,0 +1,106 @@
...
+spec:
+  template:
+    metadata:
+      name: install-korifi
+    spec:
+      serviceAccountName: korifi-installer
+      restartPolicy: Never
+      containers:
+      - name: install-korifi
+        image: eiriniuser/korifi-install
```
davewalter, line 71: I assume this will be replaced with the new `cloudfoundry/korifi-installer` image?

davewalter (commented):
Looks good overall. I posted a couple of questions inline.

scripts/installer/install-korifi-kind.yaml
danail-branekov: Yes!

Hardcoding it here was the easiest thing to do in order to showcase the idea. We should either lookup the latest release or have the CI produce a release-specific job yaml release artifact, we have not figured it out yet. This is going to be the next thing to do before merging now that we see that people like the idea.

scripts/installer/install-korifi-kind.yaml
danail-branekov: Yes :)

PR #2920, opened 2023-10-06
Introduce a systemImagePullSecrets helm value

georgethebeatle:
No

## What is this change about?
This value would allow the korifi api and controller images to be pulled from a
private registry. This can be useful in corporate environments if an internal
mirror registry is being used as well as in environments with no internet.

No

## Acceptance Steps
N/A

@cloudfoundry/wg-cf-on-k8s


julian-hj (approved):
LGTM

Issue #2749, opened 2023-08-02
As a Korifi API user I can request each api endpoint with a trailing slash


danail-branekov:
## Background

At the moment korifi rejedts all requests with a trailing slash. However traditional cf accepts both and serves valid responses

## Acceptance

**GIVEN** I have korifi deployed
**WHEN** I request any endpoint (i.e. /v3/organizations)
**THEN** I see a list of all orgs
**WHEN** I request the same endpoint with a trainling slash (i.e. /v3/organizations/)
**THEN** I see an identidal response

tcdowney:
Here's a note on what we can do for this:
https://github.com/go-chi/chi/issues/41

alperdedeoglu:
Hi @tcdowney, @danail-branekov,

First of all thanks for Korifi project, looks really promising. I'd like to suggest an alternative approach to the one shared by @tcdowney in [this issue](https://github.com/go-chi/chi/issues/41). In my opinion, the following proposal could be a more streamlined and quicker implementation:

```go
package main

import (
	"net/http"

	"github.com/go-chi/chi"
	"github.com/go-chi/chi/middleware"
)

func main() {
	r := chi.NewRouter()
	r.Use(middleware.StripSlashes)
	r.Get("/v3/apps", appsHandler)
	r.Get("/v3/apps/{guid}", guidHandler)
	http.ListenAndServe("localhost:8080", r)
}

func appsHandler(w http.ResponseWriter, r *http.Request) {
	w.Write([]byte("Apps Returned"))
}
func guidHandler(w http.ResponseWriter, r *http.Request) {
	w.Write([]byte("Guid Returned"))
}

```
I have tested via following commands:
```sh
curl http://localhost:8080/v3/apps/
curl http://localhost:8080/v3/apps
curl http://localhost:8080/v3/apps/123
```
And the result seems to be working as expected. 
```sh
Apps Returned
Apps Returned
Guid Returned
```
First two requests are handled identically, but the last one is of course handled differently.

What do you think? Is this approach also feasible?

danail-branekov:
Hi @alperdedeoglu 

[...]

PR #3080, opened 2024-01-24
Fix for issue #2749 - trailing slash request handling

alperdedeoglu:
[Related Issue](https://github.com/cloudfoundry/korifi/issues/2749)

## What is this change about?
Handling the requests with trailing slash in Korifi 

No

## Acceptance Steps
Test can be found under api/routing/router_test.go
Request to **/hello/world** and **/hello/world/** should behave identical.

@danail-branekov 

<!--
## Things to remember
Standard chi middleware has been used for removing the trailing slashes from requests as discussed [here.](https://github.com/cloudfoundry/korifi/issues/2749)

### DIFF
--- a/api/routing/router_test.go
+++ b/api/routing/router_test.go
@@ -164,6 +164,7 @@ var _ = Describe("Router", func() {
 			Expect(res).To(HaveHTTPBody(MatchJSON(`{"not":"allowed"}`)))
 		})
 	})
+
 })
 
 func mkReq(handler http.Handler, method, url string) (*http.Response, error) {
--- a/tests/e2e/root_v3_test.go
+++ b/tests/e2e/root_v3_test.go
@@ -1,0 +1,47 @@
+package e2e_test
+
+import (
+	"net/http"
+
+	"github.com/go-resty/resty/v2"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+type RootV3Resource struct {
+	Links struct {
+		Self struct {
+			Href string `json:"href"`
+		} `json:"self"`
+	} `json:"links"`
+}
+
+var _ = Describe("RootV3", func() {
+	var (
+		httpResp    *resty.Response
+		requestPath string
+	)
+
+	BeforeEach(func() {
+		requestPath = "/v3"
+	})
+	JustBeforeEach(func() {
+		var err error
+		httpResp, err = adminClient.R().
+			Get(requestPath)
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	It("succeeds", func() {
+		Expect(httpResp).To(HaveRestyStatusCode(http.StatusOK))
+	})
+
+	When("Request ends with a trailing slash", func() {
+		BeforeEach(func() {
+			requestPath = "/v3/"
+		})
+		It("succeeds", func() {
+			Expect(httpResp).To(HaveRestyStatusCode(http.StatusOK))
+		})
+	})
+})
--- a/api/main.go
+++ b/api/main.go
@@ -29,6 +29,7 @@ import (
 	toolsregistry "code.cloudfoundry.org/korifi/tools/registry"
 	"code.cloudfoundry.org/korifi/version"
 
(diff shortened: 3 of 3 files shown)


danail-branekov:
Hi @alperdedeoglu,

Thaks for your contribution. The code itself looks allright, however we are concerned that the test in the `router_test.go` does not cover the change in `api/main.go`, but rather repeats it. If we accidentally remove that middleware from the main.go no test will catch that. It would make more sense if we introduced an e2e test under `tests/e2e`. It could be a test for the `api/handlers/root_v3.go` (the `/v3`) endpoint. The test could simply request both `/v3` and `/v3/` to make sure that the respective middleware is in place without careing too much for the payload (as the payload is tested in the handler unit test). This would mean introducing a new e2e test file. You can take inspiration from one of the other available e2e tests (`tests/e2e/whoami_test.go` is a good example).

Apologies for misguiding you in the issue.

api/routing/router_test.go
```diff
@@ -164,6 +165,28 @@ var _ = Describe("Router", func() {
 			Expect(res).To(HaveHTTPBody(MatchJSON(`{"not":"allowed"}`)))
 		})
 	})
+
+	When("a request comes with trailing slash", func() {
```
danail-branekov, line 169: I believe that this context is testing whether the trailing slash middleware from chi is working correctly and making sure that this is the case is out of scope for Korifi tests. That should have been tested by chi itself.

From my point of view, having an e2e test that makes sure that requests with trailing slash do not fail should be enough. TL;DR I think you could remove the `When("a request comes with trailing slash"` context

tests/e2e/root_v3_test.go
```diff
@@ -0,0 +1,45 @@
...
+
+type RootV3Resource struct {
+	Links struct {
+		Self struct {
+			Href string `json:"href"`
+		} `json:"self"`
+	} `json:"links"`
+}
+
+var _ = Describe("TrailingSlash", func() {
```
danail-branekov, line 19: I would recommend several changes to the structure of the test:
1) change the `Describe` test to something like `RootV3` as this test is supposed to cover the `/v3` endpoint
2) Introduce a variable e.g. `requestPath` that is defaulted to `/v3` in `BeforeEach` and have a single request in `JustBeforeEach` for that path. In the top-level `It` just verify that the request succeeds.
3) Introduce a context e.g. `When("the request path ends with a trailing slack"....` whose `BeforeEach` is resetting the `requestPath` to `/v3/`. Then, when the top-level`JustBeforeEach` is executed it would pick up the value with the trailing slash. In that `When` there should be another `It` that verifies that the request still succeeds.

Following this structure would make the test comply with the standard Ginkgo idiom where there is preferably a single `JustBeforeEach` that exercises the code under test, and then multiple contexts with their `BeforeEach`es that alter the arguments to the invoccation under test.

tests/e2e/root_v3_test.go
```diff
@@ -0,0 +1,45 @@
...
+	var (
+		httpRespWithSlash    *resty.Response
+		httpRespWithoutSlash *resty.Response
+		resultWithoutSlash   RootV3Resource
+		resultWithSlash      RootV3Resource
+	)
+
+	When("Sending request to /v3 and /v3/", func() {
```
danail-branekov, line 27: As a confirmation to the test restructuring recommendation above, this `When` says `Sending request to /v3 and /v3/`. The fact that there is an `and` in the description tells that these are actually two contexts. Putting them in a single `When` sounds like a smell to me.

danail-branekov:
Hey @alperdedeoglu 
The testing approach looks good, I have put some recommendations on improving the tests. Once you do that, could you also make sure that you squash all the commits into one so that we have the feature with its tests in a single commit?

tests/e2e/root_v3_test.go
alperdedeoglu: Done

api/routing/router_test.go
alperdedeoglu: Done

danail-branekov (approved):

PR #3126, opened 2024-02-16
add dynamic provisioning for contour

spgreenberg:

No

## What is this change about?

DISCLAIMER: This is purely optional. Feel free to simply discard this PR if you want to stick to static provisioning of contour.

I was not able to get static provisioning of contour with a custom domain and certificates to work. The instructions are confusing and I am far from an expert. However, dynamic provisioning was quite simple and straightforward. 

No. Documentation only.

## Acceptance Steps

1. Follow the dynamic provisioning directions in the install guide. 
2. Verify load balancer DNS in the install guide.

NOTE: I did not change the static provisioning instructions as I was not able to get them to work.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -59,7 +59,11 @@ The Helm chart will create an example Kpack `ClusterBuilder` (with the associate
 
 ### Contour
 
-[Contour](https://projectcontour.io/) is our [ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/) controller. Follow the static provisioning [instructions](https://projectcontour.io/docs/1.26/config/gateway-api/#static-provisioning) from the Gateway API support guide to install the latest version. Note that as part of the Contour installation you have to create a gatewayclass with name `$GATEWAY_CLASS_NAME`:
+[Contour](https://projectcontour.io/) is our [ingress](https://kubernetes.io/docs/concepts/services-networking/ingress/) controller. Contour implements the [Gateway API](https://gateway-api.sigs.k8s.io/). There are two ways to deploy Contour with Gateway API support: static provisioning and dynamic provisioning.
+
+#### Static Provisioning
+
+Follow the static provisioning [instructions](https://projectcontour.io/docs/1.26/config/gateway-api/#static-provisioning) from the Gateway API support guide to install the latest version. Note that as part of the Contour installation you have to create a gatewayclass with name `$GATEWAY_CLASS_NAME`:
 
 ```bash
 kubectl apply -f - <<EOF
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -74,6 +78,23 @@ EOF
...
+    kubectl apply -f - <<EOF
+    kind: GatewayClass
+    apiVersion: gateway.networking.k8s.io/v1beta1
+    metadata:
+      name: $GATEWAY_CLASS_NAME
+    spec:
+      controllerName: projectcontour.io/gateway-controller
+    EOF
+    ```
+  - You DO NOT need to create a gateway class as per the instructions. The Korifi helm chart defines a gateway that will be used for all korifi ingress traffic. The gateway will be created in the `korifi-gateway` namespace.
```
danail-branekov, line 96: I find those two bullets a bit contradicting - do users have to create the gateway class, or not? Maybe the section should just say what users have to do.

danail-branekov:
Thanks for the PR, @spgreenberg 
This is quite useful. As a matter of fact, we could not get dynamic provisioning to work ourselves, that is why we documented just the static one. We will give your instructions a try.

INSTALL.md
danail-branekov: After re-reading the documentation, I believe that in the second bullet you meant that users do not have to create the **gateway** (instead of the ~~gateway class~~), correct?

danail-branekov:
@spgreenberg we did manage to get dynamic provisioning working following your instructions. FWIW, in order to get it working on kind we had to set the contour service type to `NodePort` (as per [contour documentation](https://projectcontour.io/docs/1.24/deploy-options/#nodeport-service)).

Besides the `gateway` vs `gateways class` comment above, the change looks good.

INSTALL.md
spgreenberg: @danail-branekov You are correct. Sorry about that.

spgreenberg:
> @spgreenberg we did manage to get dynamic provisioning working following your instructions. FWIW, in order to get it working on kind we had to set the contour service type to `NodePort` (as per [contour documentation](https://projectcontour.io/docs/1.24/deploy-options/#nodeport-service)).
> 
> Besides the `gateway` vs `gateways class` comment above, the change looks good.

That is strange. I did not have to set the contour service type. I am using GKE with custom domains and Let's Encrypt certs.

INSTALL.md
spgreenberg: Corrected.

danail-branekov (approved):

danail-branekov:
> That is strange. I did not have to set the contour service type. I am using GKE with custom domains and Let's Encrypt certs.

On a kind cluster there is no load balancer by default, that's why there is nothing to reconcile the LoadBalancer service. On managed environments (such as GKE), a load balancer is already provided with the cluster.

PR #3156, opened 2024-03-07
Add a seccompProfile to the pod

shanman190:
N/A

## What is this change about?
Presently Korifi places a seccompProfile on the container, while this works for the container when you want to layer in a service mesh such as Istio you run into a few issues. During the Korifi installation, if you follow the install guide as described it has you place a `pod-security.kubernetes.io/enforce=restricted` on the root namespace, then this cascades to created organization and space namespaces. When this happens, for a space that desires to enable the Istio service mesh, the `istio-init` and `istio-validation` sidecar containers are started with the default seccompProfile which results in the entire pod being unable to be scheduled due to two containers wanting additional privileges. Istio has support for it's own custom, chained CNI which removes the need for the additional privileges in these more secure environments, but during it's mutation it strictly observes a seccompProfile placed upon the pod itself.

No breaking changes.

## Acceptance Steps
Bring up a Korifi environment utilizing Istio with `pod-security.kubernetes.io/enforce=restricted` on the space namespaces and the API Gateway integration.
_profile.yaml_
```
apiVersion: install.istio.io/v1alpha1
kind: IstioOperator
spec:
  profile: minimal
  components:
    cni:
      enabled: true
  meshConfig:
    accessLogFile: /dev/stdout
    outboundTrafficPolicy:
      mode: ALLOW_ANY
  values:
    cni:
      excludeNamespaces:
        - istio-system
        - kube-system
        - kpack
      logLevel: debug
    pilot:
      env:
        PILOT_ENABLE_ALPHA_GATEWAY_API: true
```

```
istioctl install --file profile.yaml
```

N/A

## Additional Information
I found the reference to putting the seccompProfile on the pod in the Istio project's issue comments. You can find the specific comment here: https://github.com/istio/istio/issues/35894#issuecomment-1511634924. I've also verified upstream that kpack is already placing a seccompProfile on both the pod and the build container.

### DIFF
--- a/job-task-runner/controllers/integration/taskworkload_controller_test.go
+++ b/job-task-runner/controllers/integration/taskworkload_controller_test.go
@@ -77,6 +77,9 @@ var _ = Describe("Job TaskWorkload Controller Integration Test", func() {
 		Expect(podSpec.RestartPolicy).To(Equal(corev1.RestartPolicyNever))
 		Expect(podSpec.SecurityContext).To(Equal(&corev1.PodSecurityContext{
 			RunAsNonRoot: tools.PtrTo(true),
+			SeccompProfile: &corev1.SeccompProfile{
+				Type: corev1.SeccompProfileTypeRuntimeDefault,
+			},
 		}))
 		Expect(podSpec.AutomountServiceAccountToken).To(Equal(tools.PtrTo(false)))
 		Expect(podSpec.ImagePullSecrets).To(ConsistOf(corev1.LocalObjectReference{Name: "my-image-secret"}))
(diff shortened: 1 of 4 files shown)


danail-branekov (approved):

danail-branekov:
Thank you for the PR, @shanman190!

That is quite interesting, we were not aware that istio sidecars could run in restricted namespaces, good to know!

georgethebeatle:
Hey @shanman190 looks like there is one more test that needs to be adapted: https://github.com/cloudfoundry/korifi/actions/runs/8191034479/job/22431280462?pr=3156

Can you please fix this, so that we can merge your change?

shanman190:
The failing test has now been fixed. Thanks @danail-branekov and @georgethebeatle!

danail-branekov (dismissed):

shanman190:
Ahh, ok, I see now that my other new test isn't actually valid and is replaced by the other integration test that had failed. I'll remove the current failing test and that should get this PR fixed up finally.

danail-branekov:
> Ahh, ok, I see now that my other new test isn't actually valid and is replaced by the other integration test that had failed. I'll remove the current failing test and that should get this PR fixed up finally.

Thanks! Could you also take the opportunity to squash all the commits into one?

shanman190:
Will do!

shanman190:
Squashed and pushed. Things should be good now. Apologies about the couple of test bugs.

danail-branekov (approved):

PR #3206, opened 2024-04-04
added fake getAppFeature in apps handler and related tests for issue-2357

marsteg:
https://github.com/cloudfoundry/korifi/issues/2357

## What is this change about?
I'm adding a new API endpoint that will reply the requested feature and show it as not enabled. This is for consistency with classical CF. See more details in https://github.com/cloudfoundry/korifi/issues/2357

Not as far as I can see. 

## Acceptance Steps
GIVEN I have pushed an app to korifi
WHEN I GET /v3/apps/:guid/features/:name
THEN I see the following response
```
HTTP/1.1 200 OK

{
  "name": "feature-name",
  "description": "fature description",
  "enabled": false
}
```

<!--
## Things to remember
I stole most of this from our internal fork but changed the "PATCH" there into "GET" as requested in the Issue.

### DIFF
--- a/api/handlers/app.go
+++ b/api/handlers/app.go
@@ -34,6 +34,7 @@ const (
 	AppRestartPath                    = "/v3/apps/{guid}/actions/restart"
 	AppEnvVarsPath                    = "/v3/apps/{guid}/environment_variables"
 	AppEnvPath                        = "/v3/apps/{guid}/env"
+	AppFeaturePath                    = "/v3/apps/{guid}/features/{name}"
 	AppPackagesPath                   = "/v3/apps/{guid}/packages"
 	AppSSHEnabledPath                 = "/v3/apps/{guid}/ssh_enabled"
 	invalidDropletMsg                 = "Unable to assign current droplet. Ensure the droplet exists and belongs to this app."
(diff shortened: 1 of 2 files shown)


api/handlers/app.go
```diff
@@ -606,6 +607,33 @@ func (h *App) getSSHEnabled(r *http.Request) (*routing.Response, error) {
...
+			"enabled":     false,
+		}), nil
+	case "revisions":
+		return routing.NewResponse(http.StatusOK).WithBody(map[string]any{
+			"name":        "revisions",
+			"description": "Enable versioning of an application",
+			"enabled":     false,
+		}), nil
+	default:
+		return routing.NewResponse(http.StatusOK).WithBody(map[string]any{
```
danail-branekov, line 626: I guess we should return some sort of an HTTP error code here instead of OK. Maybe the same one as on CF for VMs

marsteg: agreed (totally missed that). Updated the PR accordingly.

api/handlers/app_test.go
```diff
@@ -1653,6 +1653,55 @@ var _ = Describe("App", func() {
 			)))
 		})
 	})
+
+	Describe("GET /v3/apps/GUID/features/ssh", func() {
```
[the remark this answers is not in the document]
marsteg: grouped three WHEN contexts into one Describe. PR is updated.

api/handlers/app.go
```diff
@@ -606,6 +607,32 @@ func (h *App) getSSHEnabled(r *http.Request) (*routing.Response, error) {
...
+			"enabled":     false,
+		}), nil
+	case "revisions":
+		return routing.NewResponse(http.StatusOK).WithBody(map[string]any{
+			"name":        "revisions",
+			"description": "Enable versioning of an application",
+			"enabled":     false,
+		}), nil
+	default:
+		return routing.NewResponse(http.StatusNotFound).WithBody(map[string]any{
```
danail-branekov, line 626: final nitpick:

Instead of returning a response containing an error, the handler here could just return the [special api not-found error](https://github.com/cloudfoundry/korifi/blob/3820a5a049d4b6169e0e29ac896b913626053ea8/api/errors/errors.go#L157) that is going to be presented in a generic manner to the user . Then the switch could be refactored into

```
if featureName == "ssh" {
  return ....
}

if featureName == "revisions" {
  return ...
}

return apierrors.NewNotFoundError(nil, "Feature")
```

danail-branekov (approved):

api/handlers/app.go
marsteg: adapted accordingly

PR #3230, opened 2024-04-23
added environment variable CF_INSTANCE_PORTS

marsteg:
https://github.com/cloudfoundry/korifi/issues/3229

## What is this change about?
Adding environment variable `CF_INSTANCE_PORTS` to workloads

not, that i expect it

## Acceptance Steps
- deploy an app, that prints the environment variable CF_INSTANCE_PORTS
- check that the app actually has the environment variable

## Additional info
I've seen strange errors when executing `go test -v`: 12:04:37.313765   45784 reflector.go:462] pkg/mod/k8s.io/client-go@v0.29.3/tools/cache/reflector.go:229: watch of *v1.Secret ended with: an error on the server ("unable to decode an event from the watch stream: context canceled") has prevented the request from succeeding - i hope that this is some local issue...

### DIFF
--- a/controllers/controllers/workloads/env/builder.go
+++ b/controllers/controllers/workloads/env/builder.go
@@ -138,9 +138,11 @@ func (b *ProcessEnvBuilder) buildPortEnv(ctx context.Context, cfApp *korifiv1alp
 	processPorts := ports.FromRoutes(cfRoutesForProcess.Items, cfApp.Name, cfProcess.Spec.ProcessType)
 	if len(processPorts) > 0 {
 		portString := strconv.FormatInt(int64(processPorts[0]), 10)
+		cfInstancePorts := fmt.Sprintf("[{\"internal\":%s}]", portString)
 		return []corev1.EnvVar{
 			{Name: "VCAP_APP_PORT", Value: portString},
 			{Name: "PORT", Value: portString},
+			{Name: "CF_INSTANCE_PORTS", Value: cfInstancePorts},
 		}, nil
 	}
 
--- a/controllers/controllers/workloads/env/builder_test.go
+++ b/controllers/controllers/workloads/env/builder_test.go
@@ -350,9 +350,12 @@ var _ = Describe("EnvBuilder", func() {
 						"Name":  Equal("PORT"),
 						"Value": Equal("1234"),
 					}),
+					MatchFields(IgnoreExtras, Fields{
+						"Name":  Equal("CF_INSTANCE_PORTS"),
+						"Value": MatchJSON("[{\"internal\":1234}]"),
+					}),
 				))
 			})
-
 			When("the route does not have destinations", func() {
 				BeforeEach(func() {
 					ensurePatch(cfRoute, func(cfRoute *korifiv1alpha1.CFRoute) {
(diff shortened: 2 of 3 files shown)


controllers/controllers/workloads/env/builder.go
```diff
@@ -138,9 +138,11 @@ func (b *ProcessEnvBuilder) buildPortEnv(ctx context.Context, cfApp *korifiv1alp
 	processPorts := ports.FromRoutes(cfRoutesForProcess.Items, cfApp.Name, cfProcess.Spec.ProcessType)
 	if len(processPorts) > 0 {
 		portString := strconv.FormatInt(int64(processPorts[0]), 10)
+		cfInstancePorts := fmt.Sprintf("[{\"internal\":%s},{\"internal\":%s}]", portString, portString)
```
danail-branekov, line 141: Why do we need `internal` twice? I guess we need `external` as well?

marsteg: very good question! So i basically copied what i saw that a productive CF instance gave me by default. I got there:

CF_INSTANCE_PORTS env is: [{"internal":8080,"external_tls_proxy":61157,"internal_tls_proxy":61001},{"internal":8080,"internal_tls_proxy":61443}]

so my assumption is, that the first {} holds ALL ports and the second {} holds all "internal" ports. since I didn't know what the tls_proxy ports are in the korifi environment, I left them out for now.

danail-branekov: Could it be that in the example above you have had two routes for the same app? Docs for that env var do not seem to mention anything about the number of json entries.

marsteg: no, there is just one route when I do "cf routes". Could the route be hidden somehow? i didn't specifically create one...

danail-branekov: I have spun a traditional cf deployment, pushed dora from CATS and got

```
vcap@a0213476-fa1a-43c1-5ec8-99af:~$ echo $CF_INSTANCE_PORTS  | jq .
[
  {
    "external": 61000,
    "internal": 8080,
    "external_tls_proxy": 61002,
    "internal_tls_proxy": 61001
  },
  {
    "external": 61000,
    "internal": 8080,
    "internal_tls_proxy": 61443
  },
  {
    "external": 61001,
    "internal": 2222,
    "external_tls_proxy": 61003,
    "internal_tls_proxy": 61002
  }
]

``` 

weird...

marsteg: Talked to a CF developer:
proposal: once external_tls_proxy and internal_tls_proxy gets added, we could simply add the internal only (external ones not exposed if envoy is active). For now, internal should be enough but there seem to be apps, that rely on the existance of all fields

controllers/controllers/workloads/env/builder_test.go
```diff
@@ -350,6 +350,10 @@ var _ = Describe("EnvBuilder", func() {
 						"Name":  Equal("PORT"),
 						"Value": Equal("1234"),
 					}),
+					MatchFields(IgnoreExtras, Fields{
+						"Name":  Equal("CF_INSTANCE_PORTS"),
+						"Value": ContainSubstring("\"internal\":1234"),
```
danail-branekov, line 355: How about we use [`MatchJSON`](https://onsi.github.io/gomega/#matchjsonjson-interface) instead of `ContainSubstring`? thus we could ensure that the content of the value is a valid json

marsteg: Done! I suppose we could also create a struct and json.Marshall it... but i'm note sure if that's better as json shouldn't change?

PR #3232, opened 2024-04-23
adapted regex to also allow small letter m, mb, t, tb, g and gb.

marsteg:
https://github.com/cloudfoundry/korifi/issues/3231

## What is this change about?
support small letters in memory definition, analog to original CF [spec](https://v3-apidocs.cloudfoundry.org/version/3.156.0/index.html#the-manifest-schema) 

not that i'm aware of.

## Acceptance Steps
push an app, with memory values like:
128m
128mb
128g
128gb
128t
128tb

@danail-branekov  @georgethebeatle

### DIFF
--- a/api/payloads/manifest.go
+++ b/api/payloads/manifest.go
@@ -264,7 +264,7 @@ func (s ManifestApplicationService) Validate() error {
 	return validation.ValidateStruct(&s, validation.Field(&s.Name, validation.Required))
 }
 
-var unitAmount = regexp.MustCompile(`^\d+(?:B|K|KB|M|MB|G|GB|T|TB)$`)
+var unitAmount = regexp.MustCompile(`^\d+(?:B|K|KB|M|m|MB|mb|G|g|GB|gb|T|t|TB|tb)$`)
 
 func validateAmountWithUnit(value any) error {
 	v, isNil := validation.Indirect(value)
@@ -273,7 +273,7 @@ func validateAmountWithUnit(value any) error {
 	}
 
 	if !unitAmount.MatchString(v.(string)) {
-		return errors.New("must use a supported unit (B, K, KB, M, MB, G, GB, T, or TB)")
+		return errors.New("must use a supported unit (B, K, KB, M, m, MB, mb, G, g, GB, gb, T, t, TB or tb)")
 	}
 
 	mbs, err := bytefmt.ToMegabytes(v.(string))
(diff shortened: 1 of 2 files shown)


danail-branekov (dismissed):

api/payloads/manifest.go
```diff
@@ -264,7 +264,7 @@ func (s ManifestApplicationService) Validate() error {
 	return validation.ValidateStruct(&s, validation.Field(&s.Name, validation.Required))
 }
 
-var unitAmount = regexp.MustCompile(`^\d+(?:B|K|KB|M|MB|G|GB|T|TB)$`)
```
danail-branekov, line 267: Have you verified what is the effect of accepting lower case letters?

If Korifi translates those units directly to k8s, then this might have unwanted behaviour, for example:

```
Pay attention to the case of the suffixes. If you request 400m of memory, this is a request for 0.4 bytes. Someone who types that probably meant to ask for 400 mebibytes (400Mi) or 400 megabytes (400M).

```
https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#meaning-of-memory

marsteg: forget my previous comment - obviously i'm getting mebibytes... i'll have another look

georgethebeatle: Maybe a stupid question, but shouldn't we also accept `b`, `k` and `kb`? The v3 docs state that all units can be in upper or lower case.

marsteg: Good point - i will also add b, k and kb.

I found that we hardcode our request and limits to mebibyte... is this on purpose? to cleanly fix this, i think, i would also adapt them to reflect actual MegaByte. What do you think? Is the cfprocess_controller.go the right place to implement this?

https://github.com/cloudfoundry/korifi/blob/f60426c07faa7bd37bbff1bc8794eae21c276cf5/controllers/controllers/workloads/cfprocess_controller.go#L314

danail-branekov: I have no memory why we have preferred mebibytes. I have posted a [question](https://cloudfoundry.slack.com/archives/C0297673ASK/p1714044427954889) on slack in case anyone remembers

danail-branekov: Here is the issue that describes that decision. Apparently docs are wrong :)
https://github.com/cloudfoundry/korifi/issues/720

danail-branekov: I will approve this PR in order not to age. If anyone needs the other units (`b`, `k`, `kb`) we could introduce them later.

danail-branekov (approved):

PR #3239, opened 2024-05-01
Add initial /v3/info endpoint

gogolok:

https://github.com/cloudfoundry/korifi/issues/2914

## What is this change about?

Add initial `v3/info` endpoint that presents the fields as implmented in
https://github.com/cloudfoundry/cloud_controller_ng/blob/main/app/controllers/v3/info_controller.rb
and
https://github.com/cloudfoundry/cloud_controller_ng/blob/main/app/presenters/v3/info_presenter.rb
with some static default values for the time being.

This change is the first step in getting a working `v3/info` endpoint implementation.
The next step would be to extend Korifi via configuration to provide some dynamic content (For example the `description` field and others) to be returned by the endpoint `v3/info`.

No, it adds a new endpoint not available before.

## Acceptance Steps

- Switch to the Git branch.
- Setup local dev environment, `./scripts/deploy-on-kind.sh korifi`
- cf api https://localhost --skip-ssl-validation
- cf login
- Run `cf curl v3/info` and see the following static content (except for the base url):

```
{"build":"","cli_version":{"minimum":"","recommended":""},"description":"","name":"","version":0,"custom":{},"links":{"self":{"href":"https://localhost/v3/info"},"support":{"href":"https://www.cloudfoundry.org/technology/korifi/"}}}
```

None

### DIFF
--- a/api/handlers/info_v3.go
+++ b/api/handlers/info_v3.go
@@ -1,0 +1,37 @@
+package handlers
+
+import (
+	"net/http"
+	"net/url"
+
+	"code.cloudfoundry.org/korifi/api/presenter"
+	"code.cloudfoundry.org/korifi/api/routing"
+)
+
+const (
+	InfoV3Path = "/v3/info"
+)
+
+type InfoV3 struct {
+	baseURL url.URL
+}
+
+func NewInfoV3(baseURL url.URL) *InfoV3 {
+	return &InfoV3{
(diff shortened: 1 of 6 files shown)


georgethebeatle (approved):

georgethebeatle:
Thanks @gogolok, looking good!

danail-branekov (approved):

PR #3240, opened 2024-05-02
Introduce Ready condition for cfServiceInstances

georgethebeatle:
No

## What is this change about?
- Replace the existing CredentialSecretAvailable condition with Ready in
  controller and repository
- Remove one of the two constants for the Ready condition
- The awaiter test now works with Ready condition (unimportant, but this
  way we can better understand what resources are not yet using the
  Ready condtion

No

## Acceptance Steps
Green tests

@danail-branekov

### DIFF
--- a/controllers/controllers/services/instances/controller.go
+++ b/controllers/controllers/services/instances/controller.go
@@ -43,8 +43,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
-const CredentialsSecretAvailableCondition = "CredentialSecretAvailable"
-
 type Reconciler struct {
 	k8sClient client.Client
 	scheme    *runtime.Scheme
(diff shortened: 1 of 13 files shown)


controllers/controllers/services/instances/controller.go
```diff
@@ -125,37 +125,24 @@ func (r *Reconciler) ReconcileResource(ctx context.Context, cfServiceInstance *k
...
 	if err != nil {
-		log.Error(err, "failed to reconcile credentials secret")
-		meta.SetStatusCondition(&cfServiceInstance.Status.Conditions, metav1.Condition{
-			Type:               CredentialsSecretAvailableCondition,
-			Status:             metav1.ConditionFalse,
-			Reason:             "FailedReconcilingCredentialsSecret",
-			Message:            err.Error(),
-			ObservedGeneration: cfServiceInstance.Generation,
-		})
+		readyConditionBuilder.
```
danail-branekov, line 128: I wonder whether returning an error that wraps the original one but providing additional details would be more "idiomatic". The `defer` statement in the beginning of the reconcile method would handle the error and set the `Ready` condition to `false` anyway

danail-branekov (approved):

PR #3248, opened 2024-05-12
Dynamic v3/info endpoint

gogolok:

https://github.com/cloudfoundry/korifi/issues/2914

## What is this change about?

Extend Korifi configuration to provide dynamic content to the endpoint `v3/info`.

The previous step was to introduce static content:
https://github.com/cloudfoundry/korifi/pull/3239

No

## Acceptance Steps

Switch to this PR's branch, set the `infoConfig` configuration in `helm/korifi/values.yaml`
to some nonempty values, run `./scripts/deploy-on-kind.sh korifi` and check `cf curl v3/info` after loggin into Korifi via `cf` CLI.

None

## Further Information

- The CLI version information properties has been named after https://github.com/cloudfoundry/capi-release/blob/develop/jobs/cloud_controller_ng/spec#L397 .
- Not sure whether the InfoConfig should have been passed via repository pattern to the v3 info endpoint handler?


api/config/config.go
```diff
@@ -61,6 +63,17 @@ type (
...
 		StagingMemoryMB int    `yaml:"stagingMemoryMB"`
 	}
+
+	InfoConfig struct {
+		Build                    string                 `yaml:"build"`
+		Description              string                 `yaml:"description"`
+		Name                     string                 `yaml:"name"`
+		Version                  int                    `yaml:"version"`
+		MinCLIVersion            string                 `yaml:"minCLIVersion"`
+		MinRecommendedCLIVersion string                 `yaml:"minRecommendedCLIVersion"`
```
danail-branekov, line 73: According to the [CF API docs](https://v3-apidocs.cloudfoundry.org/version/3.164.0/index.html#get-platform-info), 

```
"cli_version": {
    "minimum": "6.22.0",
    "recommended": "latest"
  },
```

With respect to the `recommended` attribute above, it probably makes sense to rename `MinRecommendedCLIVersion` to `RecommendedCLIVersion`. Having the `Min` prefix is a bit misleading.

helm/korifi/values.yaml
```diff
@@ -44,6 +44,16 @@ api:
       idle: 900
       readHeader: 10
 
+  infoConfig:
+    build: ""
+    description: ""
+    name: ""
+    version: 0
+    minCLIVersion: ""
+    minRecommendedCLIVersion: ""
```
danail-branekov, line 53: Same as above, renaming `minRecommendedCLIVersion` to `recommendedCLIVersion` would probably make sense

helm/korifi/values.yaml
```diff
@@ -44,6 +44,16 @@ api:
       idle: 900
       readHeader: 10
 
+  infoConfig:
+    build: ""
+    description: ""
+    name: ""
+    version: 0
```
danail-branekov, line 51: Providing the version as a helm value sounds a bit tricky as operators would be able to configure it at their will. What we already do for api and controllers binaries is to make the version available for logging purposes (see [this](https://github.com/cloudfoundry/korifi/blob/c353c9ecd6f55a14799d22bedfaf5a822d15d8cc/version/version.go#L12) variable). Therefore, the /v3/info presenter should return whatever that values is.

helm/korifi/values.yaml
```diff
@@ -44,6 +44,16 @@ api:
       idle: 900
       readHeader: 10
 
+  infoConfig:
+    build: ""
```
danail-branekov, line 48: The API docs are not really clear on what `build` semantic is. Is there a use-case where this value is used? If not, we would prefer a lean approach and only put stuff that are actually needed.

Other than that, the docs example value sounds like a commit SHA. If that is the case, the same approach as to the `version` should be applied (i.e. should not be configurable via a helm value)

danail-branekov:
Hey @gogolok 

A general question: do you know use cases where each property of the `/v3/info` endpoint is used? We would refrain from implementing things that noone uses, especially things that are not clear when reading the CF API spec (which is the case with `/v3/info`)

helm/korifi/values.yaml
```diff
@@ -44,6 +44,16 @@ api:
...
+    description: ""
+    name: ""
+    version: 0
+    minCLIVersion: ""
```
danail-branekov, line 52: I wonder whether it makes sense to default cli versions to something? No idea what default values should be though.

danail-branekov: On a second glance, it seems that we already have some ["min version" code](https://github.com/cloudfoundry/korifi/blob/c353c9ecd6f55a14799d22bedfaf5a822d15d8cc/api/middleware/cli_version.go#L14) in place. I guess returning this constant in the `/v3/info` presenter would make sense, moreover it is already used to [error](https://github.com/cloudfoundry/korifi/blob/c353c9ecd6f55a14799d22bedfaf5a822d15d8cc/api/middleware/cli_version.go#L41) if cli version is tool low.

api/config/config.go
gogolok: In this case I've tried to mimic the previous behaviour of cloud_controller_ng. They are using min_recommended_cli_version for whatever reason:

https://github.com/cloudfoundry/cloud_controller_ng/blob/b970936bc611b97b9e248489baa4ff9bd4ed06e0/app/controllers/v3/info_controller.rb#L12
https://github.com/cloudfoundry/capi-release/blob/0abe5f7855c6004f17b0b61f944b3ae412cac78c/jobs/cloud_controller_ng/templates/cloud_controller_ng.yml.erb#L131

But it might be time to get rid of that confusion :-)

helm/korifi/values.yaml
gogolok: I think the v3/info endpoint in general is very open for interpretation and how you as an operator use and configure it. I think the 'old' approach of cf-deployment to put here the current release number is nice:

https://github.com/cloudfoundry/cf-deployment/commit/ab937830c6f4d66d55ae9bc6d2d270255ab53d5c

For my future usage, I would probably put here not a Korifi specific hash tag alone, but my own interpretation of the current build (some semver for my current Korifi + Kubernetes + cert-manager + ... setup).

helm/korifi/values.yaml
gogolok: Not sure what the hisotry behind the property version is, but it seems to me that cf-deployment is setting this property to the major version number of cf-deployment,
see
https://github.com/cloudfoundry/cf-deployment/blob/ab937830c6f4d66d55ae9bc6d2d270255ab53d5c/cf-deployment.yml#L853

Might be a relict here and really better to auto populate it with https://github.com/cloudfoundry/korifi/blob/c353c9ecd6f55a14799d22bedfaf5a822d15d8cc/version/version.go#L12 and change the type to string.

helm/korifi/values.yaml
gogolok: For cf-deployment, those properties seem to be empty.

Some real-life output:
```
{"build":"v1.56.0","cli_version":{"minimum":"","recommended":""},"custom":{},"description":"","name":"cf-deployment","version":39,"links":{"self":{"href":"https://api.system.EXAMPLE.com/v3/info"},"support":{"href":"support@EXAMPLE.com"}}}
```

I think I would keep minimum and recommended version empty by default and let the operator decide how to use it.

For example if you would fix `cf whoami` ( https://github.com/cloudfoundry/korifi/issues/1724 ), you would probably require a new CF CLI version and the operator could point out the fact you need a specific CLI version.

I consider 
https://github.com/cloudfoundry/korifi/blame/c353c9ecd6f55a14799d22bedfaf5a822d15d8cc/api/middleware/cli_version.go#L14
to be a hard requirement the Korifi release imposes, but the operator might be more strict.

gogolok:
> A general question: do you know use cases where each property of the `/v3/info` endpoint is used? We would refrain from implementing things that noone uses, especially things that are not clear when reading the CF API spec (which is the case with `/v3/info`)


As I said before, I tried to mimic the old behaviour. The only property I see critical is `version`. I guess using a `string` instead of an `integer` is better.

Besides from this, I only see for my scenario the following properties to be useful:
- build: I can put there some bundled version number, but via `custom` property I could also achieve the same; so more of a  cf-deployment version number or helm chart version or whatever I will use in the future to bundle Korifi with other components
- custom: I want to put there some custom labels that make it easier to identify that Korifi installation (project, customer id)
- support: set my own support email/http(s) address
- description: explain the humans this is a Korifi installation
- name: give it a unique name OR the software (helm chart, ...) that has rolled out this Korifi installation
- version: if changed to `string`, I would put here the Korifi version (0.11.0 for example)

helm/korifi/values.yaml
georgethebeatle: As far as I can see the values in cf deployment are already auto populated. I wonder why they needed to expose both `build` and `version` and whether some tools depend on that or if it is just for information purposes as its name implies. If it happens to be the latter I guess we could simply set both `build` and `version` to the same thing as you suggested

api/config/config.go
gogolok: I've removed the `min` prefix for the recommended CLI version

helm/korifi/values.yaml
gogolok: I've changed it.

helm/korifi/values.yaml
gogolok: As far as I can tell, it's just for information purposes, but you never know :-)

I've set `build` and `version` to `version.Version` in the latest commit.

helm/korifi/values.yaml
```diff
@@ -44,6 +44,14 @@ api:
       idle: 900
       readHeader: 10
 
+  infoConfig:
+    description: ""
+    name: ""
```
georgethebeatle, line 49: Maybe we should default `name` and `description` so that it looks more similar to current CF systems. Something generic like:

```
name: "korifi"
description: "Korifi Cloud Foundry Environment"
```

georgethebeatle:
Looks good @gogolok. Dropped a small comment inline. Just one more request - can you squash your commits into one and force push before we approve and merge this PR. This way we will have cleaner history. Thank you for your contribution!

georgethebeatle (approved):

Issue #3237, opened 2024-04-30
Use the ReadyConditionBuilder for all custom resources that have ready state


georgethebeatle:
Those resources are:
* CFPackage
* CFOrg
* CFSpace
* CFProcess
* AppWorkload

Check whether there are more types that are just using strings instead of using the `korifiv1alpha1.StatusConditionReady` constant

PR #3303, opened 2024-05-27
Use the ready condition builder in relevant controllers

danail-branekov:
#3237

## What is this change about?
Korifi custom resources that have their `Ready` condition set by
controllers now utilise the ready condition builder pattern we recently
came up with. The afffected resources are:

* CFPackage
* CFOrg
* CFSpace
* CFProcess
* AppWorkload

While being here,
* take the opportunity to slightly improve their tests
* use the `korifiv1alpha1.StatusConditionReady` constant wherever it
  makes sense

No

## Acceptance Steps
No functional change

@cloudfoundry/wg-cf-on-k8s


davewalter (approved):
Looks good.

tcdowney (approved):

PR #3306, opened 2024-05-28
Fix unbound variable error in bash

gogolok:
No

## What is this change about?

6eb3943b6395d8645a4716eefe7b184e059f419f introduced a bash issue when running `./scripts/deploy-on-kind.sh korifi` more than once (on macOS Sonoma). If you run the script more than once, `TEMP_FILES` variable might be empty. With the option `set -u` an empty `TEMP_FILES` will throw an unbound error.

```bash
bash-3.2$ set -u
bash-3.2$ arr=()
bash-3.2$ echo "${arr[@]}"
bash: arr[@]: unbound variable
bash-3.2$ echo "${arr[@]-}"

bash-3.2$ exit
```

No.

## Acceptance Steps

To reproduce the issue, run the command `./scripts/deploy-on-kind.sh korifi` more than once.
It should exit the script with the following error:

`.../github.com/cloudfoundry/korifi/scripts/install-dependencies.sh: line 108: TEMP_FILES[@]: unbound variable`

None

### DIFF
--- a/scripts/install-dependencies.sh
+++ b/scripts/install-dependencies.sh
@@ -8,7 +8,7 @@ DEP_DIR="$TEST_DIR/dependencies"
 VENDOR_DIR="$TEST_DIR/vendor"
 
 TEMP_FILES=()
-trap 'for file in ${TEMP_FILES[@]}; do rm -rf $file; done' EXIT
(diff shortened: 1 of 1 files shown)


danail-branekov (approved):

danail-branekov:
Thanks @gogolok, looks good

PR #3308, opened 2024-05-30
Add node selector and tolerations options into Korifi helm chart schema

panevpla:
No

## What is this change about?
Adding the `nodeSelector` and `tolerations`options to the Korifi helm chart schema.
This will allow the users to determine the assignment of the `korifi-api` and `korifi-controllers` pods on specific nodes.
Also, used together with `taints` and `tolerations`, certain nodes can be designated for Korifi `api` and `controllers` pods only.

No.

## Acceptance Steps

1. You need to have a Kubernetes cluster with at least 2 nodes.
2. Label one of the nodes with some label of your choice, e.g. `role=korifi-controlplane`;
3. Taint the same node with some taint of your choice, e.g. `role=korifi-controlplane:NoSchedule`;
4. Checkout this branch and install Korifi from it by adding the following settings to the helm install korifi command:
```
        --set=nodeSelector.role=korifi-controlplane \
        --set tolerations[0].key=role \
        --set tolerations[0].operator=Equal \
        --set tolerations[0].value=korifi-controlplane \
        --set tolerations[0].effect=NoSchedule
```
This will allow the `korifi-api` and `korifi-controllers` pods to only be scheduled on the designated node for them. Pods that don't tolerate the taint of the node will not be scheduled on it.

N/A

### DIFF
--- a/helm/korifi/values.yaml
+++ b/helm/korifi/values.yaml
@@ -9,6 +9,8 @@ containerRegistrySecrets:
 eksContainerRegistryRoleARN: ""
 containerRegistryCACertSecret:
 systemImagePullSecrets: []
+tolerations: []
+nodeSelector: {}
 
 reconcilers:
   build: kpack-image-builder
--- a/helm/korifi/values.yaml
+++ b/helm/korifi/values.yaml
@@ -9,8 +9,6 @@ containerRegistrySecrets:
 eksContainerRegistryRoleARN: ""
 containerRegistryCACertSecret:
 systemImagePullSecrets: []
-tolerations: []
-nodeSelector: {}
 
 reconcilers:
   build: kpack-image-builder
@@ -25,7 +23,9 @@ api:
   include: true
 
   image: cloudfoundry/korifi-api:latest
-
+  
+  nodeSelector: {}
+  tolerations: []
   replicas: 1
   resources:
     requests:
@@ -66,7 +66,9 @@ api:
 
 controllers:
   image: cloudfoundry/korifi-controllers:latest
-
+  
+  nodeSelector: {}
+  tolerations: []
   replicas: 1
   resources:
     limits:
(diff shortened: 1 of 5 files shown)


helm/korifi/values.yaml
```diff
@@ -9,6 +9,8 @@ containerRegistrySecrets:
 eksContainerRegistryRoleARN: ""
 containerRegistryCACertSecret:
 systemImagePullSecrets: []
+tolerations: []
```
danail-branekov, line 12: The new values configure the cluster itself, I wonder whether it would make sense fo group them under a `k8s` section. I.e. instead of using `--set nodeSelector.role=something`, use `--set k8s.nodeSelector.role=something`

Korifi itself does not have the concept of tolerations and node selectors, so having those in the values "root" seems a bit odd to me.

danail-branekov:
Hey @panevpla, thanks for the PR. Besides the comment I left above, you would also have to sign the CLA.

I also wonder would it make sense to have separate node selector and tolerations per `api` and `controllers` so that the api deployment could run on one node and the `controllers` one on another. Or would that be an overkill?

panevpla:
> Hey @panevpla, thanks for the PR. Besides the comment I left above, you would also have to sign the CLA.
> 
> I also wonder would it make sense to have separate node selector and tolerations per `api` and `controllers` so that the api deployment could run on one node and the `controllers` one on another. Or would that be an overkill?

That seems also reasonable. If we accept this approach, I guess we don't need the `k8s` group, right?
I think I have signed the CLA, at least I received a confirmation email about it.

danail-branekov:
> I guess we don't need the k8s group

My instinct first told me that we need `k8s`groups, one under `api` and one under `controllers`. But then in `api` and `controllers` groups we already have k8s-specific values, `replicas` and `resources`. So puting the new k8s values directly under `api` and `controllers` would at least be consistent with them.

panevpla:
@danail-branekov as discussed, moved `nodeSelector` and `tolerations` under `api` and `controllers` groups.

danail-branekov (approved):

Issue #3262, opened 2024-05-22
As a Korifi operator I want to be able to enable experimental managed services support


danail-branekov:
**GIVEN** I use the default helm values
**WHEN** I install the Korifi helm chart
**THEN** I can see the `experimentalManagedServicesEnabled` set to `false` in both api and controllers config maps

**GIVEN** I have specified the `experimental.managed-services-enabled=true` in my helm values
**WHEN** I install the Korifi helm chart
**THEN** I can see the `experimentalManagedServicesEnabled` set to `true` in both api and controllers config maps

PR #3314, opened 2024-05-31
Introduce `experimental.managedServices.include` helm value

danail-branekov:
#3262

## What is this change about?
The upcoming managed services support would be initially disabled by
setting the helm value to `false` (the default)

Operators should deliberately enable the flag should they want to give
the feature a try.

This change simply introduces the flag in the helm chart, there is no
implementation that uses it yet. The `deploy-on-kind` script sets it to
`true` as it is meant to be run for development purposes.

The kind installer sets it to `true` as well - whoever is installing
Korifi on kind probably just wants to play with it, therefore enabling
the experimental support does make sense.

No

## Acceptance Steps
See story

@cloudfoundry/wg-cf-on-k8s

### DIFF
--- a/helm/korifi/values.schema.json
+++ b/helm/korifi/values.schema.json
@@ -224,7 +224,14 @@
               "type": "string"
             }
           },
-          "required": ["description", "name", "minCLIVersion", "recommendedCLIVersion", "custom", "supportAddress"]
+          "required": [
+            "description",
+            "name",
+            "minCLIVersion",
+            "recommendedCLIVersion",
+            "custom",
+            "supportAddress"
+          ]
         },
         "lifecycle": {
           "type": "object",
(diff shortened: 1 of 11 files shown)


helm/korifi/values.schema.json
```diff
@@ -579,6 +586,21 @@
...
+          "properties": {
+            "include": {
+              "description": "Enable managed services support",
+              "type": "boolean"
+            }
+          },
+          "type": "object"
+        }
+      },
+      "description": "Experimental features. Make sure you do not enable those on production. No guarantee provided! Backwards incompatible changes in future are quite probable!",
```
tcdowney, line 602:
```suggestion
      "description": "Experimental features. No guarantees are provided and breaking/backwards incompatible changes should be expected. These features are not recommended for use in production environments.",
```

Suggesting a minor wording tweak. Otherwise it all LGTM. 👍

tcdowney (approved):

helm/korifi/values.schema.json
danail-branekov: Ugh, I have merged that without noticing your comment. I am goind to put that in the next PR

PR #3351, opened 2024-06-20
add API endpoints to Terminate a instances (cf restart-app-instance)

marsteg:
https://github.com/cloudfoundry/korifi/issues/3350

## What is this change about?
Adding cf restart-app-instance API endpoints

Not that I know of.

## Acceptance Steps
in the cf cli, the command: 
cf restart-app-instance test-app 3 --process web -v
works


api/handlers/app.go
```diff
@@ -657,6 +668,51 @@ func (h *App) getAppFeature(r *http.Request) (*routing.Response, error) {
...
+	process, hasProcessType := findProcessType(appProcesses, processType)
+	if !hasProcessType {
+		return nil, apierrors.LogAndReturn(logger,
+			apierrors.NewNotFoundError(nil, repositories.ProcessResourceType),
+			"app does not have required process type",
+		)
+	}
+	instance, err := strconv.Atoi(instanceID)
+	if err != nil {
+		return nil, apierrors.LogAndReturn(logger, err, "Instance is not a string", "InstanceID", instanceID)
```
georgethebeatle, line 700: Maybe you meant to say "Instance is not an integer"?

go.mod
```diff
@@ -51,6 +51,7 @@ require (
 	github.com/containerd/log v0.1.0 // indirect
 	github.com/dgryski/go-rendezvous v0.0.0-20200823014737-9f7001d12a5f // indirect
 	github.com/distribution/reference v0.6.0 // indirect
+	github.com/evanphx/json-patch v5.9.0+incompatible // indirect
```
georgethebeatle, line 54: Is this really needed? It is an indirect dependency so no under our direct control, but it looks weird

api/handlers/app_test.go
```diff
@@ -1772,6 +1774,113 @@ var _ = Describe("App", func() {
 			})
 		})
 	})
+
+	Describe("DELETE /v3/apps/:guid/processes/:process/instances/:instance", func() {
+		When("the app has a single instance in the default (web) process", func() {
```
georgethebeatle, line 1779: This `When` could be inlined in the `Describe`, since it is a dafault case

api/handlers/app_test.go
```diff
@@ -1772,6 +1774,113 @@ var _ = Describe("App", func() {
...
+				processRepo.ListProcessesReturns([]repositories.ProcessRecord{
+					{
+						GUID:             "process-1-guid",
+						SpaceGUID:        spaceGUID,
+						AppGUID:          appGUID,
+						Type:             "web",
+						DesiredInstances: 1,
+					},
+				}, nil)
+				req = createHttpRequest("DELETE", "/v3/apps/"+appGUID+"/processes/web/instances/0", nil)
```
georgethebeatle, line 1790: It is a good practice to have the test call separated in a `JustBeforeEach`. This `JustBeforeEach` can be defined at the top level potentially parametrizing the number of the instance to be deleted. This number can be set in each `When`'s respective `BeforeEach`

api/handlers/app_test.go
```diff
@@ -1772,6 +1774,113 @@ var _ = Describe("App", func() {
...
+			BeforeEach(func() {
+				req = createHttpRequest("DELETE", "/v3/apps/boom/processes/web/instances/0", nil)
+				appRepo.GetAppReturns(repositories.AppRecord{}, apierrors.NewForbiddenError(nil, repositories.AppResourceType))
+			})
+
+			It("returns an error", func() {
+				expectNotFoundError("App")
+			})
+		})
+		When("the app exists but the process does not", func() {
```
georgethebeatle, line 1829: The app exists by default so it would be stylistically better to omit this from the `When`. We already have a `When` for the case where the app does not exists. So this `When` can be rephrazed as `When("the process does not exists")`

api/handlers/app_test.go
```diff
@@ -1772,6 +1774,113 @@ var _ = Describe("App", func() {
 			})
 		})
 	})
+
+	Describe("DELETE /v3/apps/:guid/processes/:process/instances/:instance", func() {
```
georgethebeatle, line 1778: Maybe a better structure for this describe would be 

```
Describe("DELETE /v3/apps/:guid/processes/:process/instances/:instance", func() {
  It("returns process not found error", func() {})

  When("the app does not exist", func() {
    It("returns an app not found error", func() {})
  })

  When("the app has a process", func() {
    It("returns an instance not found error", func() {})

    When("the process has an instance") {
      It("restarts the instance", func() {})
    })

    When("the app has more than one process", func() {
      It("restarts the instance", func() {})
    })
  })
})
```

The idea is that each when builds on top of its parent. This way a lot of repetition can be avoided

api/handlers/process_test.go
```diff
@@ -454,4 +460,79 @@ var _ = Describe("Process", func() {
 			})
 		})
 	})
+
+	Describe("DELETE /v3/processes/:guid/instances/:instance", func() {
```
georgethebeatle, line 464: We could restructure this `Describe` in the spirit of the apps handler one

- Start with no instance an no process
- When there is an instance
  - When there is a process

api/handlers/process_test.go
```diff
@@ -454,4 +460,79 @@ var _ = Describe("Process", func() {
...
+
+				appRepo.GetAppReturns(repositories.AppRecord{
+					GUID:      "app-guid",
+					SpaceGUID: "space-guid",
+					Revision:  "revision",
+				}, nil)
+
+			})
+
+			JustBeforeEach(func() {
```
georgethebeatle, line 483: We could pull the `JustBeforeEach` up and parametrize the instance number to avoid repeating essentially the same `JustBeforeEach` in every `When`

api/repositories/pod_repository_test.go
```diff
@@ -0,0 +1,84 @@
...
+						Image: "nginx",
+					},
+				},
+			},
+		}
+		Expect(k8sClient.Create(ctx, pod)).To(Succeed())
+
+	})
+	Describe("DeletePod", func() {
+		When("the user is not a a SpaceDeveloper", func() {
```
georgethebeatle, line 57: This `When` can be nested in the describe as it could be looked at as the default case

api/handlers/process.go
```diff
@@ -45,19 +48,25 @@ type Process struct {
...
 	requestValidator RequestValidator,
+	appRepo CFAppRepository,
+	podRepo PodRepository,
 ) *Process {
 	return &Process{
 		serverURL:        serverURL,
 		processRepo:      processRepo,
 		processStats:     processStatsFetcher,
 		requestValidator: requestValidator,
+		appRepo:          appRepo,
```
georgethebeatle, line 68: Injecting an app repo to the process handler looks weird to me. We only need the app record so that we know the app revision, so maybe we could extend the process repo by adding a `GetAppRevisionForProcess` method. It would not duplicate a lot of the app repo code since it is simply a k8sClient.Get call. On the other hand the app repo contains a lot of methods that the process repo should not be able to see.

georgethebeatle:
Hey @marsteg thanks for your PR, I have left some inline comments. In technical terms your PR looks good and the comments address some stylistic improvements. 

In more abstract terms adding more behaviour to the PodRepo is something we are trying to avoid, since it is breaking Korifi's abstraction and assuming that apps are rendered into pods in a very specific way that is implemented by the default statefulset runner. This will make it hard to potentially swap the current implementation with other ones - like a Deployment runner or a multi cluster runner.

Having said that we have not yet thought of a way to fix this abstraction leak, so I cannot propose a better approach for the time being. 

Can you elaborate on what you need that feature for and what is the priority of the use case?

api/handlers/app.go
marsteg: I wasn't sure how to phrase the problem -> i'll take your proposal :)

PR #3383, opened 2024-07-15
Use `go-functional` in repositories `list` methods

danail-branekov:
No

## What is this change about?
This eliminates quite some boilerplate code around filtering and mapping
k8s resources. `go-functional` also replaces our inhouse filtering
implementation.

Also, this change removes code duplications in methods such as
`GetSomethingByAandB` that list and filter resources. Instead of
implementing the list operation themselves, they delegate to the
relevant `list` method by constructing a list messages that matches the
criteria they need. As a further improvement we could get rid of those
methods and just invoke `repo.List` from the handlers.
As a consequence of this change those methods now return a `NotFound`
error rather than `Unauthorised` when the user is not allowed to list
the resources. Still, that should be OK

No

@georgethebeatle


BooleanCat (commented):
Using go-functional v2, see `it.Lines` - you can iterate over lines from a reader.

api/repositories/pod_repository.go
```diff
@@ -55,62 +42,71 @@ type RuntimeLogsMessage struct {
...
+	bufReader := bufio.NewReader(r)
+	for {
+		line, err = bufReader.ReadBytes('\n')
+		lines = append(lines, string(line))
+		if err != nil {
+			if err != io.EOF {
+				logger.Info("failed to parse pod logs", "err", err)
+			}
+			break
+		}
```
BooleanCat, line 103: Using go-functional v2, see `it.Lines` - you can iterate over lines from a reader.

BooleanCat:
```go
lines, errs := itx.Lines(r).Unzip()
if len(errs.Collect()) > 0 {
  panic("Ahhh")
}
lines.Collect()
```

BooleanCat: Or if you know reading cannot fail such as with reading from a string buffer:

```go
lines := itx.LinesString(r).Left().Collect()
```

api/repositories/pod_repository.go
```diff
@@ -55,62 +42,71 @@ type RuntimeLogsMessage struct {
...
-				}
-			}
+	return iter.Chain(logRecords...).Collect(), nil
+}
+
+func (r *PodRepo) getPodLogs(ctx context.Context, k8sClient k8sclient.Interface, logger logr.Logger, pod corev1.Pod, limit int64) iter.Iterator[LogRecord] {
+	logReadCloser, err := k8sClient.CoreV1().Pods(pod.Namespace).GetLogs(pod.Name, &corev1.PodLogOptions{Timestamps: true, TailLines: &limit}).Stream(ctx)
+	if err != nil {
+		logger.Info("failed to fetch logs", "pod", pod.Name, "reason", err)
+		return iter.Lift([]LogRecord{})
```
BooleanCat, line 80: For this there's the more explicit `iter.Exhausted[LogRecord]()`

api/repositories/pod_repository.go
danail-branekov: We wanted to get as much logs as we can regardless read errors. go-functional v1 LinesString did not look it supported that (i.e. it just failed) so we kept that code. When switching to v2 we would make use of `itx.LinesString` for sure

api/repositories/pod_repository.go
danail-branekov: ah, nice! Will use that :)

georgethebeatle (approved):

Issue #3292, opened 2024-05-23
As a Korifi API user I want to be able to delete managed service instances


danail-branekov:
### CF API

**GIVEN** I have enabled the experimental managed services flag in the helm values
**GIVEN** I have created a managed service instance
**WHEN** I `DELETE /v3/service_instances/service-instance-guid`
**THEN** I get a `202 Accepted` response with `Location` header pointing me to the delete operation
**THEN** eventually I see the delete operation succeeds
**THEN** I see the service instance is deleted from the service broker

### CF CLI

**GIVEN** I have **GIVEN** I have enabled the experimental managed services flag in the helm values
**GIVEN** I have registered a service broker
**GIVEN** I have created a managed service instance by running
```
cf create-service <offering> <plan> my-instance
```
**WHEN** I delete the service instance by running
```
cf delete-service my-instance
```
**THEN** the above command succeeds

Dev notes:
* The implementation should deprovision the service with the broker over OSBAPI
* When requesting service deprovisioning from the broker, the implementation should send the `accepts_incomplete=true` query parameter to enable async deprovisioning
* When the broker deprovisions the service in a sync manner (the broker responds with `200 OK`), the implementation must not check for service instance last operation as the instance has been already deprovisioned and no operation is running
* When the broker responds with `202 Accepted`, then the implementation should poll the service instance last operation until the operation completes.
* If the operation succeeds, the `CFServiceInstance` should be deleted
* If the operation fails, the failure should be reflected in the `CFServiceInstance` status

* [Deprovision service OSBAPI docs](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#deprovisioning)
* [Service instance last operation OSBAPI docs](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#polling-last-operation-for-service-instances)
* [Synchronous vs asynchronous operations OSBAPI docs](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#synchronous-and-asynchronous-operations)
* [Managed services proposal](https://docs.google.com/document/d/1d1GSdDVzIMZ9kLedOPT7ZyCqGUwzeoOvHePcrBlYAto/edit?usp=sharing)
* [Delete service instance docs](https://v3-apidocs.cloudfoundry.org/version/3.165.0/index.html#delete-a-service-instance)
* local broker registration
```
cf push sample-broker -p ~/workspace/korifi/tests/assets/sample-broker
register-local-broker sample-broker
[...]

PR #3484, opened 2024-09-19
Issues/delete service instances

zabanov-lab:
#3292 

## What is this change about?
Implenting functionality about managed service instacnes

No

### DIFF
--- a/api/handlers/service_instance_test.go
+++ b/api/handlers/service_instance_test.go
@@ -42,6 +42,7 @@ var _ = Describe("ServiceInstance", func() {
 		serviceInstanceRepo.GetServiceInstanceReturns(repositories.ServiceInstanceRecord{
 			GUID:      "service-instance-guid",
 			SpaceGUID: "space-guid",
+			Type:      korifiv1alpha1.UserProvidedType,
 		}, nil)
 
 		spaceRepo = new(fake.CFSpaceRepository)
@@ -564,6 +565,29 @@ var _ = Describe("ServiceInstance", func() {
 			Expect(message.SpaceGUID).To(Equal("space-guid"))
 
 			Expect(rr).To(HaveHTTPStatus(http.StatusNoContent))
+		})
+
+		When("the service instance is managed", func() {
+			BeforeEach(func() {
+				serviceInstanceRepo.GetServiceInstanceReturns(repositories.ServiceInstanceRecord{
+					GUID:      "service-instance-guid",
+					SpaceGUID: "space-guid",
+					Type:      korifiv1alpha1.ManagedType,
+				}, nil)
+			})
+
+			It("deletes the service instance", func() {
+				Expect(serviceInstanceRepo.DeleteServiceInstanceCallCount()).To(Equal(1))
+				_, actualAuthInfo, message := serviceInstanceRepo.DeleteServiceInstanceArgsForCall(0)
+				Expect(actualAuthInfo).To(Equal(authInfo))
+				Expect(message.GUID).To(Equal("service-instance-guid"))
+				Expect(message.SpaceGUID).To(Equal("space-guid"))
+
+				Expect(rr).To(SatisfyAll(
+					HaveHTTPStatus(http.StatusAccepted),
+					HaveHTTPHeaderWithValue("Location", ContainSubstring("/v3/jobs/managed_service_instance.delete~service-instance-guid")),
+				))
+			})
 		})
 
 		When("getting the service instance fails with not found", func() {
--- a/controllers/controllers/services/instances/managed/controller.go
+++ b/controllers/controllers/services/instances/managed/controller.go
@@ -186,7 +186,7 @@ func (r *Reconciler) provisionServiceInstance(
 		return ctrl.Result{}, err
 	}
 
-	var provisionResponse osbapi.ProvisionServiceInstanceResponse
+	var provisionResponse osbapi.ServiceInstanceOperationResponse
 	provisionResponse, err = osbapiClient.Provision(ctx, osbapi.InstanceProvisionPayload{
 		InstanceID: serviceInstance.Name,
 		InstanceProvisionRequest: osbapi.InstanceProvisionRequest{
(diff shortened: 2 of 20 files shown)


api/handlers/service_instance_test.go
```diff
@@ -566,6 +567,29 @@ var _ = Describe("ServiceInstance", func() {
...
+		When("the service instance is managed", func() {
+			BeforeEach(func() {
+				serviceInstanceRepo.GetServiceInstanceReturns(repositories.ServiceInstanceRecord{
+					GUID:      "service-instance-guid",
+					SpaceGUID: "space-guid",
+					Type:      korifiv1alpha1.ManagedType,
+				}, nil)
+			})
+
+			It("deletes the service instance", func() {
```
danail-branekov, line 579: Maybe change the description of the `It` to `responds with deletion job location`

controllers/controllers/services/instances/managed/controller.go
```diff
@@ -129,6 +117,18 @@ func (r *Reconciler) ReconcileResource(ctx context.Context, serviceInstance *kor
 		return ctrl.Result{}, fmt.Errorf("failed to create client for broker %q: %w", serviceBroker.Name, err)
 	}
 
+	if !serviceInstance.GetDeletionTimestamp().IsZero() {
```
danail-branekov, line 120: I think we should keep the deleted/ready/failed check right after setting the observed generation to the status.

Provided that the instance is being deleted, or has reached its terminal state, we should not bother getting the plan at all.

danail-branekov (changes requested):
Overall looks good, I left some minor remarks.

Also, could you squash the commits in order not to have the wip one and amend the commit description with the issue number, e.g.

```
fixes #3292
```

controllers/controllers/services/instances/managed/controller.go
zabanov-lab: Since, we implemented deprovisioning functionality the finalizeCFServiceInstance method needs the service plan, offering and etc. because we call the Deprovision method in it and we call the finalizeCFServiceInstance method only in the deleted/ready/failed check

danail-branekov: Oh, I see. Makes sense

danail-branekov (approved):

controllers/controllers/services/instances/managed/controller.go
```diff
@@ -129,6 +117,18 @@ func (r *Reconciler) ReconcileResource(ctx context.Context, serviceInstance *kor
...
+		return r.finalizeCFServiceInstance(ctx, osbapiClient, serviceInstance, serviceOffering, servicePlan)
```
danail-branekov, line 121: I think that we still need to move the finalization bit before getting getting plan/offering/etc

Our CI clusters could not be deleted as there are a couple of CFServiceInstances with finalizers that cannot be deleted. They have their deletion timestamp set but finalization fails with

```
    message: 'failed to get service plan "1ac1edee-804e-51bd-a4c0-149e8271f275": CFServicePlan.korifi.cloudfoundry.org "1ac1edee-804e-51bd-a4c0-149e8271f275" not found'
```

Indeed, there are no service plans on the cluster.

I do not know how the cluster reached this situation, but in any case, fainalizing resources should be "best effort" - they should try to delete whatever possible, but if an error occurs, still allow deletion. Having said that, I think we should move finalization at the top of the `Reconcile` method, just after setting the observed generation and have plan and offering being looked up in the finalize method. If an error occurrs while fetching the plan/offering occurs, it should be only logged AND the finalizer should be removed

controllers/controllers/services/instances/managed/controller.go
```diff
@@ -228,15 +228,103 @@ func getServiceInstanceParameters(serviceInstance *korifiv1alpha1.CFServiceInsta
...
+func (r *Reconciler) finalizeCFServiceInstance(
+	ctx context.Context,
+	serviceInstance *korifiv1alpha1.CFServiceInstance,
+) (ctrl.Result, error) {
 	log := logr.FromContextOrDiscard(ctx).WithName("finalizeCFServiceInstance")
 
-	if !controllerutil.ContainsFinalizer(cfServiceInstance, korifiv1alpha1.CFManagedServiceInstanceFinalizerName) {
+	if !controllerutil.ContainsFinalizer(serviceInstance, korifiv1alpha1.CFManagedServiceInstanceFinalizerName) {
 		return ctrl.Result{}, nil
 	}
```
danail-branekov, line 239:
```
if !containsFinalizer() {
  return ctrl.Result{}, nil
}

if deprovisionRequested() {
  if removeFinalizer(){
    log.finalizerRemoved()
  }
  return ctrl.Result{}, nil
}

err := deprovisionInstance() // only returns error if osbapiClient.Deprovision fails; also, do NOT wait for the deprovision operation to complete
iferr {
  return ctrl.Result{}, err
}

return ctrl.Result{Requeue: true}, nil
```

controllers/controllers/services/instances/managed/controller.go
```diff
@@ -129,6 +121,14 @@ func (r *Reconciler) ReconcileResource(ctx context.Context, serviceInstance *kor
 		return ctrl.Result{}, fmt.Errorf("failed to create client for broker %q: %w", serviceBroker.Name, err)
 	}
 
+	if isReady(serviceInstance) {
```
danail-branekov, line 124: Nitpick: I would move the ready/failed checks before getting the plan and creating the osbapi client - in case we have reached the terminal state, there is nothing to do, therefore no need to create the osbapi client

controllers/controllers/services/instances/managed/controller_test.go
```diff
@@ -499,16 +499,75 @@ var _ = Describe("CFServiceInstance", func() {
 	})
 
 	When("the instance is deleted", func() {
+		BeforeEach(func() {
+			brokerClient.DeprovisionReturns(osbapi.ServiceInstanceOperationResponse{
+				Operation: "deprovision-op",
+			}, nil)
+			brokerClient.GetServiceInstanceLastOperationReturns(osbapi.LastOperationResponse{
```
danail-branekov, line 506: stubbing the last operation should not be necessary as the fainalizer does not check it anymore

controllers/controllers/services/instances/managed/controller_test.go
```diff
@@ -499,16 +499,75 @@ var _ = Describe("CFServiceInstance", func() {
...
+					g.Expect(instance.Status.Conditions).To(ContainElement(SatisfyAll(
+						HasType(Equal(korifiv1alpha1.StatusConditionReady)),
+						HasStatus(Equal(metav1.ConditionFalse)),
+						HasReason(Equal("DeprovisionFailed")),
+					)))
+				}).Should(Succeed())
+			})
+		})
+
+		When("deprovision operation completes", func() {
```
danail-branekov, line 557: As the finalizer does not check last operation, this `When` does not seem to make sense

danail-branekov (changes requested):
I have left some minor comments, looks all right otherwise

controllers/controllers/services/instances/managed/controller_test.go
```diff
@@ -499,16 +499,75 @@ var _ = Describe("CFServiceInstance", func() {
...
+					ID: instance.Name,
+					InstanceDeprovisionRequest: osbapi.InstanceDeprovisionRequest{
+						ServiceId: "service-offering-id",
+						PlanID:    "service-plan-id",
+					},
+				}))
+			}).Should(Succeed())
+		})
+
+		It("does not delete the service instance", func() {
```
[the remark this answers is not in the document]
zabanov-lab: This test failed only once when I ran the tests intil they fail, which is strange for me too..

danail-branekov (approved):

PR #3485, opened 2024-09-19
Support multi-arch builds for local development

pbusko:

no

## What is this change about?

hardcoded `GOARCH` prevents local testing with `kind` when using ARM64 machine

no

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,9 +82,9 @@ test-smoke: build-dorifi bin/cf
 
 
 build-dorifi:
-	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -C tests/assets/dorifi-golang -o ../dorifi/dorifi .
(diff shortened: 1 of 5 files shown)


danail-branekov:
@pbusko Just out of curiosity - have you tried running `deploy-on-kind` on a non-amd64 architecture?

danail-branekov (approved):

pbusko:
> @pbusko Just out of curiosity - have you tried running `deploy-on-kind` on a non-amd64 architecture?

It does work with noticeable performance degradation for emulated components, like Korifi API/Controllers and `kpack`. We're also working on bringing native arm64 support for `kpack`.
Some colleagues even reported issue when running `./scripts/deploy-on-kind.sh` on certain VMs, with `fork/exec` errors highlighting architecture mismatch.

georgethebeatle:
Thanks @pbusko

PR #3511, opened 2024-10-10
Add section about stack handling differences to known differences docs

FloThinksPi:
Add section about stack handling to known-differences-with-cf-for-vms.md

Feel free to correct me about that but to my understanding the current restage/staging behaviour in korifi comes with the difference to CF on VMs below, just to get that written down somewhere as @beyhan pointed me to this doc.

It might make sense to rephrase that by someone with deeper technical knowledge in the Korifi implementation and fact check it.

No

## What is this change about?
Documentation

No

## Acceptance Steps
Crosscheck documentation

None

### DIFF
--- a/docs/known-differences-with-cf-for-vms.md
+++ b/docs/known-differences-with-cf-for-vms.md
@@ -19,6 +19,11 @@ the API shim, the user would only be able to list and view spaces which have rol
 ### Rolling Updates
 In Kofiri `--strategy=rolling` is implemented using k8S rolling update capabilities of the scheduler. At the moment korifi uses statefulsets to run the app workloads. Rolling update for statefulsets stops the old instance before starting the new one, for ordering reasons. If the app has only one instance the udpate will cause a downtime. Apps with more than one instance won't experience any downtime, but they will have one instance less up and running during the update.
 
+### Stack Changes
+While in CF for VMs the staging process yields a droplet, which is a stripped container image without base layer/operating system.
+In Korifi a fully fledged image is created which includes the base operating system(stack). 
+This alters the behaviour in case of stack updates. CF for VMs does purposefully combine stack and droplet at container instance creation time - thus a restart of an app consumes a new stack. In Korifi only a restage of an app consumes thew new stack. One might need to detect that and dynamically rebase a droplet onto a new stack in korifi before starting the container to yield the same behaviour here. 
+
 ## Apps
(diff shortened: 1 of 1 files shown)


danail-branekov:
Without looking at the code and having experimented with that, here are my two cents.

As far as I know, whenever the stack on the kpack cluster builder changes, (**I think**) kpack schedules a new kpack build for each app. Once the app image is pushed, it would be set on the CFApp resource as a current droplet. Thus, whenever the app restarts, it would pick up the new version.
TL;DR - we need to experiment to see whether a restage is actually required, or a restart is enough.

The assumptions above is of course true when the default kpack bulder is used as a build runner. A custom build runner could behave differently.

danail-branekov:
Hey @FloThinksPi 

I ran some experiments myself, and TL;DR what you suggest is completely correct.

FWIW, here is my experiment:

1. I have deployed Korifi on kind and pushed an app. Its workload referenced image sha `...d1e8b76e5`
2. I changed the cluster stack images:
    * `paketobuildpacks/build-jammy-base` -> `paketobuildpacks/build-jammy-full`
    * `paketobuildpacks/run-jammy-base` -> `paketobuildpacks/run-jammy-full`
3. Changing the stack immediately triggerred a new kpack build and that build produced an app image with sha `...4108c38a5`. However, the app workload image **DID NOT** change
4. I restarted the app, the workload image **DID NOT** change again
5. I restaged the app, and that picked up the new image (sha ...4108c38a5`)

danail-branekov (approved):

FloThinksPi:
Thanks for testing it out @danail-branekov !

Yea i also thought about possible solutions but there is no easy one that comes to my mind straight away.
The stack seperation in CF is a unique feature that is hard to replicate in K8s and yield specific benefits(cost/resource usage) and behaviour one would lose(gurantees towards app developers). Im not aware if one can combine images on K8s at startup.

Another option is a restart waits until a implicit [cnb rebase](https://buildpacks.io/docs/for-app-developers/concepts/rebase/) is finished if a new stack shall be used and "hangs" for some time - however since the app restart is a synchronous api call that just is allowed to return when the restart was propagated correctly to the container runtime - its hard to add this in a compatible behaving way as also more errors could occur in the rebase process that need different api responses likely. However i think a transparent rebase on a restart may be the most viable option here.

Issue #3293, opened 2024-05-23
As a Korifi API user I want to be able to bind to a managed service instance (async bind)


danail-branekov:
**GIVEN** I have enabled the experimental managed services flag in the helm values
**GIVEN** I have pushed an app
**GIVEN** I have created a managed service instance
**WHEN** I `POST /v3/service_credential_bindings` with the following request body

```
{
  "relationships": {
    "app": {
      "data": {
        "guid": "my-app-guid"
      }
    },
    "service_instance": {
      "data": {
        "guid": "my-managed-service-instance-guid"
      }
    }
  },
  "type": "app"
}
```
**WHEN** the OSBAPI broker responds with `202 Accepted` (i.e. the bind operation is asynchronous)
**THEN** I get a `202 Accepted` response with a `Location` header pointing me to a bind job
**THEN** eventually I see the bind job succeeding
**THEN** I see a credential binding created with the service broker

Dev notes:
* The implementation must invoke `Bind` on the broker over OSBAPI
* When requesting `Bind` from the broker, the implementation should send the `accepts_incomplete=true` query parameter to enable async provisioning
* This story implements the asynchronous bind, i.e. when the broker responds with `202 Accepted`
* The implementation should poll the binding last operation until the operation completes. Once the operation completes, its status must be reflected in the `CFServiceBinding`
* The binding response `credentials` object should be persisted in a secret in the same namespace as the `CFServiceBinding`.
* [Bind OSBAPI docs](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#binding)
* [Service binding last operation OSBAPI docs](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#polling-last-operation-for-service-bindings)
* [Synchronous vs asynchronous operations OSBAPI docs](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#synchronous-and-asynchronous-operations)
* [Managed services proposal](https://docs.google.com/document/d/1d1GSdDVzIMZ9kLedOPT7ZyCqGUwzeoOvHePcrBlYAto/edit?usp=sharing)
* [Create service binding docs](https://v3-apidocs.cloudfoundry.org/version/3.166.0/index.html#create-a-service-credential-binding)

PR #3556, opened 2024-10-24
Issues/3293 bind async

zabanov-lab:
#3293
## What is this change about?
Implement asynchronous binding
No


danail-branekov (dismissed):

helm/korifi/controllers/crds/korifi.cloudfoundry.org_appworkloads.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.16.4
+    controller-gen.kubebuilder.io/version: v0.16.3
```
gogolok, line 6: So your local ./controllers/bin/controller-gen is not using 0.16.4?

danail-branekov: Nice catch, @gogolok - I totally overlooked that this is not a bump, but de-bump. Will fix that

danail-branekov (approved):

PR #3561, opened 2024-10-25
Improve compatibility with the `code-generator` project

pbusko:

no

## What is this change about?

This PR introduces minor changes to the `controllers/api` package, to make it compatible with the https://github.com/kubernetes/code-generator project. This project can be used as an alternative for the `controller-gen`'s deepcopy generator. Also it brings missing features like generation of typed clients and `applyconfiguration`'s for proper SSA support.

no

### DIFF
--- a/controllers/api/v1alpha1/appworkload_types.go
+++ b/controllers/api/v1alpha1/appworkload_types.go
@@ -69,6 +69,8 @@ type AppWorkloadStatus struct {
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
 
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
 // AppWorkload is the Schema for the appworkloads API
 type AppWorkload struct {
 	metav1.TypeMeta   `json:",inline"`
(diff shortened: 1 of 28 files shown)


danail-branekov (approved):

controllers/api/v1alpha1/appworkload_types.go
```diff
@@ -69,6 +69,8 @@ type AppWorkloadStatus struct {
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
 
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
```
danail-branekov, line 72: Adding this annotation seem to remove [the appworload status](https://github.com/cloudfoundry/korifi/blob/724bf33dce620ee1eb3ded5882aef4bbaaae571c/helm/korifi/controllers/crds/korifi.cloudfoundry.org_appworkloads.yaml#L806-L807) from the generated manifest and this is causing controllers test to [fail](https://github.com/cloudfoundry/korifi/actions/runs/11572989343/job/32214943264) as the app workload resource has no `Status` anymore, while the test wants to update it.

In order to reproduce locally, run `make generate manifests` and then run the process controller tests: `scripts/test ./controllers/controllers/workloads/processes`

I cannot explain why this occurs

danail-branekov (dismissed):
The app workload getting its status removed by the generate/manifests make target needs to be addressed

pbusko:
@danail-branekov could you please trigger the e2e tests?

georgethebeatle (approved):

Issue #3546, opened 2024-10-23
Support floating point number values for disk_quota in application manifest


georgethebeatle:
### Backgroundo

It turns out it is possible to push a manifest like this one on CF for VMs

```
---
applications:
- name: test-sapmachine
  memory: 800M
  path: ./target/application.war
  instances: 1
  disk_quota: 1.5G
```

As of today Korifi assumes `disk_quota` to be an integer

### Acceptance

**GIVEN** I have targeted a space on a Korifi deployment
**WHEN** I push a manifest with a decimal amount of `disk_quota`
**THEN** the application starts

PR #3567, opened 2024-10-28
Support floating point number values for disk_quota and memory

gogolok:
#3546

## What is this change about?
disk_quota can be a floating point number under CF for VMs.

I've looked up cloudcontroller_ng and found
https://github.com/cloudfoundry/cloud_controller_ng/blob/9a1dcf1faf466aa286d4372623b6e75ad7d2ce1b/lib/cloud_controller/app_manifest/byte_converter.rb#L11
and tried to apply it to the current validation we have in Korifi.

I checked whether I could add some positive test cases (that 1.5G gets successfully accepted), but could not find any existing test cases so far and how it could be done with the current code setup.

Nevertheless, I've checked locally with kind to push an app with 1.5 GB disk_quota and 0.5G memory and it was successfully translated to the corresponding resources.

```diff
--- a/manifest.yml
+++ b/manifest.yml
@@ -1,6 +1,7 @@
 ---
 applications:
   - name: test-app
-    memory: 256M
+    memory: 0.5G
     instances: 1
     random-route: true
+    disk_quota: 1.5G
```

```json
                "resources": {
                    "limits": {
                        "ephemeral-storage": "1536Mi",
                        "memory": "512Mi"
                    },
                    "requests": {
                        "cpu": "50m",
                        "ephemeral-storage": "1536Mi",
                        "memory": "512Mi"
                    }
                }
```

No.

## Acceptance Steps
Try to push apps with disk_quota and memory set to floating point numbers.

### DIFF
--- a/api/payloads/manifest.go
+++ b/api/payloads/manifest.go
@@ -256,7 +256,7 @@ func (s ManifestApplicationService) Validate() error {
 	return validation.ValidateStruct(&s, validation.Field(&s.Name, validation.Required))
 }
 
-var unitAmount = regexp.MustCompile(`^\d+(?:B|K|KB|M|m|MB|mb|G|g|GB|gb|T|t|TB|tb)$`)
+var unitAmount = regexp.MustCompile(`^\d+(?:\.\d+)?(?:B|K|KB|M|m|MB|mb|G|g|GB|gb|T|t|TB|tb)$`)
 
 func validateAmountWithUnit(value any) error {
 	v, isNil := validation.Indirect(value)
--- a/api/payloads/manifest_test.go
+++ b/api/payloads/manifest_test.go
@@ -123,6 +123,16 @@ var _ = Describe("Manifest payload", func() {
 				})
 			})
 
+			When("the disk quota is a floating point number", func() {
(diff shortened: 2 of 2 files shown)


danail-branekov:
> but could not find any existing test cases so far and how it could be done with the current code setup.

Maybe [this one](https://github.com/cloudfoundry/korifi/blob/56b15ef34d46d5cdc57fc192b8825ae29d642245/api/payloads/manifest_test.go#L545). change `1G` to `1.5G`?

georgethebeatle:
Both `memory` and `disk_quota` appear on two levels: `ManifestApplication` and `ManifestApplicationProcess`, so I think we should add a decimal value for both in the two respective tests: [here](https://github.com/gogolok/korifi/blob/6ad94507c810a2cdbe2af33c7e7624abe96ca995/api/payloads/manifest_test.go#L389) and [here](https://github.com/cloudfoundry/korifi/blob/56b15ef34d46d5cdc57fc192b8825ae29d642245/api/payloads/manifest_test.go#L389).

danail-branekov (approved):

PR #3573, opened 2024-10-30
Enable purging of CFServiceInstances

klapkov:
Related to https://github.com/cloudfoundry/korifi/issues/3572 

## What is this change about?
Here we enable purging of Service Instances. We take advantage of the Owner Reference to Service Instances in the CFServiceBinding Object. This way, if purge is true, we just remove the finalizer on the CFServiceInstance, so the broker won't be contacted. Since the bindings are a child of the instance, they will be deleted. 

I wanted to include a test case in the service_instance_repository_test for checking if a service binding is purged, but I did not work for some reason. At some point I found out that envtest has some drawbacks in this case. As per: https://book.kubebuilder.io/reference/envtest#testing-considerations - "One common example is garbage collection; because there are no controllers monitoring built-in resources, objects do not get deleted, even if an OwnerReference is set up". So we could not directly verify this via test case.


api/repositories/service_instance_repository.go
```diff
@@ -114,8 +114,8 @@ func (m *ListServiceInstanceMessage) matchesNamespace(ns string) bool {
 }
 
 type DeleteServiceInstanceMessage struct {
-	GUID      string
-	SpaceGUID string
```
klapkov, line 118: SpaceGUID was not used anywhere.

api/handlers/service_instance.go
```diff
@@ -195,14 +195,19 @@ func (h *ServiceInstance) delete(r *http.Request) (*routing.Response, error) {
 
 	serviceInstanceGUID := routing.URLParam(r, "guid")
 
+	payload := new(payloads.ServiceInstanceDelete)
+	if err := h.requestValidator.DecodeAndValidateURLValues(r, payload); err != nil {
+		return nil, apierrors.LogAndReturn(logger, err, "Unable to decode request query parameters")
+	}
+
 	serviceInstance, err := h.serviceInstanceRepo.GetServiceInstance(r.Context(), authInfo, serviceInstanceGUID)
```
klapkov, line 203: Not sure if it is ideal to get the instance in the handler, since now we need to execute get again in the repo, so we can remove the finalizer if purge is true. But since we check whats the instance type so we can return the proper response, i left it for now. One alternative approach would be to return the instance in DeleteServiceInstance in the repo.

api/repositories/service_instance_repository_test.go
```diff
@@ -889,8 +889,8 @@ var _ = Describe("ServiceInstanceRepository", func() {
 			serviceInstance = createServiceInstanceCR(ctx, k8sClient, prefixedGUID("service-instance"), space.Name, "the-service-instance", prefixedGUID("secret"))
 
 			deleteMessage = repositories.DeleteServiceInstanceMessage{
-				GUID:      serviceInstance.Name,
-				SpaceGUID: space.Name,
+				GUID:  serviceInstance.Name,
+				Purge: false,
```
danail-branekov, line 893: `false` is the default value for booleans, so no need to set it explicitly. As per testing with `purge=true`, the test could simply verify that the finalizer on the `CFServiceInstance` object is removed (as this is what the repository is doing)

danail-branekov (dismissed):
Overall looks OK. I would request adding a test into the managed service instance controller to verify that whenever a CFServiceInstance without finalizers is being deleted (i.e. it has its `DeletionTimestamp` set), the controller does not talk to the broker. 

Maybe `When("the instance is purged"` next to [`When("the instance is being deleted`](https://github.com/cloudfoundry/korifi/blob/0a06cd98a8c13264054a17fd6929a955de16a238/controllers/controllers/services/instances/managed/controller_test.go#L560)

api/handlers/service_instance.go
klapkov: Yes, but here https://github.com/cloudfoundry/korifi/blob/7947cc6b4964175f7f822270166c343c1c0c2ba0/api/handlers/service_instance.go#L194 we check the type so we can return the proper response. I guess we can return the service instance type from DeleteServiceInstance :/

api/repositories/service_instance_repository.go
```diff
@@ -420,11 +420,23 @@ func (r *ServiceInstanceRepo) DeleteServiceInstance(ctx context.Context, authInf
...
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      message.GUID,
-			Namespace: message.SpaceGUID,
-		},
+	namespace, err := r.namespaceRetriever.NamespaceFor(ctx, message.GUID, ServiceInstanceResourceType)
+	if err != nil {
+		return fmt.Errorf("failed to get namespace for service instance: %w", err)
+	}
+
+	serviceInstance := &korifiv1alpha1.CFServiceInstance{}
```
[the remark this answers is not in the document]
klapkov: done

api/payloads/service_instance_test.go
```diff
@@ -466,3 +467,41 @@ var _ = Describe("ServiceInstancePatch", func() {
 		})
 	})
 })
+
+var _ = Describe("ServiceInstanceDelete", func() {
+	Describe("decoding from url values", func() {
```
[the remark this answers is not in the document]
klapkov: done

api/handlers/service_instance_test.go
```diff
@@ -520,7 +520,7 @@ var _ = Describe("ServiceInstance", func() {
 			_, actualAuthInfo, message := serviceInstanceRepo.DeleteServiceInstanceArgsForCall(0)
 			Expect(actualAuthInfo).To(Equal(authInfo))
 			Expect(message.GUID).To(Equal("service-instance-guid"))
-			Expect(message.SpaceGUID).To(Equal("space-guid"))
+			Expect(message.Purge).To(BeFalse())
```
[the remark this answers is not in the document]
klapkov: done

api/handlers/service_instance.go
georgethebeatle: Your idea of returning the deleted service type so that we don't have to lookup the instance twice sounds good. However it would look weird for a delete operation's signature to return type and error, so why not return the record for the instance just deleted. This way it would be possible to check its type and the signature will be logical

api/handlers/service_instance.go
```diff
@@ -173,14 +173,19 @@ func (h *ServiceInstance) delete(r *http.Request) (*routing.Response, error) {
...
 	if err != nil {
 		return nil, apierrors.LogAndReturn(logger, apierrors.ForbiddenAsNotFound(err), "failed to get service instance")
 	}
 
 	err = h.serviceInstanceRepo.DeleteServiceInstance(r.Context(), authInfo, repositories.DeleteServiceInstanceMessage{
```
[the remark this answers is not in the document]
klapkov: @georgethebeatle Done

georgethebeatle (approved):

georgethebeatle (approved):

PR #3579, opened 2024-10-30
Add GET v3/service_offerings/:guid

klapkov:

https://github.com/cloudfoundry/korifi/issues/3578

## What is this change about?
Here we add GetServiceOffering - GET v3/service_offerings/:guid

no


api/handlers/service_offering_test.go
```diff
@@ -37,6 +37,123 @@ var _ = Describe("ServiceOffering", func() {
...
+		Describe("include broker field", func() {
+			BeforeEach(func() {
+				serviceBrokerRepo.ListServiceBrokersReturns([]repositories.ServiceBrokerRecord{{
+					ServiceBroker: services.ServiceBroker{
+						Name: "broker-name",
+					},
+					CFResource: model.CFResource{
+						GUID: "broker-guid",
+					},
+				}}, nil)
```
klapkov, line 80: Not sure if we need this, since we already test the include broker field in the ListServiceOfferings test.

danail-branekov: Yeah, looks a bit of a duplication. I would recommend having two whens: `When("including broker name"` and `When("including broker guid` that verify that proper broker fields are being returns. Error cases, as you have pointed out, are already covered in other tests.

api/payloads/service_offering_test.go
```diff
@@ -1,14 +1,46 @@
...
+
+	DescribeTable("invalid query",
+		func(query string, errMatcher types.GomegaMatcher) {
+			_, decodeErr := decodeQuery[payloads.ServiceOfferingGet](query)
+			Expect(decodeErr).To(errMatcher)
+		},
+		Entry("invalid service broker field", "fields[service_broker]=foo", MatchError(ContainSubstring("value must be one of: guid, name"))),
+	)
+
+	It("returns an error if a unsupported param is passed", func() {
```
danail-branekov, line 32: That `it` could be an entry in the `invalid query` table above. Here is a similar [example](https://github.com/cloudfoundry/korifi/blob/d0483c8b89cc8112f731f882a8dfe1485f0b6cb5/api/payloads/service_instance_test.go#L61)

api/presenter/service_offering.go
```diff
@@ -54,3 +59,10 @@ func ForServiceOffering(serviceOffering repositories.ServiceOfferingRecord, base
 		},
 	}
 }
+
+func ForServiceOfferingWithIncluded(serviceOffering repositories.ServiceOfferingRecord, baseURL url.URL, includes ...model.IncludedResource) ServiceOfferingWithIncludedResponse {
```
danail-branekov, line 63: I would suggest adding the `Included` field to the `ServiceOfferingResponse` type rather than coming up with a new one.

api/repositories/service_offering_repository.go
```diff
@@ -61,6 +62,30 @@ func NewServiceOfferingRepo(
...
+	offering := &korifiv1alpha1.CFServiceOffering{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: r.rootNamespace,
+			Name:      guid,
+		},
+	}
+
+	if err = userClient.Get(ctx, client.ObjectKeyFromObject(offering), offering); err != nil {
+		if k8serrors.IsForbidden(err) {
+			return ServiceOfferingRecord{}, nil
```
danail-branekov, line 80: We should not be ignoring the forbidden error on `Get`. Instead, we should translate it to korifi forbidden error. You have actually done that below.
Lets just get rid of the forbidden check.

api/repositories/service_offering_repository_test.go
```diff
@@ -28,6 +28,110 @@ var _ = Describe("ServiceOfferingRepo", func() {
 		)
 	})
 
+	Describe("Get", func() {
```
danail-branekov, line 31: Could you add a test for the cases when getting the offering fails? For example, test that getting an offering with a dummy ID results into an error

danail-branekov (dismissed):

api/presenter/service_offering.go
klapkov: At first I took this approach, but this would mean that we would have to change the way it is handled in the list endpoint as well - https://github.com/cloudfoundry/korifi/blob/main/api/handlers/service_offering.go#L72. There we handle the "includes" with the ForList presenter and including it in the ForServiceOffering() would change ForList as well. Since only the LIST and GET endpoints for service offerings include fields, it did not seem right - https://v3-apidocs.cloudfoundry.org/version/3.180.0/index.html#fields-parameter. What do you think ?

api/handlers/service_offering_test.go
klapkov: Done

api/payloads/service_offering_test.go
klapkov: Done

api/repositories/service_offering_repository.go
klapkov: Done

api/repositories/service_offering_repository_test.go
klapkov: Done

api/presenter/service_offering.go
danail-branekov: Well, if the field is annotated with `omitempty`, then I believe it should be not visible in the response. If I am correct, then 
* the `get` handler **would** set it on the `ServiceOfferingResponse` object
* the `list` handler **would not**, it would use the includes as `ForList` argument. As the `Includes` would not be set on the `ServiceOfferingResponse` object, they would be just omited in the response resources. TL;DR I believe that changing the list endpoint would not be necessary.

api/handlers/service_offering.go
```diff
@@ -50,6 +52,30 @@ func NewServiceOffering(
...
+	}
+
+	serviceOfferingGUID := routing.URLParam(r, "guid")
+
+	serviceOffering, err := h.serviceOfferingRepo.GetServiceOffering(r.Context(), authInfo, serviceOfferingGUID)
+	if err != nil {
+		return nil, apierrors.LogAndReturn(logger, err, "failed to get service offering: %s", serviceOfferingGUID)
+	}
+
+	brokerIncludes, err := h.getBrokerIncludes(r.Context(), authInfo, []repositories.ServiceOfferingRecord{serviceOffering}, payload.IncludeBrokerFields, h.serverURL)
```
georgethebeatle, line 71: The generic include resolver can be used here, you can see how it works [here](https://github.com/cloudfoundry/korifi/blob/df9925de6b12eeb51cc05d3dd70019250b8e0175/api/handlers/service_plan.go#L70) and [here](https://github.com/cloudfoundry/korifi/blob/df9925de6b12eeb51cc05d3dd70019250b8e0175/api/handlers/service_instance.go#L162)

api/handlers/service_offering_test.go
georgethebeatle: After we use the generic include resolver as advised above testing becomes simpler too, for [example](https://github.com/cloudfoundry/korifi/blob/df9925de6b12eeb51cc05d3dd70019250b8e0175/api/handlers/service_plan_test.go#L126-L153)

api/payloads/service_offering.go
```diff
@@ -10,6 +10,25 @@ import (
 	jellidation "github.com/jellydator/validation"
 )
 
+type ServiceOfferingGet struct {
+	IncludeBrokerFields []string
```
georgethebeatle, line 14: We should be using the `IncludeReourceRule` type as we [do](https://github.com/cloudfoundry/korifi/blob/df9925de6b12eeb51cc05d3dd70019250b8e0175/api/payloads/service_plan.go#L29) for plans

api/handlers/service_offering.go
klapkov: In this case should I replace the implementation on the list handler as well ?

api/handlers/service_offering.go
```diff
@@ -116,6 +142,7 @@ func (h *ServiceOffering) UnauthenticatedRoutes() []routing.Route {
 
 func (h *ServiceOffering) AuthenticatedRoutes() []routing.Route {
 	return []routing.Route{
+		{Method: "GET", Pattern: ServiceOfferingPath, Handler: h.get},
```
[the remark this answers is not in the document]
georgethebeatle: Agreed

georgethebeatle (approved):

PR #3580, opened 2024-10-31
Adding v3/apps/guid/droplets endpoint; Fixes #3552

### LINKED ISSUE #3552: Implement the GET /v3/apps/guid/droplets endpoint
### Acceptance

Implement the `GET /v3/apps/guid/droplets` endpoint

### Dev Nots
[API Docs](https://v3-apidocs.cloudfoundry.org/version/3.159.0/index.html#list-droplets-for-an-app)

benjaminguttmann-avtq:
Fixes #3552 

## What is this change about?
Adding /v3/apps/guid/droplets endpoint 

No, adds a new endpoint


api/handlers/app.go
```diff
@@ -217,6 +218,24 @@ func (h *App) setCurrentDroplet(r *http.Request) (*routing.Response, error) {
...
+	appGUID := routing.URLParam(r, "guid")
+
+	app, err := h.appRepo.GetApp(r.Context(), authInfo, appGUID)
+	if err != nil {
+		return nil, apierrors.LogAndReturn(logger, apierrors.ForbiddenAsNotFound(err), "Failed to fetch app from Kubernetes", "AppGUID", appGUID)
+	}
+
+	droplets, err := h.dropletRepo.GetDroplets(r.Context(), authInfo, appGUID)
+	if err != nil {
+		return nil, apierrors.LogAndReturn(logger, apierrors.DropletForbiddenAsNotFound(err), "Failed to fetch droplet from Kubernetes", "dropletGUID", app.DropletGUID)
```
danail-branekov, line 233: I would like to avoid special forbidden-as-not-found errors. The reason we have `ForbiddenAsNotFound` is to mask "forbidden" as not found in order to not tell to unauthorised users that an app with that GUID actually exists. 

TL;DR should't the "standard"  `ForbiddenAsNotFound` work here, instead of `DropletForbiddenAsNotFound`?

api/handlers/app_test.go
```diff
@@ -1306,6 +1306,47 @@ var _ = Describe("App", func() {
...
+				MatchJSONPath("$.resources[0].guid", Equal(dropletGUID)),
+				MatchJSONPath("$.resources[0].relationships.app.data.guid", Equal(appGUID)),
+				MatchJSONPath("$.resources[0].state", Equal("STAGED")),
+			)))
+		})
+
+		When("the app is not accessible", func() {
+			BeforeEach(func() {
+				toReturnErr := apierrors.NewForbiddenError(nil, repositories.AppResourceType)
+				appRepo.GetAppReturns(repositories.AppRecord{}, toReturnErr)
```
danail-branekov, line 1341: I would suggest inlining the `toReturnErr` var. If the line becomes too long, maybe split arguments into multiple lines:

```
appRepo.GetAppReturns(
  repositories.AppRecord{}, 
  apierrors.NewForbiddenError(nil, repositories.AppResourceType),
)
```

api/repositories/droplet_repository.go
```diff
@@ -87,6 +87,40 @@ func (r *DropletRepo) GetDroplet(ctx context.Context, authInfo authorization.Inf
...
+	}
+
+	userClient, err := r.userClientFactory.BuildClient(authInfo)
+	if err != nil {
+		return []DropletRecord{}, fmt.Errorf("failed to build user client: %w", err)
+	}
+
+	var allBuilds []korifiv1alpha1.CFBuild
+	for ns := range namespaces {
+		err := userClient.List(ctx, buildList, client.InNamespace(ns))
```
danail-branekov, line 105: The [cfbuild webhook](https://github.com/cloudfoundry/korifi/blob/d0483c8b89cc8112f731f882a8dfe1485f0b6cb5/controllers/api/v1alpha1/cfbuild_webhook.go#L45) sets the `korifi.cloudfoundry.org/app-guid` label on the CFBuild to the app guid. Therefore, when listing builds you could leverage label selectors. Then the additional filtering on builds below would not be necessary.

Here is an [example](https://github.com/cloudfoundry/korifi/blob/d0483c8b89cc8112f731f882a8dfe1485f0b6cb5/controllers/controllers/workloads/processes/controller.go#L89) how to use label selectors when listing

api/repositories/droplet_repository_test.go
```diff
@@ -313,6 +313,88 @@ var _ = Describe("DropletRepository", func() {
 		})
 	})
 
+	Describe("GetDroplets", func() {
+		var (
+			dropletRecords []repositories.DropletRecord
+			//			appGUID        string
```
danail-branekov, line 319: This comment could be removed

api/repositories/droplet_repository_test.go
```diff
@@ -313,6 +313,88 @@ var _ = Describe("DropletRepository", func() {
...
+					},
+					{
+						Type:    "web",
+						Command: "bundle exec rackup config.ru -p $PORT",
+					},
+				},
+				Ports: []int32{8080, 443},
+			}
+			// Update Build Status based on changes made to local copy
+			Expect(k8sClient.Status().Update(testCtx, build)).To(Succeed())
```
danail-branekov, line 359: We have compe up with our own [utility](https://github.com/cloudfoundry/korifi/blob/d0483c8b89cc8112f731f882a8dfe1485f0b6cb5/tools/k8s/patch.go#L33) to patch resources. It uses k8s client's `Patch` instead of `Update` - `Update`s are quite probable to be rejected in a ditributed systems causing annoying flakes.

FWIW, we have two simolar patch utilities: [`Patch`](https://github.com/cloudfoundry/korifi/blob/d0483c8b89cc8112f731f882a8dfe1485f0b6cb5/tools/k8s/patch.go#L33) and [`PatchResource`](https://github.com/cloudfoundry/korifi/blob/d0483c8b89cc8112f731f882a8dfe1485f0b6cb5/tools/k8s/patch.go#L88). `Patch` would try to patch the `Status` subresource, while `PathResource` works on the main resource only. We use `PatchResource` whenever our modificator function does not touch the status, or to update resources that do not have `Status`, such as `Secret`s.

Here is an [example](https://github.com/cloudfoundry/korifi/blob/d0483c8b89cc8112f731f882a8dfe1485f0b6cb5/api/repositories/droplet_repository_test.go#L106) on how to use `Patch`

api/repositories/droplet_repository.go
```diff
@@ -87,6 +87,40 @@ func (r *DropletRepo) GetDroplet(ctx context.Context, authInfo authorization.Inf
 	return cfBuildToDroplet(build)
 }
 
+func (r *DropletRepo) GetDroplets(ctx context.Context, authInfo authorization.Info, appGuid string) ([]DropletRecord, error) {
```
danail-branekov, line 90: `GetDroplets` repository method has the same essence as [`ListDroplets`](https://github.com/cloudfoundry/korifi/blob/d0483c8b89cc8112f731f882a8dfe1485f0b6cb5/api/repositories/droplet_repository.go#L156). Instead of implementing a new method, I would extend [`ListDropletsMessage`](https://github.com/cloudfoundry/korifi/blob/d0483c8b89cc8112f731f882a8dfe1485f0b6cb5/api/repositories/droplet_repository.go#L71) with `AppGUIDs []string` and make `ListDroplet` filter droplets by app guids if specified in on the message. 

The the [cfbuild webhook](https://github.com/cloudfoundry/korifi/blob/d0483c8b89cc8112f731f882a8dfe1485f0b6cb5/controllers/api/v1alpha1/cfbuild_webhook.go#L45) sets the `korifi.cloudfoundry.org/app-guid` label on the CFBuild to the app guid. Therefore, we could optimise the requests to leverage [set-based label selectors](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#set-based-requirement) in order not to filter collected droplets by app guids as an additional step.

Here is an [example](https://github.com/cloudfoundry/korifi/blob/d0483c8b89cc8112f731f882a8dfe1485f0b6cb5/controllers/controllers/workloads/processes/controller.go#L89) how to use label selectors when listing.

api/repositories/droplet_repository_test.go
```diff
@@ -313,6 +313,88 @@ var _ = Describe("DropletRepository", func() {
 		})
 	})
 
+	Describe("GetDroplets", func() {
```
danail-branekov, line 316: If you agree to go with extending `ListDroplets` rather than introducing `GetDroplets`, then this describe would not be needed. Instead, `ListDroplets` tests should be extended to proove filtering by app guid

danail-branekov (changes requested):
TL;DR, I would recommend to extend the repostirory `ListDroplets` method to support filtering by app guids rather than introducing a new `GetDroplets` method

benjaminguttmann-avtq:
@danail-branekov thanks for detailed feedback, will have a look into it next week :)

api/handlers/app_test.go
```diff
@@ -1306,6 +1306,52 @@ var _ = Describe("App", func() {
...
+					repositories.AppRecord{},
+					apierrors.NewForbiddenError(nil, repositories.AppResourceType),
+				)
+			})
+
+			It("returns an error", func() {
+				expectNotFoundError("App")
+			})
+		})
+	})
```
georgethebeatle, line 1353: Could you please add a `When` where a generic error from the repository results in an `UnknownError`?

benjaminguttmann-avtq:
@danail-branekov @georgethebeatle please review again

georgethebeatle (commented):
He @benjaminguttmann-avtq thank you for your efforts, nice catch using a label selector to make the server do the filtering for us!

I have left some minor comments in-line

api/payloads/app.go
```diff
@@ -204,3 +204,11 @@ func (a *AppPatch) ToMessage(appGUID, spaceGUID string) repositories.PatchAppMes
 
 	return msg
 }
+
+type AppListDroplets struct{}
```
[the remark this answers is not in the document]
benjaminguttmann-avtq: so you mean removing the type and then move the (adjusted) function to handler and then use it there?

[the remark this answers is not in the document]
benjaminguttmann-avtq: if this PR is merged I will prepare a similar change for package payload

api/handlers/app_test.go
```diff
@@ -1306,6 +1306,52 @@ var _ = Describe("App", func() {
...
+				MatchJSONPath("$.pagination.total_results", BeEquivalentTo(1)),
+				MatchJSONPath("$.pagination.first.href", "https://api.example.org/v3/apps/"+appGUID+"/droplets"),
+				MatchJSONPath("$.resources", HaveLen(1)),
+				MatchJSONPath("$.resources[0].guid", Equal(dropletGUID)),
+				MatchJSONPath("$.resources[0].relationships.app.data.guid", Equal(appGUID)),
+				MatchJSONPath("$.resources[0].state", Equal("STAGED")),
+			)))
+		})
+
+		When("the app is not accessible", func() {
```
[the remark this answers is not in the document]
benjaminguttmann-avtq: added more test-cases, I hope this was what you had in mind

api/repositories/droplet_repository.go
```diff
@@ -70,12 +68,18 @@ func (r DropletRecord) Relationships() map[string]string {
...
 type ListDropletsMessage struct {
 	PackageGUIDs []string
+	AppGUIDs     []string
 }
 
-func (m *ListDropletsMessage) matches(b korifiv1alpha1.CFBuild) bool {
-	return tools.EmptyOrContains(m.PackageGUIDs, b.Spec.PackageRef.Name) &&
-		meta.IsStatusConditionFalse(b.Status.Conditions, StagingConditionType) &&
-		meta.IsStatusConditionTrue(b.Status.Conditions, SucceededConditionType)
+func (m *ListDropletsMessage) createSelector() map[string]string {
```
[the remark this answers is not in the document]
benjaminguttmann-avtq: After this one is merged I will check to find other places to apply this as well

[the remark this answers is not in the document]
georgethebeatle: Cool, thank you!

georgethebeatle (approved):

api/handlers/app_test.go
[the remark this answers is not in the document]
georgethebeatle: Looks good, thanks!

PR #3583, opened 2024-11-01
Add GET /v3/apps/:guid/environment_variables endpoint

### LINKED ISSUE #2332: [Feature]: Korifi API users should be able to get app env vars
### Blockers/Dependencies

No response

### Background

As a CF API client I want to be able to get environment variables for apps

### Acceptance Criteria

GIVEN Korifi is deployed
WHEN I request

```
curl "https://api.example.org/v3/apps/[guid]/environment_variables" \
  -X GET \
```

THEN I get response similar to

```
HTTP/1.1 200 OK

{
  "var": {
    "RAILS_ENV": "production"
  },
  "links": {
    "self": {
      "href": "https://api.example.org/v3/apps/[guid]/environment_variables"
    },
    "app": {
      "href": "https://api.example.org/v3/apps/[guid]"
    }
  }
}
```

gogolok:
Fixes #2332

## What is this change about?
It adds the missing endpoint GET /v3/apps/:guid/environment_variables.

No.

## Acceptance Steps
On a running Korifi instance push an app, set a env variable (via cf set-env), get the app guid and then run the `cf curl /v3/apps/:guid/environment_variables` command. It should output something similar to:

```json
{
   "links" : {
      "app" : {
         "href" : "https://localhost/v3/apps/a6cb4e95-65a1-475a-b905-4ce6fac4cf88"
      },
      "self" : {
         "href" : "https://localhost/v3/apps/a6cb4e95-65a1-475a-b905-4ce6fac4cf88/environment_variables"
      }
   },
   "var" : {
      "key0" : "value"
   }
}
```


api/repositories/app_repository.go
```diff
@@ -387,6 +387,36 @@ func (f *AppRepo) ListApps(ctx context.Context, authInfo authorization.Info, mes
 	return f.sorter.Sort(slices.Collect(appRecords), message.OrderBy), nil
 }
 
+func (f *AppRepo) GetAppEnvVars(ctx context.Context, authInfo authorization.Info, appGUID string) (AppEnvVarsRecord, error) {
```
danail-branekov, line 390: Instead of introducing a new method in the app repository, I would suggest you use the already existing `GetAppEnv`(https://github.com/cloudfoundry/korifi/blob/3e88807f9f567544e261d9ed573579f794d162af/api/repositories/app_repository.go#L547). The record's `EnvironmentVariables` should contain what is needed for this endpoint. 

Maybe I would consider merging those [two record types](https://github.com/cloudfoundry/korifi/blob/3e88807f9f567544e261d9ed573579f794d162af/api/repositories/app_repository.go#L165-L178). I cannot remember why we need to have them both, provided that they are almost identical. But merging them could be a separate chore if it turns out to be something big

Or am I missing something?

danail-branekov (changes requested):

api/repositories/app_repository.go
gogolok: The field `EnvironmentVariables` from `GetAppEnv` contains the content we need for the new endpoint.

I haven't picked `GetAppEnv` in my attempt since we might encounter `GetAppEnv` specific errors that might not happen if we would just use the new `GetAppEnvVars` method. For example `GetAppEnv` might fail when fetching `system_env_json` which is not required for `GetAppEnv` at all.  But I guess it's fine if you favor the already existing `GetAppEnv` to safe some lines of code.

I've updated the code in a separate commit to reuse `GetAppEnv`. Please let me know if that's what you've had in mind.

I also looked shortly into merging [two record types](https://github.com/cloudfoundry/korifi/blob/3e88807f9f567544e261d9ed573579f794d162af/api/repositories/app_repository.go#L165-L178) but ran into many code parts that needed a change and overall I failed to do it 'quickly'. I would favor a chore for this.

georgethebeatle (approved):

danail-branekov (approved):

PR #3602, opened 2024-11-07

ddraganovv:
* Added Service Credential Bindings of type Key 
* Added type label to CFServiceBinding CRD
* Tweaked list for bindings so you can filter by type ---------

## What is this change about?

No

### DIFF
--- a/api/payloads/service_binding.go
+++ b/api/payloads/service_binding.go
@@ -1,11 +1,13 @@
 package payloads
 
 import (
+	"fmt"
 	"net/url"
 
 	"code.cloudfoundry.org/korifi/api/payloads/parse"
 	"code.cloudfoundry.org/korifi/api/payloads/validation"
 	"code.cloudfoundry.org/korifi/api/repositories"
+	"code.cloudfoundry.org/korifi/tools"
 	jellidation "github.com/jellydator/validation"
 )
 
@@ -16,18 +18,40 @@ type ServiceBindingCreate struct {
 }
 
 func (p ServiceBindingCreate) ToMessage(spaceGUID string) repositories.CreateServiceBindingMessage {
+	var appGUID string
+	if p.Relationships.App != nil {
+		appGUID = p.Relationships.App.Data.GUID
+	}
+
 	return repositories.CreateServiceBindingMessage{
 		Name:                p.Name,
 		ServiceInstanceGUID: p.Relationships.ServiceInstance.Data.GUID,
-		AppGUID:             p.Relationships.App.Data.GUID,
+		AppGUID:             appGUID,
 		SpaceGUID:           spaceGUID,
+		Type:                p.Type,
 	}
 }
 
 func (p ServiceBindingCreate) Validate() error {
 	return jellidation.ValidateStruct(&p,
-		jellidation.Field(&p.Type, validation.OneOf("app")),
-		jellidation.Field(&p.Relationships, jellidation.NotNil),
+		jellidation.Field(&p.Type, validation.OneOf("app", "key")),
+		jellidation.Field(&p.Relationships, jellidation.By(func(value any) error {
+			relationships, ok := value.(*ServiceBindingRelationships)
+			if !ok || relationships == nil {
+				return fmt.Errorf("relationships is required")
+			}
+
+			if p.Type == "app" {
+				if relationships.App == nil {
+					return jellidation.NewError("when type is app", "relationships.app is required")
+				}
+				if relationships.App.Data.GUID == "" {
+					return fmt.Errorf("relationships.app.data.guid cannot be blank")
+				}
+			}
+
+			return nil
+		})),
 	)
 }
 
@@ -38,12 +62,12 @@ type ServiceBindingRelationships struct {
 
 func (r ServiceBindingRelationships) Validate() error {
 	return jellidation.ValidateStruct(&r,
-		jellidation.Field(&r.App, jellidation.NotNil),
 		jellidation.Field(&r.ServiceInstance, jellidation.NotNil),
 	)
 }
 
 type ServiceBindingList struct {
+	Type                 string
 	AppGUIDs             string
 	ServiceInstanceGUIDs string
 	Include              string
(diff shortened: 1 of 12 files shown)


api/payloads/service_binding.go
```diff
@@ -16,18 +18,40 @@ type ServiceBindingCreate struct {
...
+		Type:                p.Type,
 	}
 }
 
 func (p ServiceBindingCreate) Validate() error {
 	return jellidation.ValidateStruct(&p,
-		jellidation.Field(&p.Type, validation.OneOf("app")),
-		jellidation.Field(&p.Relationships, jellidation.NotNil),
+		jellidation.Field(&p.Type, validation.OneOf("app", "key")),
+		jellidation.Field(&p.Relationships, jellidation.By(func(value any) error {
```
georgethebeatle, line 38: This validation rule can be simplified a bit:

```
func (p ServiceBindingCreate) Validate() error {
	return jellidation.ValidateStruct(&p,
		jellidation.Field(&p.Type, validation.OneOf("app", "key")),
		jellidation.Field(&p.Relationships, jellidation.Required),
		jellidation.Field(&p.Relationships.App, jellidation.Required.When(p.Type == "app")),
		jellidation.Field(&p.Relationships.App.Data.GUID, jellidation.Required.When(p.Type == "app")),
	)
}
```

We can make use of the `jellidation.NotNil` validation rule. I suspect it can be simplified even more but couldn't... It's a complicated conditional validation anyway.

api/payloads/service_binding.go
```diff
@@ -16,18 +18,40 @@ type ServiceBindingCreate struct {
...
 		SpaceGUID:           spaceGUID,
+		Type:                p.Type,
 	}
 }
 
 func (p ServiceBindingCreate) Validate() error {
 	return jellidation.ValidateStruct(&p,
-		jellidation.Field(&p.Type, validation.OneOf("app")),
-		jellidation.Field(&p.Relationships, jellidation.NotNil),
+		jellidation.Field(&p.Type, validation.OneOf("app", "key")),
```
georgethebeatle, line 37: We should also validate the `ServiceBindingCreate.Name` field. According to the cf api [spec](https://v3-apidocs.cloudfoundry.org/version/3.159.0/index.html#create-a-service-credential-binding) it is required for bindings of type`key` and optional for bindings of type `app`. So... some more conditional validation. I can see that right now the `Name` [field](https://github.com/ddraganovv/korifi/blob/f45b42c2f4d67f302dded849e40567f91952531b/api/payloads/service_binding.go#L17) is of type `*string`, but  I guess we should change it to a value and some conditional validation here.... unfortunately.

api/payloads/service_binding_test.go
```diff
@@ -64,6 +78,32 @@ var _ = Describe("ServiceBindingCreate", func() {
...
+				Type: "key",
+			}
+		})
+
+		JustBeforeEach(func() {
+			validatorErr = validator.DecodeAndValidateJSONPayload(createJSONRequest(createPayload), serviceBindingCreate)
+			apiError, _ = validatorErr.(errors.ApiError)
+		})
+
+		It("succeeds", func() {
```
georgethebeatle, line 101: We probably need some negative test here - like when the required `name` field is omited

api/repositories/service_binding_repository.go
```diff
@@ -179,6 +193,8 @@ func (r *ServiceBindingRepo) CreateServiceBinding(ctx context.Context, authInfo
 		}
 	}
 
+	log.Printf("aaa: %+v", cfServiceBinding)
```
georgethebeatle, line 196: Some left-over debug logs, please remove

api/repositories/service_binding_repository_test.go
```diff
@@ -478,7 +479,7 @@ var _ = Describe("ServiceBindingRepo", func() {
 		})
 	})
 
-	Describe("CreateManagedServiceBinding", func() {
+	Describe("CreateManagedServiceBinding of type key", func() {
```
georgethebeatle, line 482: I can see that the `BeforeEach` and `JustBefoereEach` of the describes for type `key` and type `app` are almost identical. I would suggest a test structure as follows:

```
Describe("CreateManagedServiceBinding", func(){
  var message repositories.CreateServiceBindingMessage

  BeforeEach(func(){
    // create the service instance here
  })

  JustBeforeEach(func(){
    // call repo.CreateServiceBinding here using the var'd message
  })

  Describe("type key", func(){
    BeforeEach(fucn(){
      message = // message for type key
    })
  })

  Describe("type app", func(){
    BeforeEach(fucn(){
      message = // message for type app
    })
  })
})
```

controllers/api/v1alpha1/shared_types.go
```diff
@@ -19,6 +19,8 @@ const (
 	CFRouteGUIDLabelKey      = "korifi.cloudfoundry.org/route-guid"
 	CFTaskGUIDLabelKey       = "korifi.cloudfoundry.org/task-guid"
 
+	ServiceCredentialBindingTypeLabel = "korifi.cloudfoundry.org/service-credential-binding-type"
```
georgethebeatle, line 22: I would suggest to use a shorted name like `service-binding-type`. Even though they are really called service credential bindings by the cf v3 api spec we commonly refer to them as service bindings, as it is more concise and does not cause any confusion.

controllers/controllers/services/bindings/controller.go
```diff
@@ -145,6 +144,15 @@ func (r *Reconciler) ReconcileResource(ctx context.Context, cfServiceBinding *ko
 		return res, err
 	}
 
+	if cfServiceBinding.Labels[korifiv1alpha1.ServiceCredentialBindingTypeLabel] == korifiv1alpha1.CFServiceBindingTypeApp {
```
georgethebeatle, line 147: Why do we do this? The old version does not lookup the app and the new version should be concerned with bindings of type `key`.

controllers/webhooks/services/bindings/validator.go
```diff
@@ -47,6 +47,10 @@ func NewCFServiceBindingValidator(duplicateValidator webhooks.NameValidator) *CF
 
 func (v *CFServiceBindingValidator) ValidateCreate(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
 	serviceBinding, ok := obj.(*korifiv1alpha1.CFServiceBinding)
+	if serviceBinding.Labels[korifiv1alpha1.ServiceCredentialBindingTypeLabel] == korifiv1alpha1.CFServiceBindingTypeKey {
```
georgethebeatle, line 50: Why do we do this? If we short-cut the validation we do on create for service bindings of type key, we might get duplicate entries, as we wouldn't invoke the duplicate validator.

georgethebeatle (commented):
Left some in-line comments.

I also deployed this branch and played around with it - tries to follow this [doc](https://docs.cloudfoundry.org/devguide/services/service-keys.html)

Overall it worked, but I noticed that the. `cf service-key` and `cf delete-service-key` commands do not work. Maybe we can address this gap in a future PR (in order to keep this one from growing - it's already quite a beast.

controllers/controllers/services/bindings/controller.go
ddraganovv: When the type of the binding is "key" there is no app associated to it and the call to the k8s api returns an error two lines below the if statement: 
		`err = r.k8sClient.Get(ctx, types.NamespacedName{Name: cfServiceBinding.Spec.AppRef.Name, Namespace: cfServiceBinding.Namespace}, cfApp)`
Do we need to remove the app fetching logic entirely?

PR #3610, opened 2024-11-11
Add Get and GetCredentials for CFServiceInstances

klapkov:

https://github.com/cloudfoundry/korifi/issues/3609
https://github.com/cloudfoundry/korifi/issues/3608

## What is this change about?

Here we expose GET  /v3/service_instances/:guid and we also add GET  /v3/service_instances/:guid/credentials 

No


api/repositories/service_instance_repository.go
```diff
@@ -456,6 +456,36 @@ func (r *ServiceInstanceRepo) GetServiceInstance(ctx context.Context, authInfo a
...
 }
 
+func (r *ServiceInstanceRepo) GetServiceInstanceCredentials(ctx context.Context, authInfo authorization.Info, secretName string) (map[string]any, error) {
+	userClient, err := r.userClientFactory.BuildClient(authInfo)
+	if err != nil {
+		return map[string]any{}, fmt.Errorf("failed to build user client: %w", err)
+	}
+
+	namespace, err := r.namespaceRetriever.NamespaceFor(ctx, secretName, ServiceInstanceResourceType)
+	if err != nil {
```
klapkov, line 466: Maybe this is not the best way to do this, because we get the namespace from the service instance. Should we introduce the corev1.Secret Resource here: https://github.com/cloudfoundry/korifi/blob/main/api/repositories/namespace_retriever.go ? This seams the most reasonable approach to me.

danail-branekov: If we make the method to receive the instance guid (rather than the secret name), we could get the namespace for the service instance (something I believe we already have), not for the secret.

api/repositories/service_instance_repository_test.go
```diff
@@ -1016,6 +1016,78 @@ var _ = Describe("ServiceInstanceRepository", func() {
...
+			secret     *corev1.Secret
+			// serviceInstance *korifiv1alpha1.CFServiceInstance
+			getErr    error
+			decodeErr error
+		)
+
+		BeforeEach(func() {
+			secretName = prefixedGUID("secret")
+
+			_ = createServiceInstanceCR(ctx, k8sClient, secretName, space.Name, "the-service-instance", secretName)
```
danail-branekov, line 1033: We tend to avoid using those `createXXXCR` uitlities astheir usage is not readable and puts stuff on the objects that is not directly needed by the test. Instead, we recommend to create the service instace via the k8s client directly. Make sure to check that creation succeeds.

api/repositories/service_instance_repository_test.go
```diff
@@ -1016,6 +1016,78 @@ var _ = Describe("ServiceInstanceRepository", func() {
...
+			secretData, decodeErr = tools.ToCredentialsSecretData(map[string]any{"foo": "bar"})
```
danail-branekov, line 1034: `decodeErr` can be a local error vairable that can be asserted that does not occur here.

api/repositories/service_instance_repository_test.go
```diff
@@ -1016,6 +1016,78 @@ var _ = Describe("ServiceInstanceRepository", func() {
...
+					secretName = "does-not-exist"
+				})
+
+				It("returns a 404 error", func() {
+					Expect(errors.As(getErr, &apierrors.NotFoundError{})).To(BeTrue())
+				})
+			})
+		})
+
+		When("the context has expired", func() {
```
danail-branekov, line 1078: I assume that this context is to simulate a generic k8s error which is hard to be simulated here. It generally asserts that we do not ignore the errors - something linters do for us. Therefore, please remove the `When` as it looks a bit wierd at first glance?

tools/credentials.go
```diff
@@ -18,3 +18,13 @@ func ToCredentialsSecretData(credentials any) (map[string][]byte, error) {
 		CredentialsSecretKey: credentialBytes,
 	}, nil
 }
+
+func ToDecodedSecretDataCredentials(data map[string][]byte) (map[string]any, error) {
```
danail-branekov, line 22: Maybe `FromCredentialsSecretData` would be a better name as it would be the reverse of `ToCredentialsSecretData`

tools/credentials_test.go
```diff
@@ -27,22 +29,35 @@ var _ = Describe("Credentials", func() {
...
-	It("creates credentials secret data", func() {
-		Expect(secretData).To(MatchAllKeys(Keys{
-			tools.CredentialsSecretKey: MatchJSON(`{
-				"Foo": "foo",
-				"Bar": {
-					"InBar": "in-bar"
-				}
-			}`),
-		}))
+	When("ToCredentialsSecretData", func() {
```
danail-branekov, line 36: This `When` should be a `Describe` as it describes a function under test

tools/credentials_test.go
```diff
@@ -27,22 +29,35 @@ var _ = Describe("Credentials", func() {
 				InBar: "in-bar",
 			},
 		}
-	})
 
-	JustBeforeEach(func() {
-		var err error
 		secretData, err = tools.ToCredentialsSecretData(credsObject)
```
danail-branekov, line 33: Invoking the method under test should be performed within the describe for that function. It should not be invoked in the global describe

tools/credentials_test.go
```diff
@@ -16,8 +16,10 @@ type credsType struct {
 
 var _ = Describe("Credentials", func() {
 	var (
-		credsObject any
-		secretData  map[string][]byte
+		credsObject        any
+		secretData         map[string][]byte
```
danail-branekov, line 20: `secretData` can be local to the `ToCredentialsSecretData` describe

tools/credentials_test.go
```diff
@@ -16,8 +16,10 @@ type credsType struct {
...
+		decodedCredentials map[string]any
```
danail-branekov, line 21: `decodedCredentials` should be pushed down to `ToDecodedSecretDataCredentials` describe

tools/credentials_test.go
```diff
@@ -27,22 +29,35 @@ var _ = Describe("Credentials", func() {
...
+					}
+				}`),
+			}))
+		})
+	})
+
+	When("ToDecodedSecretDataCredentials", func() {
+		BeforeEach(func() {
+			decodedCredentials, err = tools.ToDecodedSecretDataCredentials(secretData)
+			Expect(err).NotTo(HaveOccurred())
```
danail-branekov, line 53: The error should be checked either here, or in the `It`, no need to check it twice. Up to you to decide where.

api/handlers/service_instance.go
```diff
@@ -62,6 +64,50 @@ func NewServiceInstance(
...
+	logger := logr.FromContextOrDiscard(r.Context()).WithName("handlers.service-instance.get-credentials")
+
+	serviceInstanceGUID := routing.URLParam(r, "guid")
+
+	serviceInstance, err := h.serviceInstanceRepo.GetServiceInstance(r.Context(), authInfo, serviceInstanceGUID)
+	if err != nil {
+		return nil, apierrors.LogAndReturn(logger, err, "failed to get service instance")
+	}
+
+	credentials, err := h.serviceInstanceRepo.GetServiceInstanceCredentials(r.Context(), authInfo, serviceInstance.SecretName)
```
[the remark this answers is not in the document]
klapkov: @danail-branekov I talked about this with @georgethebeatle and we agreed to do it this way, since if we use the service instance guid, we rely on the assumption that the secret has the same name as the instance guid.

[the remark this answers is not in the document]
klapkov: But you are right, we can use the instance guid, fetch it in the repo and get the secretName from there.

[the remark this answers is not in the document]
danail-branekov: The suggestion is to pass the service instance guid so that the repository can lookup the service instance again. This way we do not do any assumptions. We have had troubles in the past with such assumptions and we have already [removed](https://github.com/cloudfoundry/korifi/commit/cd3252c255e17c6d2da441ef5f60f837f8114baf#diff-35ec59eae253edff91b574d395a105f99aba017f6f392c4a50253fecd7bf4b82R213) them. Also, for the theoretical kubectl user this assumption has never existed.

We know it is extra work but we should avoid premature optimisations and this endpoint is not that critical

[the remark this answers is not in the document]
klapkov: I understand, will do it this way.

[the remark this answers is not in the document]
danail-branekov: > what is the final decision here
* Use the instance guid in `serviceInstanceRepo.GetServiceInstanceCredentials`
* Get rid of the `SecretName` field of the `ServiceInstanceRecord`

api/handlers/service_instance.go
```diff
@@ -62,6 +64,50 @@ func NewServiceInstance(
...
+
+	includedResources, err := h.includeResolver.ResolveIncludes(r.Context(), authInfo, []repositories.ServiceInstanceRecord{serviceInstance}, payload.IncludeResourceRules)
+	if err != nil {
+		return nil, apierrors.LogAndReturn(logger, err, "failed to build included resources")
+	}
+
+	return routing.NewResponse(http.StatusOK).WithBody(presenter.ForServiceInstance(serviceInstance, h.serverURL, includedResources...)), nil
+}
+
+func (h *ServiceInstance) getCredentials(r *http.Request) (*routing.Response, error) {
```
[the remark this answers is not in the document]
klapkov: @danail-branekov We just get a CF-ResourceNotFound error: 
```
Showing info of service test2 in org batman / space ss as provisioned_user_cf_admin_xHUy1PtcyY6nzPyGVEt0nr7Ax3fXLw...

name:            test2
guid:            093eaedc-4f29-4448-8d5f-449ff867811f
type:            managed
broker:          sm-xsuaa-b122df80-b1ea-44d2-839f-3d17199a5b78
offering:        xsuaa
plan:            default
tags:
offering tags:   xsuaa
description:     Manage application authorizations and trust to identity providers.
documentation:   https://help.sap.com/viewer/65de2977205c403bbc107264b8eccf4b/Cloud/en-US/6373bb7a96114d619bfdfdc6f505d1b9.html
dashboard url:

Showing status of last operation:
   status:    create failed
   message:   Service broker error: service plan not found or not accessible
   started:   2024-10-02T09:41:27Z
   updated:   2024-10-02T09:41:27Z

Showing bound apps:
   There are no bound apps for this service instance.

Showing sharing info:
   This service instance is not currently being shared.

   Service instance sharing is disabled for this service offering.

Showing upgrade status:
   Upgrades are not supported by this broker.

[root@iacbox-v714]:/users/i583051/landscape-aws-staging$ cf curl v3/service_instances/093eaedc-4f29-4448-8d5f-449ff867811f/credentials
{"errors":[{"detail":"Service instance not found","title":"CF-ResourceNotFound","code":10010}]}
```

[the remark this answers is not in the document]
klapkov: Should we just return a CF-ResourceNotFound error when we get the instance? Check if it is user-provided, if not return the error.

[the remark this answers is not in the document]
danail-branekov: Yeah, we should mimic what we have for CF-for-VMs. From the API point of view, the "credentials" resource is not defined for user-provided service instances, therefore returning "not found" makes sense.

klapkov:
@danail-branekov Did some changes, could you take a look.

api/handlers/service_instance.go
```diff
@@ -62,6 +64,54 @@ func NewServiceInstance(
...
+	logger := logr.FromContextOrDiscard(r.Context()).WithName("handlers.service-instance.get-credentials")
+
+	serviceInstanceGUID := routing.URLParam(r, "guid")
+
+	serviceInstance, err := h.serviceInstanceRepo.GetServiceInstance(r.Context(), authInfo, serviceInstanceGUID)
+	if err != nil {
+		return nil, apierrors.LogAndReturn(logger, apierrors.ForbiddenAsNotFound(err), "failed to get service instance", "GUID", serviceInstanceGUID)
+	}
+
+	if serviceInstance.Type != korifiv1alpha1.UserProvidedType {
```
danail-branekov, line 103: I think it would be better to push this check to the repository (reasoning below). The repository looks up the service instance anyway, so it could return a not found error in case the service instance type is not upsi.

From API point of view, instance credentials are just another resource. We strive to keep api handlers simple and only take care to process the request parameters, delegate to a repository and present the result. Business logic on aspects of getting and processing resources should be "outsourced" to repositories.

api/handlers/service_instance.go
```diff
@@ -62,6 +64,50 @@ func NewServiceInstance(
...
+	err := h.requestValidator.DecodeAndValidateURLValues(r, payload)
+	if err != nil {
+		return nil, apierrors.LogAndReturn(logger, err, "Unable to decode request query parameters")
+	}
+
+	serviceInstanceGUID := routing.URLParam(r, "guid")
+
+	serviceInstance, err := h.serviceInstanceRepo.GetServiceInstance(r.Context(), authInfo, serviceInstanceGUID)
+	if err != nil {
+		return nil, apierrors.LogAndReturn(logger, err, "failed to get service instance")
```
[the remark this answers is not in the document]
danail-branekov: This is still not resolved, the api should mask forbidden errors as not found

api/handlers/service_instance_test.go
```diff
@@ -73,6 +73,286 @@ var _ = Describe("ServiceInstance", func() {
...
+		When("getting the service instance fails with forbidden", func() {
+			BeforeEach(func() {
+				serviceInstanceRepo.GetServiceInstanceReturns(
+					repositories.ServiceInstanceRecord{},
+					apierrors.NewForbiddenError(nil, repositories.ServiceInstanceResourceType),
+				)
+			})
+
+			It("returns 404 Not Found", func() {
+				expectNotAuthorizedError()
```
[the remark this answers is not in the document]
danail-branekov: The api should mask forbidden error as not found. Therefore the test should expect a not found error rather then unauthorised. I think you did it the other way round :)

api/handlers/service_instance_test.go
```diff
@@ -73,6 +73,294 @@ var _ = Describe("ServiceInstance", func() {
...
+				serviceInstanceRepo.GetServiceInstanceReturns(repositories.ServiceInstanceRecord{
+					GUID: "service-instance-guid",
+					Type: korifiv1alpha1.UserProvidedType,
+				}, nil)
+
+				serviceInstanceRepo.GetServiceInstanceCredentialsReturns(map[string]any{}, apierrors.NewNotFoundError(nil, repositories.ServiceInstanceResourceType))
+			})
+
+			It("returns 404 Not Found", func() {
+				_, actualAuthInfo, actualInstanceGUID := serviceInstanceRepo.GetServiceInstanceCredentialsArgsForCall(0)
```
danail-branekov, line 320: From my point of view, checking `GetServiceInstanceCredentials` invocation and its parameter [duplicates the checks](https://github.com/cloudfoundry/korifi/pull/3610/files#diff-7e9693740ebe688850df0973f1a6dfd8480f56cd87d4e58e4fadb1fe09929a73R285-R287) in the default case. I think expecting a not found error is enough for this test.

api/handlers/service_instance_test.go
```diff
@@ -73,6 +73,294 @@ var _ = Describe("ServiceInstance", func() {
...
+				Expect(actualAuthInfo).To(Equal(authInfo))
+				Expect(actualInstanceGUID).To(Equal("service-instance-guid"))
+				expectNotFoundError("Service Instance")
+			})
+		})
+
+		When("getting the service instance fails with not found", func() {
```
danail-branekov, line 327: the handler should only interpret the "forbidden" error and present it as not found. Any other sort of errors (regiardless not found, general error, whatever) should be just returned. 

Having said that, I would alter this `When` to verify that when the repository returns a generic error, it results into an unexpected error (HTTP 500). In other words, as the handler does not handle the not found error in any particular way, I would go for the least specific test case.

api/handlers/service_instance_test.go
```diff
@@ -73,6 +73,294 @@ var _ = Describe("ServiceInstance", func() {
...
+					apierrors.NewForbiddenError(nil, repositories.ServiceInstanceResourceType),
+				)
+			})
+
+			It("returns 404 Not Found", func() {
+				expectNotFoundError("Service Instance")
+			})
+		})
+
+		When("the unmarshaling of the secret fails", func() {
```
danail-branekov, line 353: See the comment on the "not found"  error above. The handler does not handle unprocessable entity error in any particular way, it simply returns it. Another context with this particular error does not bring any more value. Therefore I would propose deleting at in order to keep the test simple

api/repositories/service_instance_repository_test.go
```diff
@@ -1016,6 +1016,77 @@ var _ = Describe("ServiceInstanceRepository", func() {
...
+				BeforeEach(func() {
+					instanceGUID = "does-not-exist"
+				})
+
+				It("returns a 404 error", func() {
+					Expect(errors.As(getErr, &apierrors.NotFoundError{})).To(BeTrue())
+				})
+			})
+		})
+	})
```
danail-branekov, line 1088: Could you also add a `When` to verify that the repository returns unprocessiable entity error when the secret data is invalid json?

danail-branekov (commented):
I have resolved the comments that have been already addressed, and left the ones that have not been unresolved. 
I have also added some more comments, please have a look

api/repositories/service_instance_repository_test.go
klapkov: @danail-branekov Maybe this is a stupid question, but how would I achieve that ? I would have to create a secret with an invalid json format, how would I marshal it ?

api/handlers/service_instance.go
danail-branekov: Forget about it, after a short discussion, having that decision in the handler is probably ok. Apologies for the noise

api/repositories/service_instance_repository_test.go
danail-branekov:
```
secret = &corev1.Secret{
				ObjectMeta: metav1.ObjectMeta{
					Name:      secretName,
					Namespace: space.Name,
				},
				Data: map[string][]byte{korifiv1alpha1.CredentialSecretKey, []byte{"not-json"}},
			}
Expect(k8sClient.Create(ctx, secret)).To(Succeed())
```

How about that?

danail-branekov (approved):

PR #3614, opened 2024-11-18
Add DELETE /v3/service_offerings/:guid

klapkov:
 
https://github.com/cloudfoundry/korifi/issues/3613

## What is this change about?
 Here we add deleting and purging of service offerings 

No


api/repositories/service_offering_repository_test.go
```diff
@@ -317,4 +330,136 @@ var _ = Describe("ServiceOfferingRepo", func() {
...
+				Expect(k8serrors.IsNotFound(err)).To(BeTrue(), fmt.Sprintf("error: %+v", err))
+
+				err = k8sClient.Get(context.Background(), types.NamespacedName{Name: instance.Name, Namespace: space.Name}, &korifiv1alpha1.CFServiceInstance{})
+				Expect(k8serrors.IsNotFound(err)).To(BeTrue(), fmt.Sprintf("error: %+v", err))
+
+				serviceBinding := new(korifiv1alpha1.CFServiceBinding)
+				err = k8sClient.Get(context.Background(), types.NamespacedName{Name: binding.Name, Namespace: space.Name}, serviceBinding)
+
+				Expect(err).ToNot(HaveOccurred())
+				Expect(serviceBinding.Finalizers).To(BeEmpty())
```
klapkov, line 461: Here the service binding won't be deleted, but we can still check if the finalizer is removed, which is what we actually do in the repo.

api/repositories/service_offering_repository.go
```diff
@@ -117,3 +157,112 @@ func offeringToRecord(offering korifiv1alpha1.CFServiceOffering) ServiceOffering
...
+
+		filtered := itx.FromSlice(instances.Items).Filter(func(serviceInstance korifiv1alpha1.CFServiceInstance) bool {
+			return tools.EmptyOrContains(planGUIDs, serviceInstance.Spec.PlanGUID)
+		}).Collect()
+
+		serviceInstances = append(serviceInstances, filtered...)
+	}
+
+	return serviceInstances, nil
+}
```
klapkov, line 247: We can combine the fetching of the service instances and bindings under one loop, so we do not need to do it twice, but this way it seems easier to read. Let me know if I should change it.

api/handlers/service_offering_test.go
```diff
@@ -271,4 +272,61 @@ var _ = Describe("ServiceOffering", func() {
...
+			Expect(serviceOfferingRepo.DeleteOfferingCallCount()).To(Equal(1))
+			_, actualAuthInfo, actualDeleteMessage := serviceOfferingRepo.DeleteOfferingArgsForCall(0)
+			Expect(actualAuthInfo).To(Equal(authInfo))
+			Expect(actualDeleteMessage.GUID).To(Equal("offering-guid"))
+			Expect(actualDeleteMessage.Purge).To(BeFalse())
+
+			Expect(rr).To(HaveHTTPStatus(http.StatusNoContent))
+		})
+
+		When("getting the service offering fails with not found", func() {
```
danail-branekov, line 294: The test verifies the error when deleting the offering fails with not found. Therefore, the text of this `When` should be `deleting the service offering fails with not found`

api/repositories/service_offering_repository.go
danail-branekov: Agree, two methods is more readable

api/repositories/service_offering_repository_test.go
```diff
@@ -317,4 +330,136 @@ var _ = Describe("ServiceOfferingRepo", func() {
...
+					},
+				},
+			}
+			Expect(k8sClient.Create(ctx, plan)).To(Succeed())
+
+			instance = createServiceInstanceCR(ctx, k8sClient, uuid.NewString(), space.Name, "my-service-instance", "secret-name")
+			instance.Spec.PlanGUID = plan.Name
+			instance.Finalizers = append(instance.Finalizers, korifiv1alpha1.CFManagedServiceInstanceFinalizerName)
+
+			Expect(k8sClient.Update(ctx, instance)).To(Succeed())
```
danail-branekov, line 386: `k8sClient.Update` would fail if there is an intermediate change on the resource. Therefore, we prefer using `Patch` instead (look at our `k8s.PatchResource` utility)

Alternatively, creating the `instance` literal can be inlined (instead of using the monster `createServiceInstanceCR`) so that needed values can be set on the literal from the very beginning. Then a simple `k8sClient.Create` would create the needed object and there would be no need for a subsequent update.

api/repositories/service_offering_repository_test.go
```diff
@@ -317,4 +330,136 @@ var _ = Describe("ServiceOfferingRepo", func() {
...
+						APIVersion: korifiv1alpha1.SchemeGroupVersion.Identifier(),
+						Name:       instance.Name,
+					},
+					AppRef: corev1.LocalObjectReference{
+						Name: "some-app-guid",
+					},
+				},
+			}
+
+			binding.Finalizers = append(binding.Finalizers, korifiv1alpha1.CFServiceBindingFinalizerName)
```
danail-branekov, line 408: Setting the finalizer could be inlined in the literal above

api/repositories/service_offering_repository_test.go
```diff
@@ -317,4 +330,136 @@ var _ = Describe("ServiceOfferingRepo", func() {
...
+			Expect(k8sClient.Create(ctx, binding)).To(Succeed())
+
+			message = repositories.DeleteServiceOfferingMessage{GUID: offering.Name}
+		})
+
+		JustBeforeEach(func() {
+			createRoleBinding(ctx, userName, spaceDeveloperRole.Name, space.Name)
```
danail-branekov, line 415: Generally, `JustBeforeEach` should be only used to invoke the code under test. Therefore, creating the role binding should be done in the related `BeforeEach`

helm/korifi/controllers/cf_roles/cf_root_namespace_user.yaml
```diff
@@ -46,6 +46,7 @@ rules:
   - cfservicebrokers
   verbs:
   - list
+  - delete
```
danail-branekov, line 49: Adding the `delete` permission here would allow all authenticated users to delete offerings. According to the CF API spec, only admins should be allowed to do that. Therefore, this permission should be added to the [`cf_admin` role](https://github.com/cloudfoundry/korifi/blob/main/helm/korifi/controllers/cf_roles/cf_admin.yaml)

danail-branekov (commented):
Left some comments

api/repositories/service_offering_repository_test.go
klapkov: @danail-branekov Done

georgethebeatle (dismissed):

georgethebeatle (approved):

PR #3616, opened 2024-11-18
Fix service instance purging

klapkov:
* Remove the finalizers from all related bindings

https://github.com/cloudfoundry/korifi/issues/3615

## What is this change about?

Previously  the service bindings did not have a finalizer, but now they do. So when purging a service instance, we need to remove them from every related service binding. 

No

### DIFF
--- a/api/repositories/service_instance_repository_test.go
+++ b/api/repositories/service_instance_repository_test.go
@@ -1028,6 +1028,7 @@ var _ = Describe("ServiceInstanceRepository", func() {
 	Describe("PurgeServiceInstance", func() {
 		var (
 			serviceInstance *korifiv1alpha1.CFServiceInstance
+			serviceBinding  *korifiv1alpha1.CFServiceBinding
 			deleteMessage   repositories.DeleteServiceInstanceMessage
 			deleteErr       error
 		)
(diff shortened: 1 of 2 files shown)


api/repositories/service_instance_repository_test.go
```diff
@@ -1052,13 +1073,14 @@ var _ = Describe("ServiceInstanceRepository", func() {
 		It("purges the service instance", func() {
```
georgethebeatle, line 1073: This test should check that both the instance and the binding are not found in an `Eventually` block. Removing the finalizer is an implementation detail.

klapkov: @georgethebeatle The service binding wont be deleted since the with envtest we do not have the garbage collector controller. As per: https://book.kubebuilder.io/reference/envtest#testing-considerations

danail-branekov: > @georgethebeatle The service binding wont be deleted since the with envtest we do not have the garbage collector controller. As per: [book.kubebuilder.io/reference/envtest#testing-considerations](https://book.kubebuilder.io/reference/envtest#testing-considerations)

makes sense, indeed

danail-branekov (approved):

PR #3620, opened 2024-11-20
Add DELETE /v3/service_plans/:guid

klapkov:

https://github.com/cloudfoundry/korifi/issues/3587
## What is this change about?

With this change we are able to delete a service plan

no


api/handlers/service_plan.go
```diff
@@ -130,6 +132,22 @@ func (h *ServicePlan) updatePlanVisibility(r *http.Request) (*routing.Response,
 	return routing.NewResponse(http.StatusOK).WithBody(presenter.ForServicePlanVisibility(visibility, h.serverURL)), nil
 }
 
+func (h *ServicePlan) delete(r *http.Request) (*routing.Response, error) {
+	authInfo, _ := authorization.InfoFromContext(r.Context())
+	logger := logr.FromContextOrDiscard(r.Context()).WithName("handlers.service-plan.delete")
+
+	planGUID := routing.URLParam(r, "guid")
+	if _, err := h.servicePlanRepo.GetPlan(r.Context(), authInfo, planGUID); err != nil {
```
georgethebeatle, line 140: Can we skip the getting of the plan and just go on and delete it?

api/handlers/service_plan.go
```diff
@@ -130,6 +132,22 @@ func (h *ServicePlan) updatePlanVisibility(r *http.Request) (*routing.Response,
...
+		return nil, apierrors.LogAndReturn(logger, err, "failed to get plan: %s", planGUID)
+	}
+
+	if err := h.servicePlanRepo.DeletePlan(r.Context(), authInfo, planGUID); err != nil {
+		return nil, apierrors.LogAndReturn(logger, err, "failed to delete plan: %s", planGUID)
```
georgethebeatle, line 145: We should use the `apierrors.ForbiddenAsNotFound` to prevent potential leaks of information for plan guids that the user is not authorized to see.

georgethebeatle (commented):
LGTM with minor comments. Added to those I think we need to have a very basic e2e test for completeness

api/handlers/service_plan.go
georgethebeatle: Fair, let's leave it like that

api/handlers/service_plan.go
klapkov: @georgethebeatle Done

georgethebeatle (approved):

georgethebeatle (approved):

Issue #3536, opened 2024-10-22
As a korifi user I want to be able to see the state of the last operation of a managed service instance


georgethebeatle:
### Background

Implement the last operation field for service instance as specified in the [docs](https://v3-apidocs.cloudfoundry.org/version/3.159.0/index.html#the-service-instance-object)

### Dev notes
- Check behaviour of traditional cf for user provided service instances

danail-branekov:
Some experiments with UPSI on mel-c:

### Create UPSI

`> cf create-user-provided-service my-upsi -p '{"k1":"v1"}'`

`> cf services`

```
Getting service instances in org org1 / space space1 as admin...

name      offering        plan   bound apps   last operation     broker   upgrade available
my-upsi   user-provided                       create succeeded
```

`> cf services -v`

```
.....
      "last_operation": {
        "created_at": "2024-11-05T15:48:19Z",
        "description": "Operation succeeded",
        "state": "succeeded",
        "type": "create",
        "updated_at": "2024-11-05T15:48:19Z"
      },
.....
```


### Update UPSI

`> cf update-user-provided-service my-upsi -p '{"foo":"bar"}'`

`> cf services`

```
Getting service instances in org org1 / space space1 as admin...

name      offering        plan   bound apps   last operation     broker   upgrade available
my-upsi   user-provided                       update succeeded
```

`> cf services -v`

```
...
      "last_operation": {
        "created_at": "2024-11-05T15:50:59Z",
        "description": "Operation succeeded",
        "state": "succeeded",
        "type": "update",
        "updated_at": "2024-11-05T15:50:59Z"
      },

...
```

PR #3628, opened 2024-11-25
Implement instances last operation

zabanov-lab:
Previously we were setting the ProvisioningFailedCondition to true if the broker returned error on provision, no matter the error. The problem with doing this is that the error can be recoverable, such as the network flaking, but the instance is sent to a state from which it cannot recover (When the ProvisioningFailedCondition is set the reconciler gives up early in the reconciliation loop).

This change fixes the problem described above by defining a new UnrecoverableError type in the osbapi client. The controller would then set the ProvisioningFailedCondition only if the error it received is an UnrecoverableError.

Also LastOperation field is added in CFServiceInstance and there is a logic implemented in the controllers about the different states and this last operation information is propagated to the presenter logic

According to the OSBAPI spec unrecoverable perovision error codes are 400, 409 and 422.

#3536 
## What is this change about?
Implement instances last operation
No


controllers/controllers/services/instances/managed/controller_test.go
```diff
@@ -330,6 +342,37 @@ var _ = Describe("CFServiceInstance", func() {
...
+				g.Expect(brokerClient.ProvisionCallCount()).To(BeNumerically(">=", 1))
+				g.Expect(instance.Status.LastOperation).To(Equal(korifiv1alpha1.LastOperation{
+					Type:  "create",
+					State: "initial",
+				}))
+			}).Should(Succeed())
+		})
+	})
+
+	When("polling the initial instance last operation", func() {
```
georgethebeatle, line 358: The OSBAPI broker will never return `initial` state, see [spec](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#body-1). This makes the whole `When` redundant.

controllers/controllers/services/instances/upsi/controller.go
```diff
@@ -133,11 +133,29 @@ func (r *Reconciler) ReconcileResource(ctx context.Context, cfServiceInstance *k
 	}
 
 	if err = r.validateCredentials(credentialsSecret); err != nil {
+		cfServiceInstance.Status.LastOperation = korifiv1alpha1.LastOperation{
+			Type:  "create",
+			State: "failed",
+		}
 		return ctrl.Result{}, k8s.NewNotReadyError().WithCause(err).WithReason("SecretInvalid")
```
georgethebeatle, line 140: We should not requeue this error as an invalid secret can only be fiexed by the client. We should return `NewNotReadyError. ... .WithNoRequeue`. As this is hard to test and the test's value is arguable we can skip the test in this case.

controllers/controllers/services/instances/upsi/controller.go
```diff
@@ -133,11 +133,29 @@ func (r *Reconciler) ReconcileResource(ctx context.Context, cfServiceInstance *k
...
 	}
 
 	log.V(1).Info("credentials secret", "name", credentialsSecret.Name, "version", credentialsSecret.ResourceVersion)
 	cfServiceInstance.Status.Credentials = corev1.LocalObjectReference{Name: credentialsSecret.Name}
+
+	if cfServiceInstance.Status.CredentialsObservedVersion == "" {
```
georgethebeatle, line 146: We should extract these two conditionals in a helper function that returns last operation to tidy up the main loop.

controllers/controllers/services/instances/upsi/controller_test.go
```diff
@@ -46,6 +46,13 @@ var _ = Describe("CFServiceInstance", func() {
 				},
 			}
 			Expect(adminClient.Create(ctx, instance)).To(Succeed())
+
+			Expect(k8s.Patch(ctx, adminClient, instance, func() {
```
georgethebeatle, line 50: No need to set last operation here - the controller will do this in the reconciliation loop

georgethebeatle (commented):
LGTM with minor comments

controllers/controllers/services/instances/upsi/controller.go
```diff
@@ -125,11 +126,18 @@ func (r *Reconciler) ReconcileResource(ctx context.Context, cfServiceInstance *k
...
+			Type:  "create",
+			State: "failed",
+		}
+		return ctrl.Result{}, k8s.NewNotReadyError().WithCause(err).WithReason("SecretInvalid").WithNoRequeue()
 	}
 
 	log.V(1).Info("credentials secret", "name", credentialsSecret.Name, "version", credentialsSecret.ResourceVersion)
 	cfServiceInstance.Status.Credentials = corev1.LocalObjectReference{Name: credentialsSecret.Name}
+
+	r.setLastOperation(cfServiceInstance, credentialsSecret)
```
danail-branekov, line 139: I would recommend this function to return the `lastOperation` value rather than set it on the service instance pointer - I think this would be a bit more readable. Also, it could be a static function, does not need to be a reconciler method - the function does not use anything from the receiver. 

TL;DR, something like

```
cfServiceInstance.Status.LastOperation = reconcileLastOperation(cfServiceInstance, credentialsSecret)
```

danail-branekov (commented):
Besides the minor nitpick on the last operation function, looks good

PR #3652, opened 2024-12-04

ddraganovv:

## What is this change about?

No


api/handlers/service_binding_test.go
```diff
@@ -72,476 +72,555 @@ var _ = Describe("ServiceBinding", func() {
...
+
+				It("returns an error and doesn't create the ServiceBinding", func() {
+					expectUnprocessableEntityError("The service instance and the app are in different spaces")
+					Expect(serviceBindingRepo.CreateServiceBindingCallCount()).To(Equal(0))
+				})
 			})
 		})
 
-		When("the user is not authorized", func() {
+		Describe("GET /v3/service_credential_bindings/{guid}", func() {
```
danail-branekov, line 337: The PR moves `Describe("GET /v3/service_credential_bindings/{guid}"` under `Describe("POST /v3/service_credential_bindings"` which is not correct. The `Describe` should remain on the same level as `Describe("POST /v3/service_credential_bindings"`

api/handlers/service_binding_test.go
```diff
@@ -72,476 +72,555 @@ var _ = Describe("ServiceBinding", func() {
...
+				})
+
+				It("returns an error", func() {
+					expectUnknownError()
+				})
 			})
 		})
 
-		When("getting the service binding fails", func() {
+		Describe("DELETE /v3/service_credential_bindings/:guid", func() {
```
danail-branekov, line 453: `"DELETE /v3/service_credential_bindings/:guid"` is now under `Describe("POST /v3/service_credential_bindings"` which is not correct. It should be on the same level as `Describe("POST /v3/service_credential_bindings"`

api/handlers/service_binding_test.go
```diff
@@ -72,476 +72,555 @@ var _ = Describe("ServiceBinding", func() {
...
-				Relationships: &payloads.ServiceBindingRelationships{
-					App: &payloads.Relationship{
-						Data: &payloads.RelationshipData{
-							GUID: "app-guid",
-						},
-					},
-					ServiceInstance: &payloads.Relationship{
-						Data: &payloads.RelationshipData{
-							GUID: "service-instance-guid",
+		When("creating a service binding of type key", func() {
```
danail-branekov, line 75: This newly introduced `When` is a copy of the existing create binding tests. Intead of putting existing tests under `When("creating a service binding of type app"`, I would recommend introduce a context that verifies that unprocessable entity error is returned when creating a binding of type key to a user provided service:


```golang
Describe("POST /v3/service_credential_bindings", func() {
  When("binding to a user provided service instance", func() {
    When("the binding is of type key", func(){
      It("returns an unprocessable entity error", func(){...})
    })
```

api/repositories/service_binding_repository_test.go
```diff
@@ -840,6 +930,9 @@ var _ = Describe("ServiceBindingRepo", func() {
 					Expect(k8s.PatchResource(ctx, k8sClient, serviceBinding3, func() {
 						serviceBinding3.Labels = map[string]string{"not_foo": "NOT_FOO"}
 					})).To(Succeed())
+					Expect(k8s.PatchResource(ctx, k8sClient, serviceBinding4, func() {
```
danail-branekov, line 933: This context is all about filtering by label selector. Does the new binding need to have those labels set? If not, then we would minimise the diff as tests within this context would not change

api/repositories/service_binding_repository_test.go
```diff
@@ -656,10 +717,10 @@ var _ = Describe("ServiceBindingRepo", func() {
 
 	Describe("ListServiceBindings", func() {
```
danail-branekov, line 718: The PR introduces the capability of filtering bindings by type. However, there is no test to verify that. Could you add one?

danail-branekov (commented):
I have added some comments, most of them on the test structure.

Also, this PR contains 30 commits, most of them are already on the main branch. Could you please rebase on top of main?

PR #3662, opened 2024-12-13

ddraganovv:

## What is this change about?

No


api/repositories/service_binding_repository_test.go
```diff
@@ -543,40 +554,61 @@ var _ = Describe("ServiceBindingRepo", func() {
...
-							Kind:       "CFServiceInstance",
-							APIVersion: korifiv1alpha1.SchemeGroupVersion.Identifier(),
-							Name:       cfServiceInstance.Name,
-						},
-						AppRef: corev1.LocalObjectReference{
-							Name: appGUID,
-						},
-					},
-				))
+				Expect(serviceBinding.Spec.Type).To(Equal(korifiv1alpha1.CFServiceBindingTypeApp))
```
danail-branekov, line 557: Checking the spec's field one by one with `Equal` is a bit awkward. What you have done is perfectly fine, but if you want to, you could only match the fields that are interesting to this test. For example, the code does not set the `Spec.Service.APIVersion` explicitly, so no need to check it

Here is an [example](https://github.com/cloudfoundry/korifi/blob/cf08b4aa29b0cd06af66fe72b0ad813d46ea65e0/api/repositories/service_broker_repository_test.go#L294) of `MatchFields` where we only check the fields we are interested in.

api/repositories/service_binding_repository_test.go
```diff
@@ -543,40 +554,61 @@ var _ = Describe("ServiceBindingRepo", func() {
...
+			BeforeEach(func() {
+				createMsg.Type = korifiv1alpha1.CFServiceBindingTypeKey
+				createMsg.AppGUID = ""
+				createRoleBinding(ctx, userName, spaceDeveloperRole.Name, space.Name)
+			})
 
-				It("creates the binding with the specified name", func() {
-					Expect(serviceBindingRecord.Name).To(Equal(bindingName))
-				})
+			It("creates a key binding", func() {
```
danail-branekov, line 587: I would suggesst shortening this `It` to only check the fields of the record that make sense to the `key` binding type, i.e. `Type` being set to `key`, `AppGUID` being empty and `app` relationship not being set. The rest of the fields are being checked in the parent context, so checking them here as well is just unneeded duplication

controllers/controllers/services/bindings/controller.go
```diff
@@ -145,6 +144,10 @@ func (r *Reconciler) ReconcileResource(ctx context.Context, cfServiceBinding *ko
 		return res, err
 	}
 
+	if cfServiceBinding.Spec.Type == korifiv1alpha1.CFServiceBindingTypeKey {
+		return ctrl.Result{}, nil
```
danail-branekov, line 148: There is no controller test that asserts that no servicebindingio bindings are created to bindings of type `key`. Please add one

danail-branekov (commented):
Looks good overall, minor comments. The most important is that there is no controller test that verifies that servicebindingio bindings are not created when the binding is of type `key`

Could you please just push the changes addressing the comments into this branch without recreating the PR? When you recreate the PR my previous comments and review notes are lost and I have to rereview all the changes all over again which takes time

danail-branekov (approved):

danail-branekov (dismissed):

danail-branekov (approved):

danail-branekov:
Closing in favour of https://github.com/cloudfoundry/korifi/pull/3707

Issue #3586, opened 2024-11-01
As a Korifi user, I want service instances not to be deleted when unprovisioning fails for whatever reason


danail-branekov:
### Background
When implementing managed service unprovision we decided that we would [ignore errors](https://github.com/cloudfoundry/korifi/blob/56e6701b570adb64d434df52d4da29fd38a35e76/controllers/controllers/services/instances/managed/controller.go#L258) (note, this method does not return error intentionally) when finalizing. The motivation was that we should perform a best-effort to delete resources and get the instance deleted as quickly as possible.

However, according to the [osbapi spec](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#orphan-mitigation), 

```
and SHOULD keep trying to delete them until the Service Broker responds with a success
```

### Acceptance

**GIVEN** I have created a managed service instance
**WHEN** I try to delete the instance
**WHEN** the broker returns an error response to the unprovision request
**THEN** the instance is **not** deleted
**AND THEN** I see the broker receives new unprovision requests for that instance

**GIVEN** I have created a managed service instance
**WHEN** the broker is not there (someone has deleted it)
**WHEN** I try to delete the instance
**THEN** the instance is **not** deleted

**GIVEN** I have created a managed service instance
**WHEN** the service offering is not available (someone deleted it for some reason)
**WHEN** I try to delete the instance
**THEN** the instance is **not** deleted

**GIVEN** I have created a managed service instance
**WHEN** the service plan is not available (someone deleted it for some reason)
**WHEN** I try to delete the instance
**THEN** the instance is **not** deleted

### Dev notes
* [instance purging](https://github.com/cloudfoundry/korifi/pull/3573) should address instances becoming not deletable
* If an isntance is purged, how should that affect its bindings? Should we ignore the [error](https://github.com/cloudfoundry/korifi/blob/56e6701b570adb64d434df52d4da29fd38a35e76/controllers/controllers/services/osbapi/assets.go#L109-L112) when getting the instance during finalization? Should the binding become unready/failed if its instance is gone?
* While being here, make use of the utility to get service instance context in the controller: https://github.com/cloudfoundry/korifi/blob/56e6701b570adb64d434df52d4da29fd38a35e76/controllers/controllers/services/osbapi/assets.go#L74

PR #3667, opened 2024-12-17
Implement polling for instances deprovisioning

zabanov-lab:
Added if staetement in OSBAPI CLient for unrecoverable errors and implemented polling to Deprovision last operation. Also, when deprovision fails the instance is not deleted, it sets the failed state to instance the last operation

#3586 

No

### DIFF
--- a/controllers/controllers/services/instances/managed/controller_test.go
+++ b/controllers/controllers/services/instances/managed/controller_test.go
@@ -591,108 +591,6 @@ var _ = Describe("CFServiceInstance", func() {
 		})
 	})
 
-	When("the instance is deleted", func() {
-		BeforeEach(func() {
-			brokerClient.DeprovisionReturns(osbapi.ServiceInstanceOperationResponse{
-				Operation: "deprovision-op",
-			}, nil)
-		})
-
-		JustBeforeEach(func() {
-			// For deletion test we want to request deletion and verify the behaviour when finalization fails.
-			// Therefore we use the standard k8s client instncea of `adminClient` as it ensures that the object is deleted
-			Expect(k8sManager.GetClient().Delete(ctx, instance)).To(Succeed())
-		})
-
-		It("deprovisions the instance with the broker", func() {
-			Eventually(func(g Gomega) {
-				err := adminClient.Get(ctx, client.ObjectKeyFromObject(instance), instance)
-				g.Expect(k8serrors.IsNotFound(err)).To(BeTrue())
-
-				g.Expect(brokerClient.DeprovisionCallCount()).To(Equal(1))
-				_, actualDeprovisionRequest := brokerClient.DeprovisionArgsForCall(0)
-				Expect(actualDeprovisionRequest).To(Equal(osbapi.InstanceDeprovisionPayload{
-					ID: instance.Name,
-					InstanceDeprovisionRequest: osbapi.InstanceDeprovisionRequest{
-						ServiceId: "service-offering-id",
-						PlanID:    "service-plan-id",
-					},
-				}))
-			}).Should(Succeed())
-		})
-
-		When("deprovision fails", func() {
-			BeforeEach(func() {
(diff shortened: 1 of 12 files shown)


controllers/controllers/services/instances/managed/controller_test.go
```diff
@@ -626,68 +627,176 @@ var _ = Describe("CFServiceInstance", func() {
...
+				Operation: "deprovision-op",
+			}, nil)
+		})
 
-		When("the service plan is not available", func() {
+		JustBeforeEach(func() {
+			// For deletion test we want to request deletion and verify the behaviour when finalization fails.
+			// Therefore we use the standard k8s client instncea of `adminClient` as it ensures that the object is deleted
+			Expect(adminClient.Get(ctx, client.ObjectKeyFromObject(instance), instance)).To(Succeed())
+			fmt.Printf("instance.Finalizers = %+v\n", instance.Finalizers)
```
danail-branekov, line 651: that is debug statement, we do not need it

controllers/controllers/services/instances/managed/controller_test.go
```diff
@@ -626,68 +627,176 @@ var _ = Describe("CFServiceInstance", func() {
...
+
+			Expect(k8sManager.GetClient().Delete(ctx, instance)).To(Succeed())
+			// Expect(adminClient.Delete(ctx, instance)).To(Succeed())
```
danail-branekov, line 654: this comment should go

controllers/controllers/services/instances/managed/controller_test.go
```diff
@@ -591,7 +592,7 @@ var _ = Describe("CFServiceInstance", func() {
 		})
 	})
 
-	When("the instance is deleted", func() {
+	When("the instance is deleted synchronously", func() {
```
danail-branekov, line 595: I would suggest restructuring the `When("the instance is deleted synchronously")` and `When("the instance is deleted asynchronously")` in order to reduce the duplication.

How about:

```
Describe("instance deletion", func(){
// the default case is sync deletion

		BeforeEach(func() {
			brokerClient.DeprovisionReturns(osbapi.ServiceInstanceOperationResponse{
				Operation: "deprovision-op",
			}, nil)
		})
		JustBeforeEach(func() {
			// For deletion test we want to request deletion and verify the behaviour when finalization fails.
			// Therefore we use the standard k8s client instncea of `adminClient` as it ensures that the object is deleted
			Expect(k8sManager.GetClient().Delete(ctx, instance)).To(Succeed())
		})

// sync deletion its go here

                 When("the instance deletion is asynchronous", func(){
		    BeforeEach(func() {
		  	brokerClient.DeprovisionReturns(osbapi.ServiceInstanceOperationResponse{
		  		IsAsync:   true,
		  		Operation: "deprovision-op",
		  	}, nil)
		    })
                     // async deletion its go here
                 })
})
```

By doing that we would deduplicate `JustBeforeEach`es and keep that pattern that every context has a default case and bunch of corner case contexts

I cannot make my mind what the default case should be though - sync or async deletion - up to you to decide, I guaess whatever makes the code more readable

danail-branekov (commented):
Overall looks good, some polishing in tests is needed.

Also, [linter is unhappy](https://github.com/cloudfoundry/korifi/actions/runs/12368140813/job/34517608526?pr=3667) - `pollProvisionOperation` and `pollDeprovisionOperation` are quite similar.

I am not sure whether extracting the common code in an utility function would make the code look better, but you could give it a try. If not, you should suppress those warnings with a linter instruction comment, look [here](https://github.com/cloudfoundry/korifi/blob/f05ed3c8eb506555bdf43460daa6c7ed63646cdb/api/handlers/domain.go#L89) for an example

danail-branekov (commented):
Left some minor comments. Also, one of the controllers tests is failing: https://github.com/cloudfoundry/korifi/actions/runs/12396243711/job/34603814936?pr=3667

danail-branekov (approved):

danail-branekov (approved):

PR #3688, opened 2025-01-07
Implement unbind from managed service instance

### LINKED ISSUE #3296: As a Korifi API user I want to be able to unbind from a managed service instance
**GIVEN** I have bound an app to a managed service
**WHEN** I `DELETE /v3/service_credential_bindings/my-binding-guid`
**THEN** I get a `202 Accepted` response with `Location` header ppointing me to a binding deletion job
**THEN** eventually I see the binding deletion job to succeed
**THEN** binding data should not be available to the app (after a restart/restage) any more

Dev notes:
* The implementation should call `Unbind` on the service broker
* When requesting unbind from the broker, the implementation should send the `accepts_incomplete=true` query parameter to enable async unbind
* When the broker accepts the request (responds either with 200 OK or `202 Accepted`, the implementation must not check for binding last operation and must remove the finalizer.
* If the operation succeeds, the `CFServiceBinding` along its credentials secrets should be deleted
* If the operation fails, the failure should be reflected in the `CFServiceBinding` status

* [Unbind OSBAPI docs](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#unbinding)
* [Synchronous vs asynchronous operations OSBAPI docs](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#synchronous-and-asynchronous-operations)
* [Managed services proposal](https://docs.google.com/document/d/1d1GSdDVzIMZ9kLedOPT7ZyCqGUwzeoOvHePcrBlYAto/edit?usp=sharing)
* [Delete service binding docs](https://v3-apidocs.cloudfoundry.org/version/3.166.0/index.html#delete-a-service-credential-binding)

zabanov-lab:
- Added IsAsync flag to UnbindResponse in OSBAPI and the logic in the Unbind method
- Added returning Gone error
- Implementing the Unbind call in binding controller and also polling the last operation for the unbind response
- Refactor binding controller so that the calls for the Bind, Unbind and Provison and Deprovision in the services controller are identical
- Implement Setting owner reference with BlockOwnerDeletion=true and added 'foregroundDeletion' finalizer in binding controller

fixes #3296

#3296

No

### DIFF
--- a/controllers/api/v1alpha1/cfservicebinding_types.go
+++ b/controllers/api/v1alpha1/cfservicebinding_types.go
@@ -24,9 +24,9 @@ import (
 )
 
 const (
-	BindingFailedCondition      = "BindingFailed"
-	BindingRequestedCondition   = "BindingRequested"
-	UnbindingRequestedCondition = "UnbindingRequested"
+	BindingFailedCondition = "BindingFailed"
+
+	UnbindingFailedCondition = "UnbindingFailed"
 
 	ServiceInstanceTypeAnnotationKey = "korifi.cloudfoundry.org/service-instance-type"
 	PlanGUIDLabelKey                 = "korifi.cloudfoundry.org/plan-guid"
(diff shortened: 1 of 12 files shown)


danail-branekov (commented):
looks good

danail-branekov (approved):

PR #3693, opened 2025-01-09
Ensure binding credentials are stored on async bind

danail-branekov:
No

## What is this change about?
Ensure credentials secret is created on async bind

* When the bind operation state is `succeeded`, requeue the reconcile
  request. Thus we ensure that the controller would try to `bind` again,
  and as the bind operation has already succeeded, it would get a
  response containing the binding credentials
* When checking the state of the last operation in asynchronous
  provision/bind, instance/binding controllers consider the operation to
  have completed when its state is either `succeeded`, or `failed`.
  Anything else is considered as if the operation is ongoing. Thus we
  protect ourselves from brokers returning nonsense operation state
* The test for the instance controller is refactored so that the
  synchronous provisioning is the "default" case and there is a
  dedicated context for asynchronous provisioning. This aligns test
  style of instance/binding controller tests

* The sample broker implements the async flow properly

No

@georgethebeatle @uzabanov

### DIFF
--- a/controllers/controllers/services/instances/managed/controller_test.go
+++ b/controllers/controllers/services/instances/managed/controller_test.go
@@ -38,14 +38,7 @@ var _ = Describe("CFServiceInstance", func() {
 		brokerClient = new(fake.BrokerClient)
 		brokerClientFactory.CreateClientReturns(brokerClient, nil)
 
-		brokerClient.ProvisionReturns(osbapi.ProvisionResponse{
-			IsAsync:   true,
-			Operation: "operation-1",
-		}, nil)
-
-		brokerClient.GetServiceInstanceLastOperationReturns(osbapi.LastOperationResponse{
-			State: "succeeded",
-		}, nil)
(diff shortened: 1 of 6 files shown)


controllers/controllers/services/instances/managed/controller_test.go
```diff
@@ -296,27 +283,181 @@ var _ = Describe("CFServiceInstance", func() {
...
+						HasType(Equal(korifiv1alpha1.StatusConditionReady)),
+						HasStatus(Equal(metav1.ConditionFalse)),
+					)))
+				}).Should(Succeed())
+			})
+		})
+
+		When("the last operation is in progress", func() {
+			BeforeEach(func() {
+				brokerClient.GetServiceInstanceLastOperationReturns(osbapi.LastOperationResponse{
```
georgethebeatle, line 355: This looks like it duplicates the test starting on line 288

danail-branekov: Indeed, the duplicate has been removed. Thanks!

georgethebeatle (approved):

PR #3697, opened 2025-01-10
Disallow `ClusterStack` configuration

pbusko:

no

## What is this change about?

Make list of buildpacks in the `ClusterStore` configurable via Helm values

no

### DIFF
--- a/helm/korifi/kpack-image-builder/cluster-builder.yaml
+++ b/helm/korifi/kpack-image-builder/cluster-builder.yaml
@@ -10,19 +10,17 @@ spec:
   - image: gcr.io/paketo-buildpacks/ruby
   - image: gcr.io/paketo-buildpacks/procfile
   - image: gcr.io/paketo-buildpacks/go
-
 ---
 apiVersion: kpack.io/v1alpha2
 kind: ClusterStack
 metadata:
   name: cf-default-stack
 spec:
-  id: {{ .Values.kpackImageBuilder.clusterStackID | quote }}
+  id: "io.buildpacks.stacks.jammy"
   buildImage:
-    image: {{ .Values.kpackImageBuilder.clusterStackBuildImage | quote }}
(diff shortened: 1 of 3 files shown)


georgethebeatle:
Hey @pbusko thank you for the PR! The cluster builder, store and stack provided by the helm chart are meant to be [examples](https://github.com/cloudfoundry/korifi/blob/main/INSTALL.md#kpack). If you need custom ones you can [create your own](https://github.com/cloudfoundry/korifi/blob/main/docs/custom-buildpacks.md) and configure the cluster builder name via the `kpackImageBuilder.clusterBuilderName` value. Would that work for you?

pbusko:
In principle you're right, we can just create a separate `ClusterStore`, `ClusterStack` and the `ClusterBuilder` objects, but we noticed that the list of resources already has parametrized parts, like this https://github.com/cloudfoundry/korifi/blob/9e8ee376d78fc79f614db65aada171ab6e879d76/helm/korifi/values.yaml#L108-L110
Wouldn't it make sense to be able to configure everything or nothing? Is there a specific reason why only stack ID/images are configurable?

danail-branekov:
> Wouldn't it make sense to be able to configure everything or nothing? Is there a specific reason why only stack ID/images are configurable?

Good point. We have discussed the question on our standup today and came to the agreement that we should go for the `nothing` option. Would you be willing to PR that instead? I.e., get rid of [those three values](https://github.com/cloudfoundry/korifi/blob/9e8ee376d78fc79f614db65aada171ab6e879d76/helm/korifi/values.yaml#L108-L110) and make the helm chart install the default cluster stack/store/builder only if [`clusterBuilderName`](https://github.com/cloudfoundry/korifi/blob/9e8ee376d78fc79f614db65aada171ab6e879d76/helm/korifi/values.yaml#L106) is not set?

georgethebeatle (approved):

Issue #3649, opened 2024-12-03
[chore] unpin cf cli


danail-branekov:
Revet https://github.com/cloudfoundry/korifi/commit/c3efa00774c258aff05dac3676c9c8f9d11b7adb once latest cli [tag](https://github.com/cloudfoundry/cli/tags) has linux binaries attached

PR #3702, opened 2025-01-13
Unpin CF cli

danail-branekov:
#3649

## What is this change about?
- **Revert "Temporarily pin cf cli version to 8.8.3"**
- **Enable `globstar` option when running bash commands**

No

@georgethebeatle @uzabanov

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,7 +109,7 @@ bin/golangci-lint:
 
 bin/cf:
 	mkdir -p $(GOBIN)
-	curl -fsSL "https://packages.cloudfoundry.org/stable?release=linux64-binary&version=8.8.3&source=github-rel" \
+	curl -fsSL "https://packages.cloudfoundry.org/stable?release=linux64-binary&version=v8&source=github" \
 	  | tar -zx cf8 \
 	  && mv cf8 $(GOBIN)/cf \
 	  && chmod +x $(GOBIN)/cf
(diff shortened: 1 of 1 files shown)


georgethebeatle (approved):

cniles:
@georgethebeatle @danail-branekov, a thought on the setting use of the globstar option.  This breaks compatability with versions of bash older than 4.0.  MacOS doesn't come bundled with bash 4.0 for licensing reasons.  Its simple enough to alter/disable it or even install bash 4.0 through homebrew, but thats a minor irritance.  Maybe its worth considering an alternative that doesn't use the globstar feature?  It looks like the only line that uses it is to clean; an alternative maybe `find ./ -d -name bin -exec rm -rf {} \;` ?

danail-branekov:
Fair point @cniles 

Do you know whether Korifi's Makefile is compatible with older bash versions? If not, we should maybe not bother about it. And I really would not want to downgrade my dev setup to test it.

We use recent bash versions (bash 4 has been released sixteen years ago!) and we do not consider supporting such old versions. But if that is a low-hanging fruit, we could totally do that.

cniles:
Thanks for the response.  All the make targets work.  I feel slightly silly for bringing it up now and take your point of how old it is 😳,  but I do think it could potentially make the local dev setup process less abrasive.

This command appears to have an equivalent effect of removing all the binaries while preserving bin directories:

`find . -wholename '*/bin/*' -delete`

danail-branekov:
It is a bit funny that today a colleague of mine complained that make does not work on his mac anymore :)

Would you be willing to PR a change with the command you pointed out - you would get some contributions credits! I would only request to put a comment on the command why globs are not used - we would probably "fix" it several months later after we have forgotten about this conversation :)

Issue #3723, opened 2025-01-17
Do not lookup the service instance when finalizing an UPSI binding


danail-branekov:
### Bacground

Currently we have a main-delegates architecture of the binding controller. The [main controller](https://github.com/cloudfoundry/korifi/blob/9d92a1b7351c747a8822ead8c0e435aaf1594a54/controllers/controllers/services/bindings/controller.go#L1) [looks up the binding instance](https://github.com/cloudfoundry/korifi/blob/9d92a1b7351c747a8822ead8c0e435aaf1594a54/controllers/controllers/services/bindings/controller.go#L120) and provides it to the appropriate delegate controller.

In the case of UPSI bindings, the delegate controller does not need the service instance when finalizing the binding. It makes no sense finalization to fail if the service instance cannot be looked up

### Action to take
Refactor the delegate controllers to
* Not receinve the instance in their finalize method
* The managed service binding controller should lookup the service instance itself

danail-branekov:
Фиьед бъ https://github.com/cloudfoundry/korifi/pull/3761

PR #3759, opened 2025-01-27
Do not lookup the service instance on UPSI binding finalization

danail-branekov:
#3723

## What is this change about?
UPSI binding finalization should not be failing if the UPSI is not
there for whatever reason.

* Pull the deletion timestamp check up in the "main" binding controller
  prior looking up the instance
* Introduce the `FinalizeResource` method on the sub-controllers
  - The upsi binding controller just removes the finalizer
  - the managed binding controller unprovisions the binding with the
    broker

@georgethebeatle @uzabanov

### DIFF
--- a/controllers/controllers/services/bindings/controller.go
+++ b/controllers/controllers/services/bindings/controller.go
@@ -43,7 +43,8 @@ const (
 )
 
 type DelegateReconciler interface {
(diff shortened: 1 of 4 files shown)


danail-branekov:
Converting the PR to "draft" as I am not really convinced that the refactoring is worth it

danail-branekov:
As discussed with @georgethebeatle  and @uzabanov, this change does not really make sense. Closing

PR #3762, opened 2025-01-27
Implemented GET /v3/service_credential_bindings/:guid/details endpoint

ddraganovv:
No

## What is this change about?
Implementation of an endpoint according to the CF api documentation: 
GET /v3/service_credential_bindings/:guid/details

No


api/repositories/service_binding_repository.go
```diff
@@ -75,6 +77,12 @@ func (r ServiceBindingRecord) Relationships() map[string]string {
 	}
 }
 
+type Credentials map[string]any
+type ServiceBindingDetailsRecord struct {
+	Credentials    Credentials
+	SyslogDrainURL *string
```
danail-branekov, line 83: Provided that the repository never populates `SyslogDrainURL` and `VolumeMounts`, I would suggest removing them from the record. We can introduce them when we properly implement them.

api/repositories/service_binding_repository.go
```diff
@@ -75,6 +77,12 @@ func (r ServiceBindingRecord) Relationships() map[string]string {
 	}
 }
 
+type Credentials map[string]any
```
danail-branekov, line 80: Is there any reason to alias `map[string]any` as `Credentials`? I.e. is the `Credentials` type needed?

api/presenter/service_binding.go
```diff
@@ -41,6 +41,12 @@ type ServiceBindingLinks struct {
 	Details         Link `json:"details"`
 }
 
+type ServiceBindingDetailsResponse struct {
+	Credentials    map[string]any `json:"credentials"`
+	SyslogDrainURL *string        `json:"syslog_drain_url,omitempty"`
```
danail-branekov, line 46: Similar to the `ServiceBindingDetailsRecord`, we do not support `SyslogDrainURL` and `VolumeMounts`, so I would recommend removing them from the response type.

api/repositories/service_binding_repository.go
```diff
@@ -272,6 +280,52 @@ func (r *ServiceBindingRepo) GetServiceBinding(ctx context.Context, authInfo aut
 	return serviceBindingToRecord(*serviceBinding), nil
 }
 
+func (r *ServiceBindingRepo) GetServiceBindingDetails(ctx context.Context, authInfo authorization.Info, bindingGUID string) (ServiceBindingDetailsRecord, error) {
+	userClient, err := r.userClientFactory.BuildClient(authInfo)
```
danail-branekov, line 284: Getting the `CFServiceBinding` here is the same as getting it in [GetServiceBinding](https://github.com/cloudfoundry/korifi/blob/89d5234409cc06abb6462b21cd34cf6f2435212e/api/repositories/service_binding_repository.go#L256-L270). I would suggest extracting a reusable utility function to remove code duplication

api/repositories/service_binding_repository.go
```diff
@@ -272,6 +280,52 @@ func (r *ServiceBindingRepo) GetServiceBinding(ctx context.Context, authInfo aut
...
+			Name:      serviceBinding.Status.Credentials.Name,
+		},
+	}
+	err = userClient.Get(ctx, client.ObjectKeyFromObject(credentialsSecret), credentialsSecret)
+	if err != nil {
+		return ServiceBindingDetailsRecord{}, fmt.Errorf("failed to get credentials: %w", apierrors.ForbiddenAsNotFound(apierrors.FromK8sError(err, ServiceBindingResourceType))) //
+	}
+
+	credentials := map[string]any{}
+	err = json.Unmarshal(credentialsSecret.Data[korifiv1alpha1.CredentialsSecretKey], &credentials)
```
danail-branekov, line 321: I would suggest using [this utility](https://github.com/cloudfoundry/korifi/blob/89d5234409cc06abb6462b21cd34cf6f2435212e/tools/secrets.go#L34) to convert secret data into credentials map.

api/repositories/service_binding_repository_test.go
```diff
@@ -1238,4 +1239,114 @@ var _ = Describe("ServiceBindingRepo", func() {
...
+				},
+			}
+
+			Expect(
+				k8sClient.Create(ctx, serviceBinding),
+			).To(Succeed())
+
+			Expect(
+				k8s.Patch(ctx, k8sClient, serviceBinding, func() {
+					serviceBinding.Status.Credentials.Name = serviceInstanceGUID
```
danail-branekov, line 1310: `serviceBinding.Status.Credentials.Name` is set to the secret that contains the credentials (see [here](https://github.com/cloudfoundry/korifi/blob/a3f8b883b33bf18e166a8d9385c134439e9ec3b9/controllers/controllers/services/bindings/managed/controller.go#L220)), it has nothing to do with the service instance. 
Therefore, the test could just create a credentials secret and set its name on the binding status, no need to create a service instance

api/repositories/service_binding_repository_test.go
```diff
@@ -1238,4 +1239,114 @@ var _ = Describe("ServiceBindingRepo", func() {
...
+
+			credentials := map[string]any{
+				"foo": "val1",
+				"bar": "val2",
+			}
+
+			credentialBytes, err := json.Marshal(credentials)
+			Expect(err).To(BeNil())
+
+			creds := map[string][]byte{
```
danail-branekov, line 1265: You could use [ToCredentialsSecretData](https://github.com/cloudfoundry/korifi/blob/89d5234409cc06abb6462b21cd34cf6f2435212e/tools/secrets.go#L14) to create the secret data in the correct format

api/repositories/service_binding_repository_test.go
```diff
@@ -1238,4 +1239,114 @@ var _ = Describe("ServiceBindingRepo", func() {
...
+		})
+
+		When("the user is a space developer", func() {
+			BeforeEach(func() {
+				createRoleBinding(ctx, userName, spaceDeveloperRole.Name, space.Name)
+			})
+
+			It("returns details", func() {
+				Expect(getErr).To(BeNil())
+				Expect(serviceBindingRecord.Credentials).To(HaveKeyWithValue("bar", "val2"))
```
danail-branekov, line 1337: A more concise way to assert two things would be 

```golang
Expect(serviceBindingRecord.Credentials).To(SatisfyAll(
  HaveKeyWithValue("bar", "val2"),
  HaveKeyWithValue("foo", "val1"),
))
```

Or even simpler:

```goalng
Expect(serviceBindingRecord.Credentials).To(Equal(map[string]any{
  "bar": "val2",
  "foo": "val1",
}))
```

api/repositories/service_binding_repository_test.go
```diff
@@ -1238,4 +1239,114 @@ var _ = Describe("ServiceBindingRepo", func() {
...
+			getErr               error
+			cfServiceInstance    *korifiv1alpha1.CFServiceInstance
+			serviceBindingRecord repositories.ServiceBindingDetailsRecord
+			credentialsSecret    *corev1.Secret
+		)
+
+		BeforeEach(func() {
+			serviceInstanceGUID := prefixedGUID("instance")
+			cfServiceInstance = createServiceInstanceCR(ctx, k8sClient, serviceInstanceGUID, space.Name, "service-instance-name", serviceInstanceGUID)
+			serviceBindingGUID = prefixedGUID("binding")
```
danail-branekov, line 1255: We generally prefer `uuid.NewString()` instead of those `prefixedXXX` utilities. We are trying to slowly get rid of them alltogether, let's not use them in new code.

api/repositories/service_binding_repository_test.go
```diff
@@ -1238,4 +1239,114 @@ var _ = Describe("ServiceBindingRepo", func() {
...
+
+			It("returns details", func() {
+				Expect(getErr).To(BeNil())
+				Expect(serviceBindingRecord.Credentials).To(HaveKeyWithValue("bar", "val2"))
+				Expect(serviceBindingRecord.Credentials).To(HaveKeyWithValue("foo", "val1"))
+			})
+
+			When("binding does not exist", func() {
+				BeforeEach(func() {
+					serviceBindingGUID = "i-dont-exist"
```
danail-branekov, line 1343: Setting the guid variable has no effect on the binding, I am surprised the test succeeds. Instead, this `BeforeEach` should patch `binding.Status.Credentials.Name` to `i-do-not-exist`

controllers/api/v1alpha1/cfserviceinstance_types.go
```diff
@@ -25,8 +25,9 @@ import (
 )
 
 const (
-	UserProvidedType = "user-provided"
-	ManagedType      = "managed"
+	UserProvidedType     = "user-provided"
+	ManagedType          = "managed"
+	CredentialsSecretKey = "credentials"
```
danail-branekov, line 30: We already have a [constant](https://github.com/cloudfoundry/korifi/blob/89d5234409cc06abb6462b21cd34cf6f2435212e/tools/secrets.go#L10) for that key, no need of a new one. But if you use the utilities from the `tools/secrets.go` file, you should probably not need to reference it.

danail-branekov:
As this is a new endpoint, I would also ask you to add an E2E test (in `tests/e2e/service_bindings_test.go`). In e2e tests we test the very-very happy path to verify that the invocation of the endpoint succeeds and maybe a field or two in the response. Thus we verify that endpoints are properly wired in the api main. No need to test error cases there, they are already tested with unit/integration tests.

api/repositories/service_binding_repository_test.go
ddraganovv: It is my understanding that the variable `serviceBindingGUID` is used in the `JustBeforeEach()` node and setting it to an invalid binding guid will cause the `repo.GetServiceBindingDetails` to fail. Am I missing something ?

danail-branekov: Oh, I see. I missed that the intent of this `BeforeEach` is to make the invocation of `JustBeforeEach` fail. All good then :)

api/handlers/service_binding_test.go
```diff
@@ -568,4 +574,44 @@ var _ = Describe("ServiceBinding", func() {
...
+			requestPath = "/v3/service_credential_bindings/service-binding-guid/details"
+			requestBody = ""
+		})
+
+		It("returns the service binding details", func() {
+			Expect(rr).To(HaveHTTPStatus(http.StatusOK))
+			Expect(rr).To(HaveHTTPHeaderWithValue("Content-Type", "application/json"))
+			Expect(rr).To(HaveHTTPBody(SatisfyAll(
+				MatchJSONPath("$.credentials.connection", "mydb://user@password:example.com"),
+				MatchJSONPath("$.syslog_drain_url", "http://syslog.example.com/drain"),
```
georgethebeatle, line 590: With regards to `syslog_drain_url` and `volume_mounts`, we would better not return anything. Furnthermore these values are not stubbed in the before each, so this test would probably fail.

api/repositories/service_binding_repository.go
```diff
@@ -253,23 +257,65 @@ func (r *ServiceBindingRepo) DeleteServiceBinding(ctx context.Context, authInfo
 }
 
 func (r *ServiceBindingRepo) GetServiceBinding(ctx context.Context, authInfo authorization.Info, guid string) (ServiceBindingRecord, error) {
-	ns, err := r.namespaceRetriever.NamespaceFor(ctx, guid, ServiceBindingResourceType)
+	binding, err := r.getServiceBinding(ctx, authInfo, guid)
 	if err != nil {
-		return ServiceBindingRecord{}, err
+		return ServiceBindingRecord{}, fmt.Errorf("get-service-binding failed to create user client: %w", err)
```
georgethebeatle, line 262: This error is already wrapped, no need to wrap again, furthermore the description is incorrect.

api/repositories/service_binding_repository_test.go
georgethebeatle: Any reason to user `prefixedGUID` instead of `uuid.NewString()`?

tests/e2e/service_bindings_test.go
```diff
@@ -169,6 +169,31 @@ var _ = Describe("Service Bindings", func() {
 		})
 	})
 
+	Describe("GET /v3/service_credential_bindings/{guid}/details", func() {
+		var result credentialsResponse
+
+		BeforeEach(func() {
+			brokerGUID := createBroker(serviceBrokerURL)
```
georgethebeatle, line 176: This test would be much simpler if we create a binding to a user provided service instance. The details endpoint is valid for both types of service binding.

api/repositories/service_binding_repository_test.go
ddraganovv: > Any reason to user `prefixedGUID` instead of `uuid.NewString()`?

My assumption is readability in the test logs.

danail-branekov (approved):

danail-branekov:
Looks good overall, you need to address linter failures though

api/handlers/service_binding_test.go
```diff
@@ -50,6 +52,12 @@ var _ = Describe("ServiceBinding", func() {
 			Type:      korifiv1alpha1.UserProvidedType,
 		}, nil)
 
+		serviceBindingRepo.GetServiceBindingDetailsReturns(repositories.ServiceBindingDetailsRecord{
```
danail-branekov, line 55: Sertting up the serviceBindingRepo to return the binding details only makes sense to `Describe("GET /v3/service_credential_bindings/{guid}/details"`, the rest of the test does not need it. Could you move that in that `Describe` `BeforeEach`?

danail-branekov (approved):

PR #3763, opened 2025-01-27
Create serviceBinding.io secret with type app only

ddraganovv:
GitHub [Issue](https://github.com/cloudfoundry/korifi/issues/3739)

## What is this change about?
When the type of the credential binding is 'key' creating a servicebinding.io secret is redundant. This PR's goal is changing the way the controller creates sbio secrets so it is only created when the type is 'app'

No

## Acceptance Steps
The servicebinding.io secret should not be created when the type of the credential binding is 'app'

### DIFF
--- a/controllers/controllers/services/bindings/managed/controller.go
+++ b/controllers/controllers/services/bindings/managed/controller.go
@@ -219,28 +219,6 @@ func (r *ManagedBindingsReconciler) reconcileCredentials(ctx context.Context, cf
 	}
 	cfServiceBinding.Status.Credentials.Name = credentialsSecret.Name
 
-	bindingSecret := &corev1.Secret{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      cfServiceBinding.Name + "-sbio",
-			Namespace: cfServiceBinding.Namespace,
-		},
-	}
-	_, err = controllerutil.CreateOrPatch(ctx, r.k8sClient, bindingSecret, func() error {
-		bindingSecret.Type = corev1.SecretType(credentials.ServiceBindingSecretTypePrefix + korifiv1alpha1.ManagedType)
-		bindingSecret.Data, err = credentials.GetServiceBindingIOSecretData(credentialsSecret)
-		if err != nil {
-			return err
-		}
-
-		return controllerutil.SetControllerReference(cfServiceBinding, bindingSecret, r.scheme)
-	})
-	if err != nil {
-		log.Error(err, "failed to create binding secret")
-		return err
-	}
-
-	cfServiceBinding.Status.Binding.Name = bindingSecret.Name
-
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


controllers/controllers/services/bindings/managed/controller.go
```diff
@@ -273,9 +252,41 @@ func (r *ManagedBindingsReconciler) finalizeCFServiceBinding(
 }
 
 func (r *ManagedBindingsReconciler) reconcileSBServiceBinding(ctx context.Context, cfServiceBinding *korifiv1alpha1.CFServiceBinding) (*servicebindingv1beta1.ServiceBinding, error) {
+	log := logr.FromContextOrDiscard(ctx)
+
+	secretLookupKey := types.NamespacedName{Name: cfServiceBinding.Name, Namespace: cfServiceBinding.Namespace}
+	createdSecret := new(corev1.Secret)
```
danail-branekov, line 258: I would suggest renaming `createdSecret` to `credentialsSecret` as the secret here stores the credentials for the binding. This name would convey what that secret is (improves readability)

I would also change the way the secret is being looked up to be align with the way we lookup objects:

```golang
credentialsSecret := &corev1.Secret{
	ObjectMeta: metav1.ObjectMeta{
			Name:      cfServiceBinding.Name,
			Namespace: cfServiceBinding.Namespace,
		},
}
if err := r.k8sClient.Get(ctx, client.ObjectKeyFromObject(credentialsSecret), credentialsSecret); err != nil {
	return nil, err
}
```

controllers/controllers/services/bindings/managed/controller.go
```diff
@@ -273,9 +252,41 @@ func (r *ManagedBindingsReconciler) finalizeCFServiceBinding(
...
+	}
+
+	bindingSecret := &corev1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      cfServiceBinding.Name + "-sbio",
+			Namespace: cfServiceBinding.Namespace,
+		},
+	}
+
+	_, err := controllerutil.CreateOrPatch(ctx, r.k8sClient, bindingSecret, func() error {
```
danail-branekov, line 270: Could you please add a test that verifies that the sbio secret is not created when the binding is of type `key`?

Maybe [here](https://github.com/cloudfoundry/korifi/blob/89d5234409cc06abb6462b21cd34cf6f2435212e/controllers/controllers/services/bindings/controller_test.go#L772) add something like

```golang
It("does not create a servicebindingio secret", func(){
  Consistently(func(g Gomega) {
    g.Expect(adminClient.Get(ctx, client.ObjectKeyFromObject(binding), binding)).To(Succeed())
    g.Expect(binding.Status.Binding.Name).To(BeEmpty())
    
    secrets := &corev1.SecretList{}
    g.Expect(k8sClient.List(ctx, secrets, client.InNamespace(binding.Namespace))).To(Succeed())  
    g.Expect(secrets.Items).NotTo(ContainElement(
        MatchFields(IgnoreExtras, Fields{"Type": HasPrefix(credentials.ServiceBindingSecretTypePrefix))),
    ))
  }).Should(Succeed())
}
```

georgethebeatle (approved):

PR #3775, opened 2025-02-03
Makefile: build-dorifi: Set GOARCH to `amd64`

gogolok:
No.

## What is this change about?
I'm running the smoke tests from my local arm64 machine against an amd64 Korifi instance.
When creating the sample apps on `arm64` machine, but pushing to `amd64` Korifi instance, the tests will fail.

No, as long as the smoke apps run on amd64 for the time being.

## Acceptance Steps
Run smoke tests.

## Things to remember
Might be possible to switch to go buildpacks instead of binaries in the future.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,9 +82,9 @@ test-smoke: build-dorifi bin/cf
 
 
 build-dorifi:
-	CGO_ENABLED=0 GOOS=linux go build -C tests/assets/dorifi-golang -o ../dorifi/dorifi .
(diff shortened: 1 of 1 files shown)


danail-branekov:
Thanks @gogolok 

We have decided to go for a `procfile` test apps vs buildpack ones as by using binaries we spare the buildpack build when running tests.

Out of curiosity, have you tried setting the `GOARCH` env var when running the make target on your environment? I.e. `GOARCH=amd64 make test-smoke`? If that works, we would avoid hardcoding it.

gogolok:
As expected, GOARCH works (for me). Closing.

Issue #3755, opened 2025-01-24
Make sure that app/process stats endpoints call the logcache api


georgethebeatle:
### Background

After #3752 exports app/process metrics to the `/api/v1/read` endpoint we should make sure that  `/v3/apps/<app-guid>/<process-type>/stats` and `/v3/processes/<process-guid>/stats` delegate to `/api/v1/read`.

### Actions
- [x] The AppWorkload status should contain information about the [state](https://github.com/cloudfoundry/korifi/blob/eea9e64bff2bca39244478b5ea1bfd4cb2b111e0/api/actions/process_stats.go#L174-L215) of each instance id.
- [x] The CFProcess status should propagate the AppWorkload instance statuses to the CFProcess Status
- [x] The Process handler should build process stats based on log cache envelopes and CFProcess' Status
- [x] The App handler should build process stats based on log cache envelopes and CFProcess' Status
- [x] Clean the process stats action from any Pod references, as it should only work with the log cache API

danail-branekov:
The stats action is now used by the logcache handler only. We would not touch it with this story, we would look for refactoring opportunities in #3812

PR #3781, opened 2025-02-05
Introduce InstancesState in the AppWorkload resrouce status

georgethebeatle:
#3755

## What is this change about?
Today korifi looks directly at the pods running on the cluster, but once
the user provides a custom appworkload controller, this does not work
anymore, as it breaches the abstraction. Korifi should not rely on any
concorete implementation below the AppWorkload interface resource

To fix this we have introduced a new InstancesState field in the AppWorkload resource status.


tcdowney (approved):
This is a good idea 👍

danail-branekov (approved):

PR #3787, opened 2025-02-10
Added GetServiceBinding to the broker client

ddraganovv:
[Issue](https://github.com/cloudfoundry/korifi/issues/3751)

## What is this change about?
This is 1 of 2 PRs related to the above issue. With this PR there is a new method introduced to the broker client which is responsible to retrieve the service binding. 

No

### DIFF
--- a/controllers/controllers/services/osbapi/client.go
+++ b/controllers/controllers/services/osbapi/client.go
@@ -298,6 +298,36 @@ func (c *Client) Unbind(ctx context.Context, payload UnbindPayload) (UnbindRespo
 	return response, nil
 }
 
+func (c *Client) GetServiceBinding(ctx context.Context, payload BindPayload) (BindingResponse, error) {
+	statusCode, respBytes, err := c.newBrokerRequester().
+		forBroker(c.broker).
+		sendRequest(
+			ctx,
+			"/v2/service_instances/"+payload.InstanceID+"/service_bindings/"+payload.BindingID,
+			http.MethodGet,
+			map[string]string{
+				"service_id": payload.ServiceId,
+				"plan_id":    payload.PlanID,
+			},
+			nil,
+		)
+	if err != nil {
+		return BindingResponse{}, fmt.Errorf("fetching service binding failed: %w", err)
+	}
+
+	if statusCode == http.StatusNotFound {
+		return BindingResponse{}, UnrecoverableError{Status: statusCode}
+	}
+
+	response := BindingResponse{}
+	err = json.Unmarshal(respBytes, &response)
+	if err != nil {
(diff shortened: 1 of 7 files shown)


controllers/controllers/services/osbapi/client.go
```diff
@@ -298,6 +298,33 @@ func (c *Client) Unbind(ctx context.Context, payload UnbindPayload) (UnbindRespo
 	return response, nil
 }
 
+func (c *Client) GetServiceBinding(ctx context.Context, payload BindPayload) (BindingResponse, error) {
+	statusCode, respBytes, err := c.newBrokerRequester().
+		forBroker(c.broker).
+		sendRequest(
+			ctx,
+			"/v2/service_instances/"+payload.InstanceID+"/service_bindings/"+payload.BindingID,
```
danail-branekov, line 306: According to the [OSBAPI spec](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#parameters-7), broker clients could provide the plan and offering ids as query parameters (as brokers seem to find such hints useful).

We are providing those IDs in the rest of the methods in the clients, we should do that here as well.

controllers/controllers/services/osbapi/client.go
```diff
@@ -298,6 +298,33 @@ func (c *Client) Unbind(ctx context.Context, payload UnbindPayload) (UnbindRespo
...
+			http.MethodGet,
+			nil,
```
danail-branekov, line 308: According to the [OSBAPI spec](https://github.com/openservicebrokerapi/servicebroker/blob/master/spec.md#parameters-7), broker clients could provide the plan and offering ids as query parameters (as brokers seem to find such hints useful).

We are providing those IDs in the rest of the methods in the clients, we should do that here as well.

danail-branekov (commented):
I have left a small comment, looks good otherwise

controllers/controllers/services/osbapi/client_test.go
```diff
@@ -619,6 +620,59 @@ var _ = Describe("OSBAPI Client", func() {
...
+				})
+			})
+
+			It("gets the service binding", func() {
+				Expect(getBindErr).NotTo(HaveOccurred())
+
+				requests := brokerServer.ServedRequests()
+				Expect(requests).To(HaveLen(1))
+				Expect(requests[0].Method).To(Equal(http.MethodGet))
+				Expect(requests[0].URL.Path).To(Equal("/v2/service_instances/my-service-instance/service_bindings/my-binding-id"))
```
danail-branekov, line 652: This test should also check the query parameters (offering and plan id) of the request url

danail-branekov (commented):
Almost there, there is a minor gap in the new method test

PR #3788, opened 2025-02-10
Add process guid to gauge envelopes

uzabanov:
#3755

## What is this change about?
The log cache api will be used by the process and app handlers, so that they
can fetch stats for a process with a particular guid. Therefore it makes
sense to filter the envelope stream for envelopes for that process guid
only. In order to do so we need to have access to the process guid that
the envelope is related to.

The cloud controller implementation is doing the same: https://github.com/cloudfoundry/cloud_controller_ng/blob/5b1af263bbb21793ba7d49f64c75f6ab5f5ab267/lib/cloud_controller/diego/reporters/instances_stats_reporter.rb#L144


api/actions/process_stats.go
```diff
@@ -28,33 +28,34 @@ const (
...
-		Index     int
-		State     string `default:"DOWN"`
-		Usage     Usage
-		MemQuota  *int64
-		DiskQuota *int64
-	}
+type PodStatsRecord struct {
+	ProcessGUID string
+	ProcessType string
+	// TODO: remove, this is the "instance_id" tag from the gauge envelope
```
danail-branekov, line 45: I guess we do not need those TODOs

api/actions/process_stats.go
```diff
@@ -28,33 +28,34 @@ const (
...
+	Index int
+	// TODO: remove, this is the instance state from the process record
```
danail-branekov, line 47: TODO not needed

danail-branekov (commented):
Looks good, apart from a couple of TODOs you have probably added unintentionally

danail-branekov (approved):

Issue #1949, opened 2022-11-28
[Spike]: CFBindingController should not assume any concrete AppWorkload implementation


georgethebeatle:
### Background

Currently the CFServiceBinding controller is reconciling `cfservicebindings` to `servicebindings.servicebinding.io` resources with a [`statefulset` workload](https://github.com/cloudfoundry/korifi/blob/d155777c455da0b14ee8d2359e66dcf95c34e94f/controllers/controllers/services/cfservicebinding_controller.go#L213-L221). This leaks app workload specific implementation detais into the binding controller and prevents the development of alternative app workload implementations (e.g. one backed by `deployments`).

Furthermore, the servicebinding.io is causing apps to restart upon binding a service as the statefulset spec changes. This is unexpected behaviour from the CF perspective

Lastly, the servicebinding.io does not seem to be well maintained anymore. It is vendoring an old controller runtime version and this is preventing us from bumping controller runtime ourselves

### Dev Notes:

- Instead of depending on the servicebinding.io controller, let's do the projection ourselves. This way we will fix the problems above and still have the bindings mounted to the app containers in the desired format

### TODO

- [x] Remove servicebinding.io from tests/vendir.yml
- [x] Revert back to using `controllerutil.CreateOrPatch` in AppWorkloadController
- [x] Implement the projection of the mount secrets
- [x] ~Address [TODO](https://github.com/cloudfoundry/korifi/blob/3ed1e89b069ebf03ac1c70915843ffb189447fe6/controllers/controllers/workloads/apps/controller.go#L144) in app controller~ we hace decided not to pass the references to the binding secrets as the env builder needs the binding itself, alongside the instance, to build the VCAP_SERVICES env var
- [x] (Optional) Fill in empty test: https://github.com/cloudfoundry/korifi/blob/3ed1e89b069ebf03ac1c70915843ffb189447fe6/controllers/controllers/workloads/processes/controller_test.go#L401
- [x] Remove servicebinding.io from go.mod
- [x] Check whether we are backwards compatible with 0.14
- [x] Remove all references of servicebinding.io from docs
- [x] Write a new ADR documenting the decision to drop the servicebinding.io controller

danail-branekov:
We are observing statefulset pods being restarted several times while an app is being deleted. We believe that the reason for this bahviour is the way servicebinding.io projects its binding into the statefulset.
[...]

PR #3801, opened 2025-02-13
Refactor cfprocess controller

uzabanov:
#1949

## What is this change about?
- Use a single app workload var instead of deep copying objects, which
  is confusing
- Instead of using controllerutil.CreateOrPatch, implement a similar
  behaviour ourselves
  - By first trying to creaate the new appWorkload, we avoid a potential
    race
  - We know that we are going to need fine control over what fields are
    being mutated on create vs update

- Also add a new TryPatchResource utility that can handle an error
  returned by the modify func

### DIFF
--- a/controllers/controllers/workloads/processes/controller.go
+++ b/controllers/controllers/workloads/processes/controller.go
@@ -29,6 +29,8 @@ import (
 	"code.cloudfoundry.org/korifi/tools"
 	"code.cloudfoundry.org/korifi/tools/k8s"
 
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+
 	"github.com/go-logr/logr"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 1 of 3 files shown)


georgethebeatle (dismissed):

controllers/controllers/workloads/processes/controller.go
```diff
@@ -236,28 +238,44 @@ func (r *Reconciler) createOrPatchAppWorkload(ctx context.Context, cfApp *korifi
...
 			Name:      getDesiredAppWorkloadName(cfApp, cfProcess),
+			Namespace: cfProcess.Namespace,
 		},
 	}
 
-	var desiredAppWorkload *korifiv1alpha1.AppWorkload
-	desiredAppWorkload, err = r.generateAppWorkload(actualAppWorkload, cfApp, cfProcess, cfBuild, appPorts, envVars)
-	if err != nil {
-		log.Info("error when initializing AppWorkload", "reason", err)
+	if err = r.modifyAppWorkload(appWorkload, cfApp, cfProcess, cfBuild, appPorts, envVars); err != nil {
```
danail-branekov, line 248: `modifyAppWorkload` function name sounds a bit weird to me. What this function does is that it applies desired state onto the app workload, which is what reconciliation in general is all about. How about `reconcileAppWorkload`? Or maybe `applyToWorkload`?

danail-branekov (approved):

PR #3816, opened 2025-02-17
Added GetServiceBinding to the broker client

ddraganovv:
[Issue](https://github.com/cloudfoundry/korifi/issues/3751)

## What is this change about?
This is 2 of 2 PRs related to the above issue. With this PR there is a new handler implemented to the service binding endpoints - "/v3/service_credential_bindings/{guid}/parameters"

No

### DIFF
--- a/api/handlers/service_binding_test.go
+++ b/api/handlers/service_binding_test.go
@@ -397,6 +397,48 @@ var _ = Describe("ServiceBinding", func() {
 		})
 	})
 
+	Describe("GET /v3/service_credential_bindings/{guid}/parameters", func() {
+		BeforeEach(func() {
+			requestMethod = http.MethodGet
+			requestBody = ""
+			requestPath = "/v3/service_credential_bindings/service-binding-guid/parameters"
+
+			serviceBindingRepo.GetServiceBindingParametersReturns(map[string]any{
+				"foo": "val1",
+				"bar": "val2",
+			}, nil)
+		})
+
+		It("returns the service binding parameters", func() {
+			Expect(rr).To(HaveHTTPStatus(http.StatusOK))
+			Expect(rr).To(HaveHTTPHeaderWithValue("Content-Type", "application/json"))
+			Expect(rr).To(HaveHTTPBody(SatisfyAll(
+				MatchJSONPath("$.foo", "val1"),
+				MatchJSONPath("$.bar", "val2"),
+			)))
+		})
+
+		When("the service bindding repo returns an error", func() {
+			BeforeEach(func() {
+				serviceBindingRepo.GetServiceBindingParametersReturns(map[string]any{}, errors.New("get-service-binding-error"))
+			})
+
+			It("returns an error", func() {
+				expectUnknownError()
+			})
+		})
+
+		When("the user is not authorized", func() {
+			BeforeEach(func() {
+				serviceBindingRepo.GetServiceBindingParametersReturns(map[string]any{}, apierrors.NewForbiddenError(nil, "CFServiceBinding"))
+			})
+
+			It("returns 404 NotFound", func() {
+				expectNotFoundError("CFServiceBinding")
+			})
+		})
+	})
+
 	Describe("DELETE /v3/service_credential_bindings/:guid", func() {
 		BeforeEach(func() {
 			requestMethod = "DELETE"
--- a/api/repositories/broker_parameters_client.go
+++ b/api/repositories/broker_parameters_client.go
@@ -1,0 +1,53 @@
+package repositories
+
+import (
+	"context"
+	"fmt"
+
+	korifiv1alpha1 "code.cloudfoundry.org/korifi/controllers/api/v1alpha1"
+	"code.cloudfoundry.org/korifi/controllers/controllers/services/osbapi"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type ParamsResult struct {
+	Parameters map[string]any
+}
+
+type ServiceBrokerClient struct {
+	clientFactory osbapi.BrokerClientFactory
+	k8sClient     client.Client
+	rootNamespace string
+}
+
+func NewServiceBrokerClient(clientFactory osbapi.BrokerClientFactory, k8sClient client.Client, rootNamespace string) *ServiceBrokerClient {
+	return &ServiceBrokerClient{
+		clientFactory: clientFactory,
+		k8sClient:     k8sClient,
+		rootNamespace: rootNamespace,
+	}
+}
+
+func (c *ServiceBrokerClient) GetServiceBindingParameters(ctx context.Context, serviceBinding *korifiv1alpha1.CFServiceBinding) (map[string]any, error) {
+	assetsClient := osbapi.NewAssets(c.k8sClient, c.rootNamespace)
+	sbAssets, err := assetsClient.GetServiceBindingAssets(ctx, serviceBinding)
+	if err != nil {
+		return map[string]any{}, fmt.Errorf("faild to get service binding assets: %w", err)
+	}
+
+	osbapiClient, err := c.clientFactory.CreateClient(ctx, sbAssets.ServiceBroker)
+	if err != nil {
+		return map[string]any{}, fmt.Errorf("faild to create osbapi client: %w", err)
+	}
+
+	payload := osbapi.BindPayload{
+		BindingID:  serviceBinding.Name,
+		InstanceID: serviceBinding.Spec.Service.Name,
+	}
+
+	binding, err := osbapiClient.GetServiceBinding(ctx, payload)
+	if err != nil {
+		return map[string]any{}, fmt.Errorf("faild to create osbapi client: %w", err)
+	}
+
+	return binding.Parameters, nil
+}
--- a/api/repositories/service_binding_repository.go
+++ b/api/repositories/service_binding_repository.go
@@ -35,21 +35,28 @@ const (
 	ServiceBindingResourceType            = "Service Binding"
 )
 
+type ParametersClient interface {
+	GetServiceBindingParameters(ctx context.Context, serviceBinding *korifiv1alpha1.CFServiceBinding) (map[string]any, error)
+}
+
 type ServiceBindingRepo struct {
 	userClientFactory       authorization.UserClientFactory
 	namespaceRetriever      NamespaceRetriever
 	bindingConditionAwaiter Awaiter[*korifiv1alpha1.CFServiceBinding]
+	paramsClient            ParametersClient
 }
 
 func NewServiceBindingRepo(
 	namespaceRetriever NamespaceRetriever,
 	userClientFactory authorization.UserClientFactory,
 	bindingConditionAwaiter Awaiter[*korifiv1alpha1.CFServiceBinding],
+	paramsClient ParametersClient,
 ) *ServiceBindingRepo {
 	return &ServiceBindingRepo{
 		userClientFactory:       userClientFactory,
 		namespaceRetriever:      namespaceRetriever,
 		bindingConditionAwaiter: bindingConditionAwaiter,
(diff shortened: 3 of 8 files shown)


controllers/controllers/services/osbapi/assetsfactory.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+//counterfeiter:generate -o fake -fake-name AssetsClient code.cloudfoundry.org/korifi/controllers/controllers/services/osbapi.AssetsClient
+type AssetsClient interface {
+	GetServiceInstanceAssets(context.Context, *korifiv1alpha1.CFServiceInstance) (ServiceInstanceAssets, error)
+	GetServiceBindingAssets(context.Context, *korifiv1alpha1.CFServiceBinding) (ServiceBindingAssets, error)
+}
+
+//counterfeiter:generate -o fake -fake-name AssetsClientFactory code.cloudfoundry.org/korifi/controllers/controllers/services/osbapi.AssetsClientFactory
+type AssetsClientFactory interface {
```
danail-branekov, line 17: I think that this abstraction (the `AssetsClient` and its factory) are a bit of an overkill:
1) `CreateAssetsClient` literally does the same thing as the `Assets` [constructor](https://github.com/cloudfoundry/korifi/blob/d08d345d393945ba6b499c78442c53a492020f16/controllers/controllers/services/osbapi/assets.go#L17)
2) The only reason to use that factory rather than creating the `Assets` object via its constructor is to fake a bunch of k8s objects (plans, offerings, brokers, etc.). While this generally saves substantial setup code, this abstracts the test away from k8s - that is something we try to avoid, especially in repositories test as repositories are the API "gateway" to the k8s world. That is one of the reason we use envtest integration tests.

TL;DR, I would suggest 
a) removing the assets factory and use `Assets` constructor in the repository
b) setup binding dependencies (CFServiceBroker, CFServiceOffering, etc) in the setup of the binding repository test `Describe("GetServiceBindingParameters`

controllers/controllers/services/osbapi/client.go
```diff
@@ -199,15 +199,15 @@ func (c *Client) Bind(ctx context.Context, payload BindPayload) (BindResponse, e
 			payload.BindRequest,
 		)
 	if err != nil {
-		return BindResponse{}, fmt.Errorf("bind request failed: %w", err)
+		return BindResponse{}, fmt.Errorf("fetch service binding failed: %w", err)
```
danail-branekov, line 202: I don't think that changing the error message here makes sense

controllers/controllers/services/osbapi/client.go
```diff
@@ -199,15 +199,15 @@ func (c *Client) Bind(ctx context.Context, payload BindPayload) (BindResponse, e
...
 	}
 
 	if statusCode == http.StatusBadRequest || statusCode == http.StatusConflict || statusCode == http.StatusUnprocessableEntity {
 		return BindResponse{}, UnrecoverableError{Status: statusCode}
 	}
 
 	if statusCode >= 300 {
-		return BindResponse{}, fmt.Errorf("binding request failed with code: %d", statusCode)
+		return BindResponse{}, fmt.Errorf("fetch service binding request failed with code: %d", statusCode)
```
danail-branekov, line 210: I do not think that changing the error message here makes sense as well

api/repositories/broker_parameters_client.go
```diff
@@ -0,0 +1,55 @@
...
+
+func NewBrokerParamsClient(clientFactory osbapi.BrokerClientFactory, k8sClient client.Client, rootNamespace string) *BrokerParamsClient {
+	return &BrokerParamsClient{
+		clientFactory: clientFactory,
+		k8sClient:     k8sClient,
+		rootNamespace: rootNamespace,
+	}
+}
+
+func (c *BrokerParamsClient) GetServiceBindingParameters(ctx context.Context, serviceBinding *korifiv1alpha1.CFServiceBinding) (ParamsResult, error) {
```
danail-branekov, line 30: Could the method simply return `map[string]any, error`? I.e., do we need to wrap the map into `ParamsResult`?

api/repositories/service_binding_repository_test.go
```diff
@@ -54,7 +58,9 @@ var _ = Describe("ServiceBindingRepo", func() {
 			userClientFactory.WithWrappingFunc(func(client client.WithWatch) client.WithWatch {
 				return authorization.NewSpaceFilteringClient(client, k8sClient, nsPerms)
 			}),
-			conditionAwaiter)
+			conditionAwaiter,
+			nil,
```
danail-branekov, line 62: Instead of using `nil` here, I would suggest wire the parameters client, basically what you do [here](https://github.com/cloudfoundry/korifi/pull/3816/files#diff-cdff009460ea405b57b56faac218be31449c0dc56f8d004a1a25030688811a12R1186-R1201). Thus you would not need to reinitialise the repository in the nested `Describe`

api/repositories/service_binding_repository_test.go
```diff
@@ -1122,7 +1128,7 @@ var _ = Describe("ServiceBindingRepo", func() {
 			serviceBinding, getErr = repo.GetServiceBinding(ctx, authInfo, searchGUID)
 		})
 
-		It("returns a forbidden error as no user bindings are in place", func() {
+		It("returns a forbidden error", func() {
```
danail-branekov, line 1131: I think that this is unrelated

api/repositories/service_binding_repository_test.go
```diff
@@ -1149,6 +1155,174 @@ var _ = Describe("ServiceBindingRepo", func() {
...
+
+			serviceBroker = &korifiv1alpha1.CFServiceBroker{
+				ObjectMeta: metav1.ObjectMeta{
+					Namespace: rootNamespace,
+					Name:      serviceBrokerGUID,
+				},
+				Spec: korifiv1alpha1.CFServiceBrokerSpec{
+					ServiceBroker: services.ServiceBroker{
+						Name: uuid.NewString(),
+						// URL:  brokerServer.URL(),
```
danail-branekov, line 1222: I don't think this comment is needed

api/repositories/service_binding_repository_test.go
```diff
@@ -1149,6 +1155,174 @@ var _ = Describe("ServiceBindingRepo", func() {
 		})
 	})
 
+	Describe("GetServiceBindingParameters", func() {
+		var (
+			serviceBindingGUID   string
+			searchGUID           string
```
danail-branekov, line 1161: I think you could be using the `serviceBindingGUID` instead of having to have a `searchGUID` here. In the context below where you set the `searchGUID`, you could just set `serviceBindingGUID` to `i-do-not-exist` - I think it should have the same effect.

api/repositories/service_binding_repository_test.go
```diff
@@ -1149,6 +1155,174 @@ var _ = Describe("ServiceBindingRepo", func() {
...
+			serviceBindingParams repositories.ServiceBindingParametersRecord
+			getErr               error
+			serviceInstance      *korifiv1alpha1.CFServiceInstance
```
danail-branekov, line 1164: Those four variables (`serviceInstance`, `servicePlan`, `serviceBroker` and `offering`) are only used locally in `BeforeEach`. Therefore they do not need to be declared here

api/repositories/service_binding_repository_test.go
```diff
@@ -1149,6 +1155,174 @@ var _ = Describe("ServiceBindingRepo", func() {
...
+			repo = repositories.NewServiceBindingRepo(
+				namespaceRetriever,
+				userClientFactory.WithWrappingFunc(func(client client.WithWatch) client.WithWatch {
+					return authorization.NewSpaceFilteringClient(client, k8sClient, nsPerms)
+				}),
+				conditionAwaiter,
+				paramsClient,
+			)
+
+			credentialsSecret := &corev1.Secret{
```
danail-branekov, line 1203: the credentials secret is used by the broker client factory to create a client. provided that this test uses a mocked client factory, creating this broker credentials secret does not seem to be needed

api/repositories/broker_parameters_client.go
```diff
@@ -0,0 +1,55 @@
...
+	korifiv1alpha1 "code.cloudfoundry.org/korifi/controllers/api/v1alpha1"
+	"code.cloudfoundry.org/korifi/controllers/controllers/services/osbapi"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type ParamsResult struct {
+	Parameters map[string]any
+}
+
+type BrokerParamsClient struct {
```
danail-branekov, line 16: I would personally call this struct `ServiceBrokerClient` as I can see how this utility can grow to support more endpoints that need to talk to a service broker. But up to you

api/handlers/service_binding_test.go
```diff
@@ -397,6 +397,50 @@ var _ = Describe("ServiceBinding", func() {
...
+					"bar": "val2",
+				},
+			}, nil)
+		})
+
+		It("returns the service binding parameters", func() {
+			Expect(rr).To(HaveHTTPStatus(http.StatusOK))
+			Expect(rr).To(HaveHTTPHeaderWithValue("Content-Type", "application/json"))
+			Expect(rr).To(HaveHTTPBody(SatisfyAll(
+				MatchJSONPath("$.parameters.foo", "val1"),
```
georgethebeatle, line 418: According to the [docs](https://v3-apidocs.cloudfoundry.org/version/3.159.0/index.html#get-parameters-for-a-service-credential-binding) there is no `parameters` key in the json reponse of `/v3/service_credential_bindings/[guid]/parameters`. Instead the params json is presented as-is to the user. The cloud controller implementation does not even define a presenter for this particular endpoint.

api/presenter/service_binding.go
```diff
@@ -27,6 +27,10 @@ type ServiceBindingResponse struct {
 	Metadata      Metadata                            `json:"metadata"`
 }
 
+type ServiceBindingParametersResponse struct {
+	Parameters map[string]any `json:"parameters"`
```
georgethebeatle, line 31: Again, there should be no `parameters` key here, but rather the json object shoud be directly embedded in the request. This makes the presenter redundant. The `record.Parameters` can be directly written to as response body. This is what the cloud controller does: https://github.com/cloudfoundry/cloud_controller_ng/blob/9cf8c4581aaab3d361e5dc694fb97b86a315dbfa/app/controllers/v3/service_credential_bindings_controller.rb#L164

api/repositories/service_binding_repository.go
```diff
@@ -69,6 +76,10 @@ type ServiceBindingRecord struct {
 	Ready               bool
 }
 
+type ServiceBindingParametersRecord struct {
```
georgethebeatle, line 79: I would say that this type is redundant. We can just make the `GetServiceBindingParameters` method return `map[sting]any`. This is similar to the presenter case above, as the struct does not really add anything.

api/repositories/service_binding_repository_test.go
```diff
@@ -49,12 +53,30 @@ var _ = Describe("ServiceBindingRepo", func() {
 			korifiv1alpha1.CFServiceBindingList,
 			*korifiv1alpha1.CFServiceBindingList,
 		]{}
+		brokerClient = new(fake.BrokerClient)
+		brokerClient.GetServiceBindingReturns(osbapi.BindingResponse{
```
georgethebeatle, line 57: Mocking the brokerClient to return binding parameters should be only needed in the `GetServiceBindingParameters` describe, so this line would be a better fit in the `BeforeEach` there

georgethebeatle (commented):
Left some inline comments. Most importantly I believe the v3 API contract is violated.

danail-branekov (dismissed):

danail-branekov (approved):

PR #3828, opened 2025-02-26
Making route controller optional enabling customising of route controllers

Dariquest:
## What is this change about?

This change makes cf routing controller optional. A boolean controller config property "DisableRouteController" was introduced for it. It allows to define custom routing controllers in future.

The default value of the boolean config property "DisableRouteController" is false, so everything stays the same, for those using default routing controller.

## Acceptance Steps

make test

### DIFF
--- a/controllers/config/config.go
+++ b/controllers/config/config.go
@@ -49,11 +49,10 @@ type Networking struct {
 }
 
 const (
-	defaultTaskTTL                      = 30 * 24 * time.Hour
-	defaultTimeout                int32 = 60
-	defaultJobTTL                       = 24 * time.Hour
-	defaultBuildCacheMB                 = 2048
-	defaultDisableRouteController       = false
+	defaultTaskTTL            = 30 * 24 * time.Hour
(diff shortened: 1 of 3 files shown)


helm/korifi/values.yaml
```diff
@@ -145,6 +145,8 @@ networking:
   gatewayClass:
 
 experimental:
+  routing:
+    disableRouteController: false
```
danail-branekov, line 149: Could you document the new property under the [experimental section](https://github.com/cloudfoundry/korifi/blob/e43410e8e0762ee3b07e7112c847df1d436b88fd/helm/korifi/values.schema.json#L600-L648) of the values yaml schema? Make sure to also run `make generate` in the root directory to update the [helm values doc](https://github.com/cloudfoundry/korifi/blob/main/README.helm.md)

danail-branekov (dismissed):

danail-branekov (approved):

danail-branekov (approved):

PR #3842, opened 2025-03-04
Remove redundant counterfeiter annotations

georgethebeatle:
No

## What is this change about?
Remove unneeded annotations


danail-branekov:
Shouldn't we also delete related fakes as well?

danail-branekov (approved):

PR #3853, opened 2025-03-12
Updated org patch endpoint to support org name

ddraganovv:
[Issue](https://github.com/cloudfoundry/korifi/issues/3852)

## What is this change about?
This change enables support for updating the organization name via the `PATCH /v3/organizations/:guid `API endpoint.

No

### DIFF
--- a/api/handlers/org_test.go
+++ b/api/handlers/org_test.go
@@ -208,9 +208,10 @@ var _ = Describe("Org", func() {
 				Name: "test-org",
 			}, nil)
 
-			orgRepo.PatchOrgMetadataReturns(repositories.OrgRecord{
+			newOrgName := tools.PtrTo("new-org-name")
+			orgRepo.PatchOrgReturns(repositories.OrgRecord{
 				GUID: "org-guid",
-				Name: "test-org",
+				Name: *newOrgName,
 				Labels: map[string]string{
 					"env":                           "production",
 					"foo.example.com/my-identifier": "aruba",
@@ -224,6 +225,7 @@ var _ = Describe("Org", func() {
 			requestBody = "the-json-body"
 
 			requestValidator.DecodeAndValidateJSONPayloadStub = decodeAndValidatePayloadStub(&payloads.OrgPatch{
+				Name: newOrgName,
 				Metadata: payloads.MetadataPatch{
 					Annotations: map[string]*string{
 						"hello":                       tools.PtrTo("there"),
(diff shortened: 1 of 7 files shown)


api/handlers/org_test.go
```diff
@@ -194,7 +194,10 @@ var _ = Describe("Org", func() {
 	})
 
 	Describe("Updating Orgs", func() {
-		var requestBody string
+		var (
+			requestBody string
+			newOrgName  *string
```
danail-branekov, line 199: `newOrgName` is only used in the `BeforeEach` below, therefore it could be a local variable there

api/repositories/org_repository_test.go
```diff
@@ -471,32 +471,37 @@ var _ = Describe("OrgRepository", func() {
...
+			orgNewName                    *string
 			cfOrg                         *korifiv1alpha1.CFOrg
 			patchErr                      error
 			orgRecord                     repositories.OrgRecord
 			labelsPatch, annotationsPatch map[string]*string
 		)
 
 		BeforeEach(func() {
-			cfOrg = createOrgWithCleanup(ctx, prefixedGUID("org-name"))
+			displayName = prefixedGUID("org-name")
```
danail-branekov, line 486: while being here, could you use `uuid.NewString()` instead of `prefixedGUID`?

api/repositories/org_repository_test.go
```diff
@@ -471,32 +471,37 @@ var _ = Describe("OrgRepository", func() {
...
+			cfOrg = createOrgWithCleanup(ctx, displayName)
 			orgGUID = cfOrg.Name
 			labelsPatch = nil
 			annotationsPatch = nil
+			orgNewName = tools.PtrTo("new-org-name")
```
danail-branekov, line 491: I wonder whether using a hardcoded new name could cause name collisions when running tests in parallel. Would recommend `uuid.NewString()` instead of a hardcoded value

danail-branekov (commented):
Minor nitpicks on tests, looks good otherwise

danail-branekov (approved):

api/repositories/org_repository_test.go
```diff
@@ -553,6 +556,7 @@ var _ = Describe("OrgRepository", func() {
 							"key-two": "value-two",
 						},
 					))
+					Expect(updatedCFOrg.Spec.DisplayName).To(Equal(*orgNewName))
```
georgethebeatle, line 559: We need to also check that the org record name is updated (I am referring to the `It` on line 526, the UI is not letting me comment on that line)

georgethebeatle (commented):
Left a minor comment inline

georgethebeatle (approved):

PR #3858, opened 2025-03-17
Introduce CFSecurityGroups and the create endpoint 

klapkov:
## What is this change about?
 This is an introduction of the CF Security Groups to Korifi. This PR in particular defines the CFSecurityGroup CR and adds the POST v3/securiy_groups endpoint for creating them.


gogolok:
Is there already an idea how CFSecurityGroups gets 'executed'? Using Kubernetes [Resource Quotas](https://kubernetes.io/docs/concepts/policy/resource-quotas/)?

danail-branekov:
> Is there already an idea how CFSecurityGroups gets 'executed'/enforced? Using Kubernetes [Resource Quotas](https://kubernetes.io/docs/concepts/policy/resource-quotas/)?

Network policies. @klapkov right?

klapkov:
@gogolok Sorry for the late response, if you are interested on what is the current proposal, here is the RFC: https://docs.google.com/document/d/15DpnbkKnno8qYSbMKKQofURtKMUXWgyCbGJr7xXMPXY/edit?tab=t.0#heading=h.4d34og8

danail-branekov:
> if you are interested on what is the current proposal, here is the RFC: [docs.google.com/document/d/15DpnbkKnno8qYSbMKKQofURtKMUXWgyCbGJr7xXMPXY/edit?tab=t.0#heading=h.4d34og8](https://docs.google.com/document/d/15DpnbkKnno8qYSbMKKQofURtKMUXWgyCbGJr7xXMPXY/edit?tab=t.0#heading=h.4d34og8)

@klapkov could you allow commenting on the proposal doc?

api/presenter/security_group_test.go
```diff
@@ -0,0 +1,171 @@
...
+            }
+        }`))
+	})
+
+	When("UpdatedAt is nil", func() {
+		BeforeEach(func() {
+			record.UpdatedAt = nil
+		})
+
+		It("returns an empty string for updated_at", func() {
```
danail-branekov, line 98: Instead of verifying the whole json document, would it not be easier (and more readable) to assert that the "updated_at" field in the result is empty (via `MatchJSONPath` [example](https://github.com/cloudfoundry/korifi/blob/dd74aa0cdc289a6eb108cc62af6856954fe06015/api/presenter/service_binding_test.go#L117))

Same goes for the context when there are no rules and no spaces

api/repositories/security_group_repository.go
```diff
@@ -0,0 +1,101 @@
...
+) *SecurityGroupRepo {
+	return &SecurityGroupRepo{
+		userClientFactory: userClientFactory,
+		rootNamespace:     rootNamespace,
+	}
+}
+
+type CreateSecurityGroupMessage struct {
+	DisplayName     string
+	Rules           []korifiv1alpha1.SecurityGroupRule
```
danail-branekov, line 34: Again, we should not be using korifiv1alpha1 in the message or record types (see above) :(

They should be modelled in the `repository` package and the `Create` method should convert korifiv1lpha1 types to record ones ones

api/repositories/security_group_repository_test.go
```diff
@@ -0,0 +1,78 @@
...
+		It("errors with forbidden for users with no permissions", func() {
+			Expect(createErr).To(matchers.WrapErrorAssignableToTypeOf(apierrors.ForbiddenError{}))
+		})
+
+		When("the user is a CF admin", func() {
+			BeforeEach(func() {
+				createRoleBinding(ctx, userName, adminRole.Name, rootNamespace)
+			})
+
+			It("creates a CFSecurityGroup successfully", func() {
```
danail-branekov, line 61: We also want to check that the `Create` call also created expected `CFSecurityGroup` object in kubernetes, maybe in another `It`

api/repositories/security_group_repository.go
```diff
@@ -0,0 +1,101 @@
...
+		},
+		Spec: korifiv1alpha1.CFSecurityGroupSpec{
+			DisplayName:     message.DisplayName,
+			Rules:           message.Rules,
+			Spaces:          message.Spaces,
+			GloballyEnabled: message.GloballyEnabled,
+		},
+	}
+
+	if err = userClient.Create(ctx, cfSecurityGroup); err != nil {
```
danail-branekov, line 70: creating the object may fail with validation webhook error which we need to translate to cfapi-friendly error - here is an [example](https://github.com/cloudfoundry/korifi/blob/dd74aa0cdc289a6eb108cc62af6856954fe06015/api/repositories/app_repository.go#L272-L281)

controllers/webhooks/networking/security_groups/validator.go
```diff
@@ -0,0 +1,78 @@
...
+//+kubebuilder:webhook:path=/validate-korifi-cloudfoundry-org-v1alpha1-cfsecuritygroup,mutating=false,failurePolicy=fail,sideEffects=None,groups=korifi.cloudfoundry.org,resources=cfsecuritygroups,verbs=create;update;delete,versions=v1alpha1,name=vcfsecuritygroup.korifi.cloudfoundry.org,admissionReviewVersions={v1,v1beta1}
+
+type Validator struct {
+	duplicateValidator webhooks.NameValidator
+	rootNamespace      string
+}
+
+var _ webhook.CustomValidator = &Validator{}
+
+func NewValidator(duplicateValidator webhooks.NameValidator, rootNamespace string) *Validator {
```
danail-branekov, line 30: We should implement destination and port [validation](https://github.com/cloudfoundry/korifi/pull/3858/files#diff-3ee3f212d5c30f2623d51977a5a65a43e557442d47abd133326bb5fbf7b7424dR59) here as well (on create and update) in order to guarantee that a kubectl user is not bypassing API request validation

Then if we had a webhook that validates create and update, then the API could simply propagate the webhook error to the user, i.e. it does not need to implement the very same checks in the payload validation

We should look into how such errors look form the cf cli point of view though, i.e. are they reasonably presented to the user

api/handlers/security_group.go
```diff
@@ -0,0 +1,86 @@
...
+func (h *SecurityGroup) create(r *http.Request) (*routing.Response, error) {
+	authInfo, _ := authorization.InfoFromContext(r.Context())
+	logger := logr.FromContextOrDiscard(r.Context()).WithName("handlers.security-group.create")
+
+	payload := new(payloads.SecurityGroupCreate)
+	if err := h.requestValidator.DecodeAndValidateJSONPayload(r, payload); err != nil {
+		return nil, apierrors.LogAndReturn(logger, err, "failed to decode payload")
+	}
+
+	if len(payload.Relationships.RunningSpaces.Data) != 0 || len(payload.Relationships.StagingSpaces.Data) != 0 {
```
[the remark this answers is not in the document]
klapkov: @danail-branekov If we want to mimic the classic CF here, we need to return an error if the space does not exist. Leaving the controller to handle this would work, but how would the API return the error ?

api/presenter/security_group.go
```diff
@@ -0,0 +1,63 @@
...
+		},
+		Links: SecurityGroupLinks{
+			Self: Link{
+				HRef: buildURL(baseURL).appendPath(securityGroupBase, securityGroupRecord.GUID).build(),
+			},
+		},
+	}
+}
+
+func toManyRelationshipData(guids []string) []payloads.RelationshipData {
```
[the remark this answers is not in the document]
klapkov: @danail-branekov I kept it so we can return [] if the spaces are empty, not sure if there is a better way

api/handlers/security_group_test.go
```diff
@@ -0,0 +1,178 @@
...
+				MatchJSONPath("$.name", "test-security-group"),
+				MatchJSONPath("$.rules[0].protocol", "tcp"),
+				MatchJSONPath("$.rules[0].ports", "80"),
+				MatchJSONPath("$.rules[0].destination", "192.168.1.1"),
+				MatchJSONPath("$.relationships.running_spaces.data[0].guid", "space1"),
+				MatchJSONPath("$.relationships.staging_spaces.data[0].guid", "space2"),
+			)))
+		})
+
+		When("a space in the security group does not exist", func() {
```
danail-branekov, line 148: We should have another context on this level to test that the handler returns an error when listing spaces fails. The space repo could be stubbed to return a random error and the test should expect an unknown error. No need to check for various error types, as the handler just returns the error from the repository

api/presenter/security_group.go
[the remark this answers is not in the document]
danail-branekov: You could wrap `slices.Collect` with [`IfNil`](https://github.com/cloudfoundry/korifi/blob/9f78ad3f12be137111b63940ca79eada8c091608/tools/compare.go#L29), i.e. 

```
return tools.IfNil(slices.Collect(...), []string{})
```

up to you really

api/presenter/security_group_test.go
```diff
@@ -0,0 +1,126 @@
...
+	})
+
+	When("there are no rules and no spaces", func() {
+		BeforeEach(func() {
+			record.Rules = []repositories.SecurityGroupRule{}
+			record.RunningSpaces = []string{}
+			record.StagingSpaces = []string{}
+		})
+
+		It("returns empty arrays for rules and relationships", func() {
```
danail-branekov, line 99: The test could be simplified a bit - instead of checking for the complete response, it could use [`MatchJSONPath`](https://github.com/cloudfoundry/korifi/blob/9f78ad3f12be137111b63940ca79eada8c091608/tests/matchers/jsonpath.go#L19) to verify that running spaces and staging spaces are empty arrays. 

The way you did it is fine as well, up to you whether you change the test or not

controllers/webhooks/networking/security_groups/validator_test.go
```diff
@@ -0,0 +1,285 @@
...
+				duplicateValidator.ValidateDeleteReturns(errors.New("foo"))
+			})
+
+			It("disallows the request", func() {
+				Expect(retErr).To(MatchError("foo"))
+			})
+		})
+	})
+
+	Describe("Validating the security group rules", func() {
```
danail-branekov, line 159: We should have a test to verify that validation is performed on `Update` as well. It does not have to repeat the complete validation test, maybe just one specific case in order to verify that invalid updates are rejected

danail-branekov (commented):
There are a couple of tests that need to be backfilled, some nitpicks as well. Looks good overall

danail-branekov:
I just saw that there is no e2e test for the new endpoint - we would like to have such one as well.

klapkov:
@danail-branekov Since we do not have the DELETE endpoint implemented for security groups, we cannot clean up the create group during the e2e test. Is that okay ?

danail-branekov:
> @danail-branekov Since we do not have the DELETE endpoint implemented for security groups, we cannot clean up the group during the e2e test. Is that okay ?
> 

Yes, e2e tests are [deleting the test org](https://github.com/cloudfoundry/korifi/blob/d4c4ac63f29587b80382a9a1398628410151d787/tests/e2e/e2e_suite_test.go#L398) on suite tear down.


> One more thing, since we do not have a controller yet, where should I define the rules for the korifi-controllers-manager-role - like [domains](https://github.com/cloudfoundry/korifi/blob/main/controllers/controllers/networking/domains/controller.go?rgh-link-date=2025-04-01T11%3A25%3A57Z#L55) Without them, the e2e test fails.

I would suggest to create a new package for the security groups [next to `domains`](https://github.com/cloudfoundry/korifi/tree/main/controllers/controllers/networking) and a `package.go` with the kubebuilder annotations.

danail-branekov (dismissed):

controllers/webhooks/networking/security_groups/validator_test.go
```diff
@@ -0,0 +1,320 @@
...
+				duplicateValidator.ValidateDeleteReturns(errors.New("foo"))
+			})
+
+			It("disallows the request", func() {
+				Expect(retErr).To(MatchError("foo"))
+			})
+		})
+	})
+
+	Describe("Validating the security group rules when creating", func() {
```
danail-branekov, line 159: Minor recommendation on the test structure (feel free to ignore it) - I would inline this group of tests under `Describe("ValidateCreate"` as this is the flow they exercise. I would also get rid of `When("the rule ports is invalid"` and `When("the rule destination is invalid"` as they are not altering the default setup with additional context. If you think those improve readability, I would use `Describe` instead of `When`

Same goes to `Describe("Validating the security group rules when updating"`

danail-branekov (approved):

danail-branekov:
@klapkov e2e tests are failing as I need to enable the new experiemental feature on concourse. Will rerun the tests once I do that

PR #3911, opened 2025-04-02
Enable deleting unmapped routes for a space

klapkov:

https://github.com/cloudfoundry/korifi/issues/3909
## What is this change about?

Introducing the DELETE v3/spaces/:guid/routes?unmapped=true endpoint. This one is kind of strange, since it aways needs to be called with the unmapped param. API Doc: https://v3-apidocs.cloudfoundry.org/version/3.190.0/index.html#delete-unmapped-routes-for-a-space It deletes all the unmapped routes for a certain space

No


api/payloads/space.go
```diff
@@ -84,3 +85,33 @@ func (l *SpaceList) DecodeFromURLValues(values url.Values) error {
...
+	Unmapped string `json:"unmapped"`
+}
+
+func (d *SpaceDeleteRoutes) SupportedKeys() []string {
+	return []string{"unmapped"}
+}
+
+func (d SpaceDeleteRoutes) Validate() error {
+	return validation.ValidateStruct(&d,
+		validation.Field(&d.Unmapped, validation.Required, validation.By(func(value any) error {
```
georgethebeatle, line 99: Can we use `validation.OneOf` here, `true` being the only allowed value?

api/repositories/route_repository_test.go
```diff
@@ -429,6 +429,20 @@ var _ = Describe("RouteRepository", func() {
 					Expect(routeRecords).To(BeEmpty())
 				})
 			})
+
+			When("is-unmapped filter is provided", func() {
+				BeforeEach(func() {
+					message = ListRoutesMessage{IsUnmapped: true}
+				})
+
+				It("returns a list of routeRecords that are unmapped to a app", func() {
```
georgethebeatle, line 438: How about `returns a list of routeRecords that are not mapped to an app`?

api/handlers/space.go
```diff
@@ -148,6 +151,36 @@ func (h *Space) get(r *http.Request) (*routing.Response, error) {
...
+	}
+
+	for _, route := range routes {
+		err = h.routeRepo.DeleteRoute(r.Context(), authInfo, repositories.DeleteRouteMessage{SpaceGUID: spaceGUID, GUID: route.GUID})
+		if err != nil {
+			return nil, apierrors.LogAndReturn(logger, err, "Failed to delete unmapped route", route.GUID, "for space", "SpaceGUID", spaceGUID)
+		}
+	}
+
+	return routing.NewResponse(http.StatusAccepted), nil
```
georgethebeatle, line 181: We should somehow return a job url in a Location header here

api/handlers/space.go
```diff
@@ -148,6 +151,36 @@ func (h *Space) get(r *http.Request) (*routing.Response, error) {
...
+	if err != nil {
+		return nil, apierrors.LogAndReturn(logger, apierrors.ForbiddenAsNotFound(err), "Failed to fetch space", "spaceGUID", spaceGUID)
+	}
+
+	routes, err := h.routeRepo.ListRoutes(r.Context(), authInfo, repositories.ListRoutesMessage{SpaceGUIDs: []string{spaceGUID}, IsUnmapped: true})
+	if err != nil {
+		return nil, apierrors.LogAndReturn(logger, err, "Failed to delete unmapped routes for space", "SpaceGUID", spaceGUID)
+	}
+
+	for _, route := range routes {
```
georgethebeatle, line 174: I am wondering if we should push the deletion to the repository to simplify the handler. It wouldn't change much, don't have a hard opinion...

danail-branekov: Playing the devils advocate, I would suggest not to push route deletion to the space repository because the space repository would have to either depend on the route repo, or know how to handle routes (which is responsibility of the route repo). Handlers typically delegate to repositories one they handle the request parameters, therefore delegating to different repositories kind of make sense to me.

api/payloads/space.go
klapkov: Yes, this is a lot simpler, thanks.

api/handlers/space.go
danail-branekov: A couple of days later, when looking at this PR again, I support @georgethebeatle suggestion to push the logic down to the repository. But instead of trying to alter the `DeleteRoute` method (where its message specifies the route to be deleted explicitly), I would suggest creating a new method - `DeleteUnmappedRoutes(...., spaceGUID)`

api/handlers/space.go
danail-branekov: @georgethebeatle's point is that the response should contain the `Location` header in order requester to start polling for the job status. 

The way Korifi implements those jobs is a synthetic manner where the job ID encodes the job type and additional metadata. For example, the job ID returned here could be `space.delete_unapped_routes~<space-guid`. Then, in the job handler, we need a special handling of this job type to return `in progress` state while the space has unmapped routes. In order to achieve it, we would probably need to enhance the route repository list to allow filtering by unmapped routes.

As this operation by specification is a mixture of list and delete, this job type should have its own special handling in the job handler, similarly to the [apply manifest one](https://github.com/cloudfoundry/korifi/blob/3ec473e5fdf1dc1782842f5ca9062d86b8f4ab55/api/handlers/job.go#L87)

api/payloads/space.go
klapkov: @danail-branekov @georgethebeatle What do you think regarding this ?

danail-branekov: How about 

```
func (d SpaceDeleteRoutes) Validate() error {
	return validation.ValidateStruct(&d,
		validation.Field(&d.Unmapped, validation.Required, validation.StringIn(true, "true").Error("mass delete not supported for mapped routes. Use 'unmapped=true' parameter to delete all unmapped routes")),
	)
}
```

the reason to have the custom `OneOf` validator is to customise the error message in a generic manner. Here we could just not use this generic message and use our own.

danail-branekov: Or even

```
func (d SpaceDeleteRoutes) Validate() error {
	return validation.ValidateStruct(&d,
		validation.Field(&d.Unmapped,
			validation.Required,
			validation.StringIn(true, "true", "false").Error("must be a boolean"),
			validation.StringIn(true, "true").Error("mass delete not supported for mapped routes. Use 'unmapped=true' parameter to delete all unmapped routes"),
		),
	)
}

```

that validation even makes the tests @klapkov wrote happy :)

klapkov: Yep, this should do the job, thanks.

api/handlers/space.go
klapkov: @danail-branekov  Just to see if we are on the same page. The idea is to create a job that will track when the unmapped routes are deleted. During the job, the state of the space should not be ready, once all the routes are deleted, the state is ready. The cf cli will track the job until it is finished. Am I on the right track ?

danail-branekov: > Am I on the right track ?

Almost :)

The job will check whether there are unmapped route in the space. If there are, the job state should be `running`. Once the space no longer contains unmapped routes, the job state should be `completed`

While routes are being deleted, the state of the space is irrelevant.

> The cf cli will track the job until it is finished

I do not know whether the cli supports a command to clear unmapped routes, but a theoretical client could do that

api/repositories/route_repository.go
```diff
@@ -421,6 +455,19 @@ func (r *RouteRepo) PatchRouteMetadata(ctx context.Context, authInfo authorizati
 	return cfRouteToRouteRecord(*route), nil
 }
 
+func (r *RouteRepo) GetState(ctx context.Context, authInfo authorization.Info, spaceGUID string) (ResourceState, error) {
```
danail-branekov, line 458: From my point of view, the logic to check whether the space contains unmapped routes should not be here. Whether a space contains unmapped routes or not is orthogonal to whether the space itself is ready or not.

How about we have a special handling of the `SpaceDeleteUnmappedRoutesType` job type in the job handlers, similar to the [special handling of applying space manifest](https://github.com/cloudfoundry/korifi/blob/b170c35905716ab3bb577122950180f132ad8b84/api/handlers/job.go#L87). I.e.

```
	switch job.Type {
	case syncSpaceJobType:
		// ....
	case SpaceDeleteUnmappedRoutesType:
		// list unmapped routes in that space via `routeRepo.ListRoutes(umapped=true)`. If the slice is empty, return "ready", "unknown" otherwise
	default:
		// ....
	}
```

api/repositories/route_repository_test.go
```diff
@@ -423,6 +423,20 @@ var _ = Describe("RouteRepository", func() {
...
 				})
 			})
+
+			When("is-unmapped filter is provided", func() {
+				BeforeEach(func() {
+					message = ListRoutesMessage{IsUnmapped: true}
+				})
+
+				It("returns a list of routeRecords that are not mapped to an app", func() {
+					Expect(routeRecords).To(HaveLen(2))
```
danail-branekov, line 433: No need to check `HaveLen` here, `ConsistOf` would do that for you :)

api/repositories/route_repository_test.go
```diff
@@ -694,6 +708,84 @@ var _ = Describe("RouteRepository", func() {
...
+			).To(Succeed())
+		})
+
+		JustBeforeEach(func() {
+			deleteErr = routeRepo.DeleteUnmappedRoutes(ctx, authInfo, space.Name)
+		})
+
+		It("deletes only the unmapped route resource", func() {
+			Expect(deleteErr).NotTo(HaveOccurred())
+			Expect(k8sClient.Get(ctx, client.ObjectKey{Namespace: space.Name, Name: cfRoute2.Name}, &korifiv1alpha1.CFRoute{})).To(MatchError(ContainSubstring("not found")))
```
danail-branekov, line 784: I would suggest listing all `CFRoutes` in the `space.Name` and verify that the result only consist of the `cfRoute1` route instead of performing two gets

danail-branekov (commented):
Looks good, I just have a question about the links to the space in the delete unmapped routes job

api/presenter/job.go
```diff
@@ -90,6 +91,14 @@ func ForManifestApplyJob(job Job, baseURL url.URL) JobResponse {
 	return response
 }
 
+func ForSpaceDeleteUnmappedRoutesJob(job Job, state repositories.ResourceState, baseURL url.URL) JobResponse {
+	response := ForJob(job, []JobResponseError{}, state, baseURL)
+	response.Links.Space = &Link{
```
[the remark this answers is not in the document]
klapkov: Yep, here is what we have on cf classic: 
```
{
  "guid": "769699e0-4372-4353-b33f-96d801d4594e",
  "created_at": "2025-04-28T14:01:58Z",
  "updated_at": "2025-04-28T14:01:58Z",
  "operation": "space.delete_unmapped_routes",
  "state": "COMPLETE",
  "errors": [],
  "warnings": [],
  "links": {
    "self": {
      "href": "https://v3/jobs/769699e0-4372-4353-b33f-96d801d4594e"
    },
    "space": {
      "href": "https://v3/spaces/d01c0e3e-c3e6-4d1e-9cf5-1aec595993bd"
    }
  }
}
```

danail-branekov (approved):

Issue #3863, opened 2025-03-18
`PATCH /v3/organizations/:guid` does not support updating `suspended` field


ddraganovv:
**Issue:**
PATCH /v3/organizations/:guid does not support updating the organization `suspended` status

**Description:**
Currently, the `PATCH /v3/organizations/:guid` endpoint only allows updating organization's metadata and name, but it does not support updating the `suspended` status for an organization.
As per the [API documentation](https://v3-apidocs.cloudfoundry.org/version/3.180.0/index.html#update-an-organization), the endpoint should support updating the `suspended` status.

PR #3938, opened 2025-04-16
Accept `suspended` property when PATCHing organization

gogolok:
The endpoint `PATCH /v3/organizations/:guid` can accept the field `suspended` now.  The logic for `suspended` is not implemented.

Part of supporting #3863

GitHub issue #3863

## What is this change about?

It's about accepting the `suspended` field during `PATCH /v3/organizations/:guid`, but NOT actually persisting the change.
This way the API won't error out if someone always specifies the `suspended` field in the PATCH request.

No.

## Acceptance Steps

Try to patch a CF org's suspended field, for example
`cf curl -X PATCH /v3/organizations/d3bff835-3603-4749-9e25-714626e204b5 -d '{"suspended": true }'`

The previous version should error out and the patched version should accept the request (and do nothing).

### DIFF
--- a/api/payloads/org.go
+++ b/api/payloads/org.go
@@ -30,8 +30,9 @@ func (p OrgCreate) ToMessage() repositories.CreateOrgMessage {
 }
 
 type OrgPatch struct {
-	Name     *string       `json:"name"`
-	Metadata MetadataPatch `json:"metadata"`
+	Name      *string       `json:"name"`
+	Suspended *bool         `json:"suspended"`
+	Metadata  MetadataPatch `json:"metadata"`
 }
 
 func (p OrgPatch) Validate() error {
--- a/api/payloads/org_test.go
+++ b/api/payloads/org_test.go
@@ -64,7 +64,8 @@ var _ = Describe("Org", func() {
 
 		BeforeEach(func() {
 			payload = payloads.OrgPatch{
-				Name: tools.PtrTo("new-org-name"),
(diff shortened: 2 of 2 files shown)


danail-branekov:
@gogolok do you plan on implementing the property yourself? If yes, how about inlining it into the subsequent PR?

gogolok:
@danail-branekov I haven't looked into implementing it yet. Adding the field should hopefully not be hard. On the other hand, I don't know (yet) what to do if an org is actually marked as suspended. So I would not implement the actual buisness logic in the near future. Do you have an idea?

danail-branekov:
We are a bit on the fence to introduce a "dummy" support for a query parameter as people often come to us asking us what part of the API spec Korifi supports. When they see an allowed query parameter, they accept that it is really supported.

Of course, there are exceptions - for example we had to stub `page` as cli sends those all over the place and without doing that `cf push` would not work. If you have a critical scenario that requires that stub though, yes, we could totally do it.

As per implementing suspended orgs - yesterday we had a short chat with @georgethebeatle on it. We do not know how to implement it for realz, we thought that maybe Korifi could delete rolebinidngs for non-admin users in the org namespace and its related space namespaces. Another option would be an HTTP middleware that denies requests for suspended orgs. And probably there are more...

gogolok:
I think it's then better to close this PR, thanks for the feedback!

danail-branekov:
Closing as discussed

PR #3993, opened 2025-05-19
[API] Enable listing droplets

klapkov:

https://github.com/cloudfoundry/korifi/issues/3992
## What is this change about?

Implement GET v3/droplets 

no


api/repositories/droplet_repository.go
```diff
@@ -58,14 +59,21 @@ func (r DropletRecord) Relationships() map[string]string {
 }
 
 type ListDropletsMessage struct {
+	GUIDs        []string
 	PackageGUIDs []string
 	AppGUIDs     []string
+	SpaceGUIDs   []string
+}
+
+func (m *ListDropletsMessage) matches(droplet korifiv1alpha1.CFBuild) bool {
```
georgethebeatle, line 68: We are currently pushing filtering to the server side leveraging kubernetes' label selectors. We no longer use the `matches` method for client side filtering.

The way this works is via a indexer webhook that puts various labels on each registered object that we can later lookup by. The right way to filter by guid would be to add a [rule](https://github.com/cloudfoundry/korifi/blob/386e2605057d5185dc580dcb59b7bcb6959d6853/controllers/webhooks/label_indexer/webhook.go#L44-L48) to the indexer webhook that puts the drioplet guid as a label in the CFBuild resource and then to translate message.GUIDs to `WithLabelIn(<guid-label-name>, m.GUIDs)` as it is done for space guids.

api/repositories/droplet_repository.go
```diff
@@ -144,7 +152,7 @@ func (r *DropletRepo) ListDroplets(ctx context.Context, authInfo authorization.I
 		return []DropletRecord{}, apierrors.FromK8sError(err, BuildResourceType)
 	}
 
-	filteredBuilds := itx.FromSlice(buildList.Items)
+	filteredBuilds := itx.FromSlice(buildList.Items).Filter(message.matches)
```
georgethebeatle, line 155: No need to do client side filtering, see comment above

api/repositories/droplet_repository_test.go
```diff
@@ -386,29 +386,46 @@ var _ = Describe("DropletRepository", func() {
...
+				})
+
+				It("translates filter parameters to klient list options", func() {
+					Expect(listErr).NotTo(HaveOccurred())
+					Expect(fakeKlient.ListCallCount()).To(Equal(1))
+					_, _, listOptions := fakeKlient.ListArgsForCall(0)
+					Expect(listOptions).To(ConsistOf(
+						repositories.WithLabelIn(korifiv1alpha1.CFPackageGUIDLabelKey, []string{"p1", "p2"}),
+						repositories.WithLabelIn(korifiv1alpha1.CFAppGUIDLabelKey, []string{"a1", "a2"}),
+						repositories.WithLabelIn(korifiv1alpha1.SpaceGUIDKey, []string{"a1", "a2"}),
```
georgethebeatle, line 427: Add a test that check that message.GUIDs is translated to the right list option. Once you do that you do not need an explicit test by actually creating some dropets and then filtering them by guid, since:
- we trust that kubernetes label selectors work
- we should have tests that the build is labeled with its droplet guid
- we know that repositories.Klient delegates those list options to the kube apiserver

api/repositories/droplet_repository_test.go
```diff
@@ -373,7 +373,7 @@ var _ = Describe("DropletRepository", func() {
 				Expect(dropletRecords).To(HaveLen(2))
 			})
 
-			Describe("filtering", func() {
+			Describe("filtering by package guid", func() {
```
georgethebeatle, line 376: No need to test filtering for every filtering param, see explanation below. Generally one test that filters by just one param is enough as a smoke test that proves that we have wired all the pieces together correctly

georgethebeatle (commented):
Thanks for contributing, left some comments inline

georgethebeatle (approved):

PR #4014, opened 2025-06-05
Introduce Controller for Security Groups and bind, get endpoints

klapkov:
## What is this change about?

With this PR, we introduce the get, and bind endpoint for CFSecurityGroups. We also bring the controller, which will translate the security group rules to network policies and global network policies


klapkov:
One thing I haven't addressed is the kind installation, should we replace contour with calico ? I have also not updated any documentation.

Issue #4025, opened 2025-06-10
Enable tls certificate check when talking to the local log cache loopback in the Korifi API shim


georgethebeatle:
### Background

There is a security warning with high severity in github: https://github.com/cloudfoundry/korifi/security/code-scanning/3

We initially ignored the linter warning, because we are talking to localhost, but maybe it is still a good idea to fix it to leave no room for exploits

### Dev Notes

We recently introduced the capability for users to bring their own self signed certs if they do not wat to depend on the cert manager. Maybe we could introduce some helm value for passing a secret with log cache certs, with the option to generate it with cert manager for simplicity.

PR #4026, opened 2025-06-10
Perform tls validation in log cache client unless explicitly disabled

georgethebeatle:
#4025

## What is this change about?
Do tls validation when talking to local log cache. Do not skip validation unless explicitly configured via helm value

### DIFF
--- a/api/main.go
+++ b/api/main.go
@@ -3,6 +3,7 @@ package main
 import (
 	"context"
 	"crypto/tls"
+	"crypto/x509"
 	"fmt"
 	"log"
 	"net/http"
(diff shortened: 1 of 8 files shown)


api/main.go
```diff
@@ -573,3 +556,43 @@ func wireIdentityProvider(client client.Client, restConfig *rest.Config) authori
...
+			},
+		},
+	}
+	if !cfg.Experimental.ExternalLogCache.Enabled {
+		logCacheURL, err = url.Parse(fmt.Sprintf("https://localhost:%d", cfg.InternalPort))
+		if err != nil {
+			return nil, nil, fmt.Errorf("could not parse local log cache URL: %w", err)
+		}
+
+		tlsPath := os.Getenv("TLSCONFIG")
```
danail-branekov, line 578: Aren't those the certificate bundle for the api server itself? An enternal log cache cert bundle could have a different cert authority

danail-branekov (dismissed):

danail-branekov (approved):

PR #4053, opened 2025-06-23
Implement API PATCH /v3/service_offerings/:guid

ybykov-a9s:
The PR is aimed to solve an existing [issue 3588](https://github.com/cloudfoundry/korifi/issues/3588)

## What is this change about?
It implements API PATCH /v3/service_offerings/:guid

There are no breaking changes in the PR

## Acceptance Steps
Changes are covered by tests in the PR. I would also assume that passing a workload to the endpoint should work.


api/handlers/service_offering.go
```diff
@@ -79,6 +80,30 @@ func (h *ServiceOffering) get(r *http.Request) (*routing.Response, error) {
...
+	serviceOfferingGUID := routing.URLParam(r, "guid")
+
+	var payload payloads.ServiceOfferingUpdate
+	if err := h.requestValidator.DecodeAndValidateJSONPayload(r, &payload); err != nil {
+		return nil, apierrors.LogAndReturn(logger, err, "failed to decode payload")
+	}
+
+	_, err := h.serviceOfferingRepo.GetServiceOffering(r.Context(), authInfo, serviceOfferingGUID)
+	if err != nil {
+		return nil, apierrors.LogAndReturn(logger, apierrors.ForbiddenAsNotFound(err), "Error getting service offering in repository")
```
danail-branekov, line 96: The `handlers\service_offering_test` does not cover the case that whenever the repository returns a forbidden error, it is "masked" to `NotFound` (via the `ForbiddenAsNotFound` utility) because not to leak the knowledge that an object with that ID actually exists.

[Here](https://github.com/cloudfoundry/korifi/blob/9090f8b1d77c3442e0e1f42df2774b831c69dec8/api/handlers/process_test.go#L483-L491) is a similar test form the processes handler

api/repositories/service_offering_repository_test.go
```diff
@@ -448,4 +450,118 @@ var _ = Describe("ServiceOfferingRepo", func() {
...
+			updateErr              error
+			updateMessage          repositories.UpdateServiceOfferingMessage
+		)
+
+		BeforeEach(func() {
+			offeringGUID = uuid.NewString()
+
+			brokerGUID := uuid.NewString()
+			offeringName := uuid.NewString()
+			broker = &korifiv1alpha1.CFServiceBroker{
```
danail-branekov, line 469: Does the test for updating an offering require a broker?

api/repositories/service_offering_repository_test.go
```diff
@@ -448,4 +450,118 @@ var _ = Describe("ServiceOfferingRepo", func() {
...
+
+		When("the user is a CFAdmin", func() {
+			BeforeEach(func() {
+				createRoleBinding(ctx, userName, adminRole.Name, rootNamespace)
+			})
+
+			It("updates the service offering metadata", func() {
+				Expect(updateErr).NotTo(HaveOccurred())
+
+				Expect(updatedServiceOffering.Metadata.Labels).To(HaveKeyWithValue(korifiv1alpha1.RelServiceBrokerGUIDLabel, broker.Name))
```
danail-branekov, line 552: You could combine multiple assertions on the same map via `SatisfyAll`, for example:

```golang
Expect(updatedServiceOffering.Metadata.Labels).To(SatisfyAll(
  HaveKeyWithValue(....),
  HaveKeyWithValue(....),
))
```

Alternatively, you could use `gstruct.MatchKeys`, i.e.

```golang
Expect(updatedServiceOffering.Metadata.Labels).To(MatchKeys(IgnoreExtras, Keys{
  "k1": Equal("v1"),
  "k2": Equal("v2"),
}))
```

api/handlers/service_offering.go
```diff
@@ -125,6 +150,7 @@ func (h *ServiceOffering) UnauthenticatedRoutes() []routing.Route {
 func (h *ServiceOffering) AuthenticatedRoutes() []routing.Route {
 	return []routing.Route{
 		{Method: "GET", Pattern: ServiceOfferingPath, Handler: h.get},
+		{Method: "PATCH", Pattern: ServiceOfferingPath, Handler: h.update},
```
danail-branekov, line 153: As this introduces a new endpoint, we would like to have an basic e2e test to ensure that the endpoint generally works. 

[Here](https://github.com/cloudfoundry/korifi/blob/9090f8b1d77c3442e0e1f42df2774b831c69dec8/tests/e2e/processes_test.go#L177-L198) is a similar test for patching processes metadata

api/repositories/service_offering_repository_test.go
ybykov-a9s: Cool. Updated test.

Issue #3701, opened 2025-01-13
[Explore]: How to implement pagination with the current approach?


c0d1ngm0nk3y:
### Background

Most (if not even all) cf list endpoints (e.g. `/v3/apps`) have the possibility for pagination by using the parameters `page` and `per_page`.

But when calling `List` on `k8s` there is no way to to implement such a pagination.

How should this 

### Acceptance Criteria

Implementation of pagination for an existing endpoint.

### Timebox

No timebox since it has to be solved to be api compatible.

### Dev Notes

Just listing everything and returning the requested page to the user, is not really an option because this would scale very poorly.

c0d1ngm0nk3y:
> [github-project-automation](https://github.com/apps/github-project-automation) bot moved this to 🧊 Icebox in [Korifi - Backlog](https://github.com/orgs/cloudfoundry/projects/35)

@danail-branekov What does that mean?

danail-branekov:
> @danail-branekov What does that mean?

Whenever someone creates a new issue, the automation puts it into the `Icebox` bucket of the [Korifi backlog](https://github.com/orgs/cloudfoundry/projects/35)

georgethebeatle:
In this [spike branch](https://github.com/cloudfoundry/korifi/tree/spike-paging) we have demonstrated how we could model offset paging using the kubernetes cursor paging behind the scenes:

Here is roughly how it could work:
- it is possible to advance the cursor with different count each time
- the continue token can be reused accross client instances
- it is possible to advance the cursor by a desired amount whout transmitting much over the wire. We only read a few kb until we see the continue token, then we close the body reader, thus telling the server to not send any more data
- using the point above we could do random access via first advancing the cursor by the desired offset, then get the continue token and use it with the real client to obtain the next offset

However this approach would not work when combined with sorting, because the kubernetes api [does not support](https://github.com/kubernetes/kubernetes/issues/80602) object ordering. This means that in order to present all resources in order we need to first fetch them all, then sort them on the client side, which is suboptimal.

This finding leaves us with several alternatives:

- Leave paging unimplemented (status quo, client side sorting performed at the korifi api shim)
- Implement paging, and ignore sorting if both paging and sorting are requested
- Implement paging, but if both paging and sorting are requested return a single sorted page with all results
- Serve korifi resources with an extension server that stores them in a relational database, then Implement all list operations with sql queries, but leave other operations unchanged

[...]

PR #4070, opened 2025-07-09
Ordering for droplets, orgs, packages

georgethebeatle:
#3701

## What is this change about?
Most of this change deals with introducing dedicated payloads for
listing resources in the context of other resources. For example:
- there is a payload for listing droplets via /v3/droplets
- there is an alternative payload for listing the droplets in the
  context of an app via /v3/apps/:guid/droplets
- these payloads are a bit different
- both should support ordering and paging

### DIFF
--- a/api/handlers/app.go
+++ b/api/handlers/app.go
@@ -222,7 +222,13 @@ func (h *App) listDroplets(r *http.Request) (*routing.Response, error) {
 		return nil, apierrors.LogAndReturn(logger, apierrors.ForbiddenAsNotFound(err), "Failed to fetch app from Kubernetes", "AppGUID", appGUID)
 	}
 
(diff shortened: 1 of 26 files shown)


danail-branekov (dismissed):

api/handlers/app.go
```diff
@@ -563,15 +564,9 @@ func (h *App) getProcess(r *http.Request) (*routing.Response, error) {
 	appGUID := routing.URLParam(r, "guid")
 	processType := routing.URLParam(r, "type")
 
-	app, err := h.appRepo.GetApp(r.Context(), authInfo, appGUID)
```
danail-branekov, line 566: We have those sort of gets when listing related resources (e.g. routes for an app) to ensure the user is allowed to get the app, and if not, mask the forbidden error as not found. We did that on purpose in order not to leak information whether a resource ID exists to unauthorised users. Not sure whether we have to remove that.

PR #4117, opened 2025-08-05
Issue 4100

marsteg:
https://github.com/cloudfoundry/korifi/issues/4100

## What is this change about?
installing calico, enabling global-default-deny-policy, implement workaround in dev installation to allow traffic

yes, if the global policy is enabled and no other security groups get applied

## Acceptance Steps
- remove "allow_calico" from main function in scripts/deploy-on-kind.sh
- create a new calico installation
- note, that builds via cf push fail
- apply half of the "workaround"-policy:
 kubectl apply -f - <<EOF
apiVersion: projectcalico.org/v3
kind: GlobalNetworkPolicy
metadata:
  name: allow-apps-outbound
spec:
  namespaceSelector: has(korifi.cloudfoundry.org/space-guid)
  types:
  - Egress
  egress:
  - action: Allow
    protocol: TCP
    destination:
      ports:
      - 1:65535
EOF

- deploy dorifi-golang
- note, that cf push works now, however apps are not accessible
- delete the policy:
kubectl delete globalnetworkpolicy allow-apps-outbound

- apply the other half of the workaround of the"workaround"-policy:
  kubectl apply -f - <<EOF
apiVersion: projectcalico.org/v3
kind: GlobalNetworkPolicy
metadata:
  name: allow-apps
spec:
  namespaceSelector: has(korifi.cloudfoundry.org/space-guid)
  types:
  - Ingress
  - Egress
  ingress:
  - action: Allow
    protocol: TCP
    source:
      selector: 'app == "envoy-korifi"'
    destination:
      ports:
      - 8080
EOF

- note, that dorifi-golang is accessible, but cannot reach the google: 
curl dorifi.apps-127-0-0-1.nip.io/outbound -vk
- apply the whole workaround:
  kubectl apply -f - <<EOF
apiVersion: projectcalico.org/v3
kind: GlobalNetworkPolicy
metadata:
  name: allow-apps
spec:
  namespaceSelector: has(korifi.cloudfoundry.org/space-guid)
  types:
  - Ingress
  - Egress
  ingress:
  - action: Allow
    protocol: TCP
    source:
      selector: 'app == "envoy-korifi"'
    destination:
      ports:
      - 8080
  egress:
  - action: Allow
    protocol: TCP
    destination:
      ports:
      - 1:65535
EOF

- note, everything works

@danail-branekov


scripts/install-dependencies.sh
```diff
@@ -105,3 +105,13 @@ if ! kubectl get apiservice v1beta1.metrics.k8s.io >/dev/null 2>&1; then
     kubectl apply -f "$VENDOR_DIR/metrics-server-local"
   fi
 fi
+
+echo "********************"
+echo " Installing Calico"
+echo "********************"
+
+sed -i 's/cidr: 192.168.0.0\/16/cidr: 10.244.0.0\/16/g' "$VENDOR_DIR/calico/custom-resources.yaml"
```
georgethebeatle, line 113: Modifying the vendored resource definition will leave unwanted diff in the vendor dir every time we run `scripts/deploy-on-kind.sh`. When we have to modify depencencies prior to applying we tend to use `kustomize`.

An [example](https://github.com/cloudfoundry/korifi/blob/5e2bd0c14f852173fa35663d0956565da3d119e7/scripts/install-dependencies.sh#L97-L99) of this is the metrics server. We provide a [kustomization](https://github.com/cloudfoundry/korifi/blob/5e2bd0c14f852173fa35663d0956565da3d119e7/tests/dependencies/insecure-metrics-server/kustomization.yaml) file, then we copy the dependency over and do `kubectl apply -k`. We also append the copied file name to the `TEMP_FILES` array which is cleaned up when the script exits, so that we leave no debris behind.

It could make sense to define a common function that does the [boilerplate](https://github.com/cloudfoundry/korifi/blob/5e2bd0c14f852173fa35663d0956565da3d119e7/scripts/install-dependencies.sh#L97-L99) in a generic way to make the script a bit more concise.

georgethebeatle (commented):
Thanks for the PR @marsteg! Looks good, left some comments inline.

helm/korifi/templates/global-deny-apps-policy.yaml
```diff
@@ -0,0 +1,31 @@
+{{- if .Values.experimental.securityGroups.enabled }}
```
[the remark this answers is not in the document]
danail-branekov: @marsteg Could we also have those excludes as a file inder `assets` directory so that we can apply it directly on our CI environment as well later on?

scripts/deploy-on-kind.sh
```diff
@@ -262,6 +262,35 @@ function create_cluster_builder() {
   kubectl wait --for=condition=ready clusterbuilder --all=true --timeout=15m
 }
 
+function allow_calico() {
```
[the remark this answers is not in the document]
marsteg: changed the function name

scripts/install-dependencies.sh
marsteg: i changed the approach to the same as for the metrics server - however, i didn't create an extra function - so far it seems still simple to understand the three lines of code as they're quite basic.

scripts/assets/calico-allow-policy.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: projectcalico.org/v3
+kind: GlobalNetworkPolicy
+metadata:
+  name: allow-apps
+spec:
+  namespaceSelector: has(korifi.cloudfoundry.org/space-guid)
+  types:
+  - Ingress
+  - Egress
+  ingress:
```
danail-branekov, line 10: Security groups are meant to configure apps egress (outbound traffic) but not ingress. Is it necessary to configure ingress here?

helm/korifi/templates/global-deny-apps-policy.yaml
```diff
@@ -0,0 +1,30 @@
...
+apiVersion: projectcalico.org/v3
+kind: GlobalNetworkPolicy
+metadata:
+  name: deny-apps-policy
```
danail-branekov, line 5: How about setting the network policy name to something that describes that app egress should be denied, for example `deny-apps-egress-policy`? That name should be also reflected into the file name as well

scripts/assets/calico-allow-policy.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: projectcalico.org/v3
+kind: GlobalNetworkPolicy
+metadata:
+  name: allow-apps
```
danail-branekov, line 4: Maybe `allow-apps-egress` to reflect the intention of the policy?

danail-branekov (commented):
Looks good, minor comments/questions

helm/korifi/templates/global-deny-apps-policy.yaml
```diff
@@ -0,0 +1,30 @@
...
+      selector: 'k8s-app == "kube-dns"'
+      ports:
+      - 53
+  - action: Allow
+    protocol: TCP
+    destination:
+      selector: 'k8s-app == "kube-dns"'
+      ports:
+      - 53
+  - action: Allow
```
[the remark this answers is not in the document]
marsteg: local cluster resources seem to be treated differently. I needed to keep this (however moved it to the deploy-on-kind)

tests/dependencies/calico/kustomization.yaml
```diff
@@ -0,0 +1,13 @@
+resources:
+  - custom-resources.yaml
+
+patches:
+  - patch: |-
+      - op: replace
+        path: /spec/calicoNetwork/ipPools/0/cidr
+        value: 10.244.0.0/16
```
danail-branekov, line 8: We wonder whether patching that cidr to `10.244.0.0/16` would work for our CI clusters (running on GCP and AWS)? What is the purpose of this replace?

danail-branekov (approved):

tests/dependencies/calico/kustomization.yaml
marsteg: The default cidr of calico is the 192.168.0.0/16 range. I thought we're explicitly running on a 10. range but i just realize these were just my test installations. I think a 10.* range would be more common in production but i see that it is not really needed. Should I remove it?

danail-branekov: I don't know, you are the networking expert :)

If that helps, on our GCP cluster services' IPs start with `34.`

PR #4142, opened 2025-08-21
fix(deploy-on-kind): increase timeout for Helm deployment

geigerj0:
# Problem 
On my machine, deploying the Helm release took more than the [default `5m`](https://helm.sh/docs/intro/using_helm/#helpful-options-for-installupgraderollback) to run the `post-install`-hook. 

Screenshot 2025-08-21 at 15 31 04

# Solution 
Increasing the `--timeout` to `15m` made the kind deployment pass on my machine.

### DIFF
--- a/scripts/deploy-on-kind.sh
+++ b/scripts/deploy-on-kind.sh
@@ -212,6 +212,7 @@ function deploy_korifi() {
       --set=experimental.securityGroups.enabled="true" \
       --set=experimental.managedServices.trustInsecureBrokers="true" \
       --set=api.list.defaultPageSize="5000" \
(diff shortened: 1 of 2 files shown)


danail-branekov:
thanks @geigerj0 
Out of curiosity, do you see those timeouts when creating a new cluster, or during an upgrade of an existing deployment?

geigerj0:
> thanks @geigerj0 Out of curiosity, do you see those timeouts when creating a new cluster, or during an upgrade of an existing deployment?

I ran into these issues deploying the Helm release for the first time on a fresh kind-cluster.

Issue #4123, opened 2025-08-11
Filtering routes by more than one app guid yields no results


georgethebeatle:
### Reproduction

- push an app
- run `cf apps`
- app route is displayed in last column
- push another app
- run `cf apps`
- two apps are displayed, but with no routes
- delete on app
- run `cf apps`
- route is visible again

### Dev Notes

- When there is a single app the cli requests `GET /v3/routes?app_guids=<app-guid>`, which yields a single result
- When there are more than one apps the cli requests `GET /v3/routes?app_guids=<app-guid-1>,<app-guid-2>`, which yields no results
- It looks like the two values of the `app-guids` query parameter are being ANDed instead of ORed (It looks like kubernetes labels cannot be ORed, therefore we will need to do the filtering by app guid on the client side).
- TODO on top of the wip branch below:
  - Remove the constant `DestinationAppGUIDLabelPrefix` and fix the
  compilation errors - we do not need that label anymore
  - Good tests in the repository that focus on filtering by app

PR #4162, opened 2025-09-16
Filter routes by app guid in klient

uzabanov:
#4123

## What is this change about?
Filtering the routes in klient

No


danail-branekov:
e2e tests are panicing: https://ci.korifi.cf-app.com/builds/38360107#L68a3dd5b:10631

api/repositories/route_repository_test.go
```diff
@@ -231,6 +233,31 @@ var _ = Describe("RouteRepository", func() {
 				k8sClient.Create(ctx, cfRoute),
 			).To(Succeed())
 
+			appGUID1 = uuid.NewString()
```
danail-branekov, line 236: `appGUID1` is only used here, therefore can be inlined into line 252. Then the `appGUID1` variable can be removed

danail-branekov (commented):
We have left some comments. 

One additional important remark is that we need to guarantee that existing routes would get the annotation as well. Therefore, we need to adjust the migration procedure to touch existing routes, similar to what we had with the previous release

danail-branekov (commented):
We have left some comments. 

One additional important remark is that we need to guarantee that existing routes would get the annotation as well. Therefore, we need to adjust the migration procedure to touch existing routes, similar to what we had with the previous release. I.e. just leaving `CFRouteList` only in [this list](https://github.com/cloudfoundry/korifi/blob/8d0fb9d35ad16b4d57283fc1f61229abf3d9795b/migration/migration/executor.go#L19) should suffice

api/repositories/k8sklient/lister_test.go
```diff
@@ -252,4 +252,48 @@ var _ = Describe("Lister", func() {
...
+			Expect(fakeDescriptor.GUIDsCallCount()).To(Equal(1))
+			Expect(objectListMapper.GUIDsToObjectListCallCount()).To(Equal(1))
+		})
+
+		It("filters the objects", func() {
+			Expect(err).NotTo(HaveOccurred())
+			Expect(fakeDescriptor.FilterCallCount()).To(Equal(1))
+			by, filterFunc := fakeDescriptor.FilterArgsForCall(0)
+			Expect(by).To(Equal("foo"))
+			Expect(filterFunc).NotTo(BeNil())
```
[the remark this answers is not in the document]
uzabanov: this is not possible because the func var points to different address

api/repositories/list_options_test.go
```diff
@@ -288,6 +288,22 @@ var _ = Describe("ListOptions", func() {
...
+				option = repositories.WithFiltering("field", func(value any) bool {
+					return value == "value"
+				})
+			})
+
+			It("adds the filter by field and function to the list options", func() {
+				Expect(applyToListErr).NotTo(HaveOccurred())
+				Expect(listOptions.Filter).To(PointTo(MatchFields(IgnoreExtras, Fields{
+					"By":         Equal("field"),
+					"FilterFunc": Not(BeNil()),
```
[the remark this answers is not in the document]
uzabanov: The same problem here...

api/repositories/filter/route_apps_filtering_test.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/google/uuid"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"code.cloudfoundry.org/korifi/api/repositories/filter"
+)
+
+var _ = Describe("RoutesByAppGUID", func() {
+	var (
+		appGUID1     string
```
danail-branekov, line 13: `appGUID1` is never initialised, i.e. its default value `""` is being used. It should be set to a random uuid in `BeforeEach`

api/repositories/route_repository_test.go
```diff
@@ -1119,6 +1164,7 @@ var _ = Describe("RouteRepository", func() {
 					))
 					Expect(routeRecord.Annotations).To(Equal(
 						map[string]string{
+							korifiv1alpha1.CFRouteAppGuidsAnnotationKey: "",
```
danail-branekov, line 1167: In this test we actually care what happens with the annotations specified in the annotations patch, we do not really care about the rest (except that they do not change). Checking whether the `korifiv1alpha1.CFRouteAppGuidsAnnotationKey` annotation is set to an empty string looks weird, provided that neither the annotation patch specify it, nor it was set when the object was created.

Therefore, how about changing this expect to 

```golang
Expect(routeRecord.Annotations).To(SatisfyAll(
  HaveKeyWithValue("before-key-one": "value-one"),
  HaveKeyWithValue("key-one": "value-one-updated"),
  HaveKeyWithValue("key-two": "value-two"),
  Not(HaveKey("before-key-two")),
))
```

We could probably apply that pattern to the labels above as well

api/repositories/route_repository_test.go
```diff
@@ -1137,6 +1183,7 @@ var _ = Describe("RouteRepository", func() {
 					))
 					Expect(updatedCFRoute.Annotations).To(Equal(
 						map[string]string{
+							korifiv1alpha1.CFRouteAppGuidsAnnotationKey: "",
```
danail-branekov, line 1186: Same comment about checking annotations as above

danail-branekov (commented):
A couple of comments on tests, looks good otherwise

api/repositories/route_repository_test.go
uzabanov: Actually i added this because the test was failing and the empty korifiv1alpha1.CFRouteAppGuidsAnnotationKey
is set when object is created. Maybe I should handle it on webhook not to set the empty annotation on Create when no destinations?

danail-branekov: The purpose of the webhook is to ensure that the annotation is set consistently, so it is fine.
My comment was towards the way the keys of the route annotations map are being checked by the test

uzabanov: Got it

PR #4167, opened 2025-09-17
Implement PATCH /v3/service_instances for managed services

ybykov-a9s:
The PR is aimed to solve this [Issue 4163](https://github.com/cloudfoundry/korifi/issues/4163)

## What is this change about?
Implementing PATCH /v3/service_instances for managed services

It shouldn't be a breaking change

## Acceptance Steps
The PR is still WIP

[danail-branekov](https://github.com/danail-branekov)


api/repositories/service_instance_repository.go
```diff
@@ -77,7 +77,27 @@ type PatchServiceInstanceMessage struct {
 	MetadataPatch
 }
 
-func (p PatchServiceInstanceMessage) Apply(cfServiceInstance *korifiv1alpha1.CFServiceInstance) {
+type PatchManagedSIMessage struct {
+	GUID        string
+	SpaceGUID   string
+	PlanGUID    string
```
danail-branekov, line 83: I wonder whether `PlanGUID` should not be a pointer? A patch may want to update the metadata, but on the plan

api/repositories/service_instance_repository.go
```diff
@@ -77,7 +77,27 @@ type PatchServiceInstanceMessage struct {
...
+	if p.Name != nil {
+		cfServiceInstance.Spec.DisplayName = *p.Name
+	}
+	if p.Tags != nil {
+		cfServiceInstance.Spec.Tags = *p.Tags
+	}
+	p.MetadataPatch.Apply(cfServiceInstance)
+}
+
+func (p PatchManagedSIMessage) Apply(cfServiceInstance *korifiv1alpha1.CFServiceInstance) {
```
danail-branekov, line 100: I guess this function should set the plan on the instance spec, right?

api/repositories/service_instance_repository.go
```diff
@@ -274,7 +294,40 @@ func (r *ServiceInstanceRepo) servicePlanVisible(ctx context.Context, planGUID s
...
+		err = r.patchCredentialsSecret(ctx, cfServiceInstance, *message.Credentials)
+		if err != nil {
+			return ServiceInstanceRecord{}, apierrors.FromK8sError(err, ServiceInstanceResourceType)
+		}
+	}
+
+	return cfServiceInstanceToRecord(*cfServiceInstance), nil
+}
+
+func (r *ServiceInstanceRepo) PatchManagedServiceInstance(ctx context.Context, authInfo authorization.Info, message PatchManagedSIMessage) (ServiceInstanceRecord, error) {
```
danail-branekov, line 330: If I recall correctly, credentials make only sense for UPSIs, therefore dealing with the credentials secret makes no sense for managed services. Speaking of that, the patch message for managed services probably does not need the `Credentials` field

controllers/controllers/services/instances/managed/controller.go
```diff
@@ -166,18 +165,33 @@ func (r *Reconciler) ReconcileResource(ctx context.Context, serviceInstance *kor
 		serviceInstance.Spec.ServiceLabel = tools.PtrTo(serviceInstanceAssets.ServiceOffering.Spec.Name)
 	}
 
-	provisionResponse, err := r.provisionServiceInstance(ctx, serviceInstance, serviceInstanceAssets, osbapiClient)
-	if err != nil {
-		log.Error(err, "failed to provision service instance")
-		return ctrl.Result{}, fmt.Errorf("failed to provision service instance: %w", err)
-	}
+	if serviceInstance.Status.ObservedGeneration == 0 {
```
danail-branekov, line 168: If I get the intention of this check correctly, the idea here is to perform `provision` when the CFServiceInstance has been just created, otherwise proceed to update.

The generation might drift because of whatever updates (maybe a new label appeared) that may occur while provisioning is already running. 

Instead of relying on the observed generation, I would suggest to add `PlanGUID` to `CFServiceInstance.Status`. Then the plan guid on the spec would have the semantics of "desired plan", while the plan guid on the status would mean "actual plan". Then if both differ, the controller should perform a plan update. Once the update succeed, the controller should set the new plan guid on the status, thus "writing down" that this is already the actual plan.

Also, it is worth mentioning that once the instance is ready (`isReady` function returns true), the controller would stop listening for updates. Therefore you should change the function to also check whether the "desired plan" matches the "actual plan"

controllers/controllers/services/instances/managed/controller.go
```diff
@@ -249,6 +263,7 @@ func (r *Reconciler) processProvisionOperation(
 	lastOpResponse osbapi.LastOperationResponse,
 ) (ctrl.Result, error) {
 	if lastOpResponse.State == "succeeded" {
+		setObservedGeneration(serviceInstance)
```
danail-branekov, line 266: setting the observed generation is only committed to the etcd if the reconcile method succeeds. This is why the controller sets it on the very beginning of the reconcile method. If it succeeds, it is committed, if it fails, or needs requeue - it is discarded. Setting it here violates that pattern.

I believe you set it here to figure out whether to provision or update in the reconcile method, but I think my suggestion for the plan guid in the status would provide you with a more reliable criteria

controllers/controllers/services/instances/managed/controller.go
```diff
@@ -261,16 +276,88 @@ func (r *Reconciler) processProvisionOperation(
...
+			PlanID:    assets.ServicePlan.Spec.BrokerCatalog.ID,
+		},
+	})
+	if err != nil {
+		log.Error(err, "failed to update service")
+
+		if osbapi.IsUnrecoveralbeError(err) {
+			serviceInstance.Status.LastOperation.State = "failed"
+			meta.SetStatusCondition(&serviceInstance.Status.Conditions, metav1.Condition{
+				Type:               korifiv1alpha1.UpdateFailedCondition,
```
danail-branekov, line 314: I guess that if the update failed with unrecoverable error, there is no point for the controller to retry update, correct?

If so, you could ammend the `isFailed` function to return try if the `UpdateFailedCondition` is true in order to prevent further update attempts with the broker

danail-branekov (commented):
Looks promising. The most important bits of my comments are on the managed instance controller

controllers/controllers/services/instances/managed/controller.go
ybykov-a9s: The same - changed logic to use Status.

controllers/controllers/services/instances/managed/controller.go
```diff
@@ -261,16 +276,88 @@ func (r *Reconciler) processProvisionOperation(
...
+	lastOpResponse osbapi.LastOperationResponse,
+) (ctrl.Result, error) {
+	if lastOpResponse.State == "succeeded" {
+		setObservedGeneration(serviceInstance)
+		return ctrl.Result{}, nil
+	}
+
+	if lastOpResponse.State == "failed" {
+		meta.SetStatusCondition(&serviceInstance.Status.Conditions, metav1.Condition{
+			Type:               korifiv1alpha1.ProvisioningFailedCondition,
```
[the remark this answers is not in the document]
ybykov-a9s: Thank you for the catch.

controllers/controllers/services/instances/managed/controller_test.go
```diff
@@ -753,6 +783,264 @@ var _ = Describe("CFServiceInstance", func() {
 		})
 	})
 
+	When("updates instance", func() {
```
danail-branekov: Wording: would recommend `When("the service instance is being updated", ...`

controllers/controllers/services/instances/managed/controller_test.go
```diff
@@ -753,6 +783,264 @@ var _ = Describe("CFServiceInstance", func() {
...
+				meta.SetStatusCondition(&instance.Status.Conditions, metav1.Condition{
+					Type:    korifiv1alpha1.StatusConditionReady,
+					Status:  metav1.ConditionFalse,
+					Reason:  "UpdateRequested",
+					Message: "managed service instance update is requested",
+				})
+			})).To(Succeed())
+		})
+
+		It("sets the ObservedGeneration status field", func() {
```
danail-branekov: Setting the observed generation is something the controller does regardless whether the instance is updated or not, so having this `It` here does not really make sense

danail-branekov (commented):
See my comments below, significant bits are on the controller. I think you are going in the right direction.

A general style comment - we usually put a blank line between ginkgo blocks - `Describe`, `When`, `It`, etc to visually separate them. If you could keep that convention in the PR, that would be great.

controllers/api/v1alpha1/cfserviceinstance_types.go
```diff
@@ -90,6 +91,10 @@ type CFServiceInstanceStatus struct {
 	//+kubebuilder:validation:Optional
 	MaintenanceInfo MaintenanceInfo `json:"maintenanceInfo"`
 
+	// The service instance actual plan
```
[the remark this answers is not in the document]
ybykov-a9s: done

controllers/controllers/services/instances/managed/controller.go
```diff
@@ -472,9 +561,14 @@ func (r *Reconciler) getNamespace(ctx context.Context, namespaceName string) (*c
 }
 
 func isFailed(instance *korifiv1alpha1.CFServiceInstance) bool {
-	return meta.IsStatusConditionTrue(instance.Status.Conditions, korifiv1alpha1.ProvisioningFailedCondition)
+	return meta.IsStatusConditionTrue(instance.Status.Conditions, korifiv1alpha1.ProvisioningFailedCondition) ||
+		meta.IsStatusConditionTrue(instance.Status.Conditions, korifiv1alpha1.UpdateFailedCondition)
```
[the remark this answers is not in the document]
ybykov-a9s: yeah... Removed it again :D 
TBF there might be use cases where even ProvisionFailed might be fixed by a consequent Update. But I wouldn't concentrate on it atm.

controllers/controllers/services/instances/managed/controller_test.go
```diff
@@ -102,6 +103,29 @@ var _ = Describe("CFServiceInstance", func() {
 		}
 		Expect(adminClient.Create(ctx, servicePlan)).To(Succeed())
 
+		servicePlan2 = &korifiv1alpha1.CFServicePlan{
```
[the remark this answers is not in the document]
ybykov-a9s: done

controllers/controllers/services/instances/managed/controller_test.go
```diff
@@ -753,6 +783,264 @@ var _ = Describe("CFServiceInstance", func() {
...
+		})
+
+		It("sets the ObservedGeneration status field", func() {
+			Eventually(func(g Gomega) {
+				g.Expect(adminClient.Get(ctx, client.ObjectKeyFromObject(instance), instance)).To(Succeed())
+				g.Expect(instance.Status.ObservedGeneration).To(Equal(instance.Generation))
+			}).Should(Succeed())
+		})
+
+		It("sets the PlanGUID status field", func() {
```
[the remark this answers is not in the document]
ybykov-a9s: done. again - an original text from Provision.

controllers/webhooks/networking/security_groups/validator_test.go
```diff
@@ -438,7 +438,6 @@ var _ = Describe("CFSecurityGroupValidatingWebhook", func() {
 				))
 			})
 		})
-
```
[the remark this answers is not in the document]
ybykov-a9s: I bet my local automation did it when i touched the file. I wasn't sure if it should be included to the commit or not. No problem - removed it to decrease diff.

tests/e2e/e2e_suite_test.go
```diff
@@ -757,6 +757,46 @@ func createServiceInstance(spaceGUID, name string, credentials map[string]string
 	return serviceInstance.GUID
 }
 
+func createManagedServiceInstance(brokerGUID, spaceGUID string, name string) string {
```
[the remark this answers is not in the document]
ybykov-a9s: Existing createUPServiceInstance (renamed from the original createServiceInstance) a bit above has it. So i preserved the logic.

tests/e2e/service_instances_test.go
```diff
@@ -179,21 +192,91 @@ var _ = Describe("Service Instances", func() {
...
+				Expect(serviceInstance.Tags).To(ConsistOf("some", "tags"))
+			})
+			It("changes a plan", func() {
+				expectJobCompletes(httpResp)
+
+				httpRespService, httpErrorService := adminClient.R().SetResult(&result).Get("/v3/service_instances/" + serviceInstanceGUID)
+				Expect(httpErrorService).NotTo(HaveOccurred())
+				Expect(httpRespService).To(HaveRestyStatusCode(http.StatusOK))
+				Expect(
+					result.resource.Relationships["service_plan"].Data.GUID,
```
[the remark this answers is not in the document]
ybykov-a9s: It's always a dilemma - make more specific tests or combine them to minimize their number :D
In this case i would agree that your way makes more sense.

controllers/controllers/services/instances/managed/controller.go
```diff
@@ -267,6 +283,74 @@ func (r *Reconciler) processProvisionOperation(
...
+			serviceInstance.Status.LastOperation.State = "failed"
+			meta.SetStatusCondition(&serviceInstance.Status.Conditions, metav1.Condition{
+				Type:               korifiv1alpha1.UpdateFailedCondition,
+				Status:             metav1.ConditionTrue,
+				ObservedGeneration: serviceInstance.Generation,
+				LastTransitionTime: metav1.NewTime(time.Now()),
+				Reason:             "UpdateFailed",
+				Message:            err.Error(),
+			})
+			return osbapi.UpdateResponse{},
```
danail-branekov: If the error is not recovarable (we are in the `if osbapi.IsUnrecoveralbeError(err)` case), does it make sense for the controller to retry? If not you could prevent that by

```
			return osbapi.UpdateResponse{},
				k8s.NewNotReadyError().WithReason("UpdateFailed").WithNoRequeue()
```

danail-branekov (commented):
I think you are making a steady progress, there are just few things to iron out

controllers/controllers/services/instances/managed/controller.go
ybykov-a9s: It's better to not to requeue UpdateFailed. There might be many reasons to not to do that.
Updated as you recommended including a test.

danail-branekov (commented):
Today I spent some time checking out how the update works. I am afraid that I found out that when updating the service instance (via `cf update-service managed-service  -p sample-2`), no communication with the broker is performed.

While investigating the issue I took the liberty to come up with a diff that addresses that (it also addresses the comments below):

```
diff --git a/controllers/controllers/services/instances/managed/controller.go b/controllers/controllers/services/instances/managed/controller.go
index bcf440a5..c2f52bcb 100644
--- a/controllers/controllers/services/instances/managed/controller.go
+++ b/controllers/controllers/services/instances/managed/controller.go
@@ -143,10 +143,6 @@ func (r *Reconciler) ReconcileResource(ctx context.Context, serviceInstance *kor
 
 	serviceInstance.Status.UpgradeAvailable = serviceInstance.Status.MaintenanceInfo.Version != serviceInstanceAssets.ServicePlan.Spec.MaintenanceInfo.Version
 
-	if isReady(serviceInstance) {
-		return ctrl.Result{}, nil
-	}
-
 	if isProvisioningFailed(serviceInstance) {
 		return ctrl.Result{}, k8s.NewNotReadyError().WithReason("ServiceInstanceFailed").WithNoRequeue()
 	}
@@ -220,7 +216,6 @@ func (r *Reconciler) reconcileUpdatedServiceInstance( //nolint:dupl
 		return r.processUpdateOperation(serviceInstance, lastOpResponse)
 	}
 
-	serviceInstance.Status.Provisioned = true
 	serviceInstance.Status.MaintenanceInfo = serviceInstanceAssets.ServicePlan.Spec.MaintenanceInfo
 	serviceInstance.Status.LastOperation.State = "succeeded"
 	return ctrl.Result{}, nil
@@ -584,7 +579,3 @@ func (r *Reconciler) getNamespace(ctx context.Context, namespaceName string) (*c
 func isProvisioningFailed(instance *korifiv1alpha1.CFServiceInstance) bool {
 	return meta.IsStatusConditionTrue(instance.Status.Conditions, korifiv1alpha1.ProvisioningFailedCondition)
 }
-
-func isReady(instance *korifiv1alpha1.CFServiceInstance) bool {
-	return meta.IsStatusConditionTrue(instance.Status.Conditions, korifiv1alpha1.StatusConditionReady)
-}
diff --git a/controllers/controllers/services/instances/managed/controller_test.go b/controllers/controllers/services/instances/managed/controller_test.go
index 256641fb..d1537a79 100644
--- a/controllers/controllers/services/instances/managed/controller_test.go
+++ b/controllers/controllers/services/instances/managed/controller_test.go
@@ -793,21 +793,11 @@ var _ = Describe("CFServiceInstance", func() {
 		JustBeforeEach(func() {
 			Eventually(func(g Gomega) {
 				g.Expect(adminClient.Get(ctx, client.ObjectKeyFromObject(instance), instance)).To(Succeed())
-				g.Expect(instance.Status.Conditions).To(ContainElement(SatisfyAll(
-					HasType(Equal(korifiv1alpha1.StatusConditionReady)),
-					HasStatus(Equal(metav1.ConditionTrue)),
-				)))
 				g.Expect(instance.Status.Provisioned).To(BeTrue())
 			}).Should(Succeed())
 
 			Expect(k8s.Patch(ctx, adminClient, instance, func() {
 				instance.Spec.PlanGUID = servicePlan2.Name
-				meta.SetStatusCondition(&instance.Status.Conditions, metav1.Condition{
-					Type:    korifiv1alpha1.StatusConditionReady,
-					Status:  metav1.ConditionFalse,
-					Reason:  "UpdateRequested",
-					Message: "managed service instance update is requested",
-				})
 			})).To(Succeed())
 		})
 
@@ -823,6 +813,20 @@ var _ = Describe("CFServiceInstance", func() {
 			}).Should(Succeed())
 		})
 
+		It("sends an update request to the broker", func() {
+			Eventually(func(g Gomega) {
+				g.Expect(brokerClient.UpdateCallCount()).To(BeNumerically(">=", 1))
+				_, actualUpdatePayload := brokerClient.UpdateArgsForCall(brokerClient.UpdateCallCount() - 1)
+				g.Expect(actualUpdatePayload).To(Equal(osbapi.UpdatePayload{
+					InstanceID: instance.Name,
+					UpdateRequest: osbapi.UpdateRequest{
+						ServiceId: "service-offering-id",
+						PlanID:    "service-plan-id-2",
+					},
+				}))
+			}).Should(Succeed())
+		})
+
 		When("service update fails with recoverable error", func() {
 			BeforeEach(func() {
 				brokerClient.UpdateReturns(osbapi.UpdateResponse{}, errors.New("update-failed"))
@@ -831,7 +835,7 @@ var _ = Describe("CFServiceInstance", func() {
 			It("keeps trying to update the instance", func() {
 				Eventually(func(g Gomega) {
 					g.Expect(brokerClient.UpdateCallCount()).To(BeNumerically(">=", 1))
-					_, updatePayload := brokerClient.UpdateArgsForCall(1)
+					_, updatePayload := brokerClient.UpdateArgsForCall(brokerClient.UpdateCallCount() - 1)
 					g.Expect(updatePayload).To(Equal(osbapi.UpdatePayload{
 						InstanceID: instance.Name,
 						UpdateRequest: osbapi.UpdateRequest{
```

While this diff ensures the broker's update method is invoked, it breaks existing tests. The reason is that when a k8s object is being created, it could be reconciled a couple of times, hence a provisioned service could yield a noop update with the broker. While this should not be an issue from OSBAPI spec point of view (still, no idea how real-life brokers react to noop-updates), this has the consequence that when a service instance is created, its last operation is reported as `update` rather than `create`.

If we are sure brokers do not mind noop-updates, we could probably ignore the effect above as an "usability issue" within this PR. Still, existing tests should be adapted.

If we want to solve the issue and avoid noop updates with the broker, I guess we need to replicate fields that can be updated to the status. Thus the values on the spec would have the semantic of "desired state", while the ones in the status would mean "actual state". The user sets the field on the spec, while the controller communicates the "desired state" with the broker and updates the status ("actual state") depending on the outcome.
Having those fields in the status would let the controller only perform broker updates that are noop. The criteria of an update with the broker would be `desired-state != actual_state`

Now, the question is what fields from the spec we should put on the status. According to the [docs](https://docs.cloudfoundry.org/devguide/services/managing-services.html#update_service), those are plan, tags and parameters. I think that for the scope of this PR we could just add the plan guid

controllers/api/v1alpha1/cfserviceinstance_types.go
```diff
@@ -90,6 +91,18 @@ type CFServiceInstanceStatus struct {
...
+	//+kubebuilder:validation:Optional
+	PlanGUID string `json:"planGuid"`
+
+	// True if the service instance has been successfully provisioned. Only makes seense for managed service instances
+	//+kubebuilder:validation:Optional
+	Provisioned bool `json:"provisioned"`
+
+	// The generation at which the instance was last successfully reconciled with the broker.
+	//+kubebuilder:validation:Optional
+	BrokerReconciledGeneration int64 `json:"brokerReconciledGeneration,omitempty"`
```
danail-branekov, line 104: I had a look at the usages of this field and from what I see, it is only used in the controller to decide whether to perform update. 

From my point of view this pattern is quite imperative and does not convey the intention. Also, relying on generation to filter out noop broker updates would probably not work - a generation could change in case e.g. a label is added to the CFServiceInstance - that is not a change a broker is interested in.  

Would it not be better if we had a `needsUpdate` function in the controller instead that would (for now) return `true` only if `spec.PlanGUID != status.PlanGUID`. Then, the controller proceeds with updating the instance with the broker only if `needsUpdate` returns `true`
Such an approach would also remove setting the field all over the place and would be easier to reason about. 

Speaking of avoiding unnecessary updates with the broker - you could introduce a test in the controller tests that verify that no broker communication is performed when something irrelevant has changed, e.g. a label has been added to the CFServiceInstance

controllers/controllers/services/instances/managed/controller.go
```diff
@@ -157,15 +153,34 @@ func (r *Reconciler) ReconcileResource(ctx context.Context, serviceInstance *kor
 		return ctrl.Result{}, err
 	}
 
-	if !planVisible {
+	if !planVisible && !serviceInstance.Status.Provisioned {
```
danail-branekov, line 156: I guess the intention here is not to set failed status if the service has been provisioned but the plan has been disabled after that. Sounds reasonable, I did not see a test about that case though.

Provided the check only makes sense for unprovisioned services, I would suggest moving it to `reconcileProvisionedServiceInstance`

danail-branekov: Actually, we need to check for plan visibility on update as well. If the new desired plan is not visible, then the update should fail

controllers/api/v1alpha1/cfserviceinstance_types.go
ybykov-a9s: done

controllers/controllers/services/instances/managed/controller.go
ybykov-a9s: done

ybykov-a9s:
@danail-branekov I've made the changes you'd proposed. TBH the logic which triggers a broker update should be more broad. But I would say it's a one of the future topics.

controllers/controllers/services/instances/managed/controller_test.go
```diff
@@ -753,6 +763,329 @@ var _ = Describe("CFServiceInstance", func() {
...
+					InstanceID: instance.Name,
+					UpdateRequest: osbapi.UpdateRequest{
+						ServiceId: "service-offering-id",
+						PlanID:    "service-plan-id-2",
+					},
+				}))
+			}).Should(Succeed())
+		})
+
+		It("does not call the broker when only metadata changes", func() {
```
danail-branekov: I am afraid that this test might be racy as k8s might call the broker client asynchronously. In order to address this, how about the following structure:

```golang
Describe("service instance update", 
   JustBeforeEach(func(){
        // update metadata (i.e. no updates with broker
    })

    It("does not perform broker update", func(){
    	// assert consistently broker update method is never invoked
     })

    When("the plan is updated", func(){
        JustBeforeEach(func(){
             // update the plan
        })
          
        It("updates the instance with the broker", func(){..})

         When("the plan is not visible", func(){
              // expect an error
         })

         // al the other contexts/tests that verify broker update interaction go here
     })
  
```

controllers/controllers/services/instances/managed/controller_test.go
```diff
@@ -753,6 +763,329 @@ var _ = Describe("CFServiceInstance", func() {
...
+					g.Expect(brokerClient.UpdateCallCount()).To(BeNumerically(">=", 1))
+					g.Expect(instance.Status.LastOperation).To(Equal(korifiv1alpha1.LastOperation{
+						Type:  "update",
+						State: "failed",
+					}))
+				}).Should(Succeed())
+
+				currentCallCount := brokerClient.UpdateCallCount()
+				Consistently(func(g Gomega) {
+					g.Expect(brokerClient.UpdateCallCount()).To(Equal(currentCallCount))
```
danail-branekov: I think hardcoding `1` instead of using `currentCallCount` would be more explicit. We thus emphasize that one the update fails with unrecoverable error, the controller give up updating and does not talk to the broker anymore

danail-branekov (commented):
There are few edges in the controller test that are worth ironing out, other than that looks good. 

While playing around with the change I wondered what should happen to a binding when a service instance plan is update. Do you know what happens in the CF on VMs world? Should the binding be updated with new credentials if an instance plan is updated? Or do old credentials should remain valid?

With your implementation the binding credentials do not change but I can easily see how this would be OK.

danail-branekov:
> TBH the logic which triggers a broker update should be more broad. But I would say it's a one of the future topics.

Agree. I think it makes sense to extend the condition that triggers broker update as we add new features. From my point of view, triggering update on plan update only is adequate for the scope of this PR

controllers/controllers/services/instances/managed/controller_test.go
ybykov-a9s: I adjusted the test.

controllers/controllers/services/instances/managed/controller_test.go
ybykov-a9s: Agree. Adjusted.

ybykov-a9s:
@danail-branekov I adjusted parts you mentioned.
I also returned back the logic `if isUpdateFailed` to Reconciler. because updated tests didn't want to work :)

controllers/controllers/services/instances/managed/controller.go
```diff
@@ -143,27 +163,46 @@ func (r *Reconciler) ReconcileResource(ctx context.Context, serviceInstance *kor
...
-		return ctrl.Result{}, k8s.NewNotReadyError().WithReason("ProvisioningFailed").WithNoRequeue()
+	if !serviceInstance.Status.Provisioned {
+		return r.reconcileProvisionedServiceInstance(ctx, serviceInstance, serviceInstanceAssets, osbapiClient)
 	}
 
-	planVisible, err := r.isServicePlanVisible(ctx, serviceInstance, serviceInstanceAssets.ServicePlan)
-	if err != nil {
-		log.Error(err, "failed to check service plan visibility")
-		return ctrl.Result{}, err
+	// Backfill status.PlanGUID for legacy provisioned instances without it (one-time migration)
```
danail-branekov, line 179: I am a bit on the fence about backfilling the status. For legacy provisioned instances, without this backfill, the controller would call update on the instance, which I think should be a noop from the broker point of view. Then `reconcileUpdatedServiceInstance` would backfill the status (it would be invoked as spec.Plan != status.Plan)

Furthermore, managed services are an "experimental" feature, I wonder whether it is worth polluting the controller code with migration logic. If backfilling the status is absolutely needed, I think it would be better to put that backfill in the migration component (and delete the migration code next release) rather than having it in the controller forever

controllers/controllers/services/instances/managed/controller.go
```diff
@@ -143,27 +163,46 @@ func (r *Reconciler) ReconcileResource(ctx context.Context, serviceInstance *kor
...
-	if !planVisible {
-		return ctrl.Result{},
-			k8s.NewNotReadyError().WithMessage("The service plan is disabled").WithReason("InvalidServicePlan").WithNoRequeue()
+	if !needsUpdate(serviceInstance) {
+		return ctrl.Result{}, nil
 	}
 
-	if serviceInstance.Spec.ServiceLabel == nil {
-		serviceInstance.Spec.ServiceLabel = tools.PtrTo(serviceInstanceAssets.ServiceOffering.Spec.Name)
+	if isUpdateFailed(serviceInstance) {
```
danail-branekov, line 189: That would block the instance from a subsequent update (i.e. an update failed is a terminal state for the instance). For example, imagine we first try to update with invalid plan -> update fails. Then we realise our mistake, and update it with a valid plan - should not that cause successful update?

controllers/controllers/services/instances/managed/controller_test.go
```diff
@@ -753,6 +763,330 @@ var _ = Describe("CFServiceInstance", func() {
...
+						g.Expect(instance.Status.Conditions).To(ContainElement(SatisfyAll(
+							HasType(Equal(korifiv1alpha1.StatusConditionReady)),
+							HasStatus(Equal(metav1.ConditionFalse)),
+							HasReason(Equal("InvalidServicePlan")),
+						)))
+					}).Should(Succeed())
+
+					Consistently(func(g Gomega) {
+						g.Expect(brokerClient.UpdateCallCount()).To(BeZero())
+					}).WithTimeout(time.Second).Should(Succeed())
```
danail-branekov, line 881: I think the e2e suite has a default consistently timeout, therefore specifying explicit consistently timeout here should not be needed

danail-branekov (commented):
Looks good. My only concern is that a failed update would become a terminal state, i.e. further updates would not be possible. Is that on purpose?

Issue #2328, opened 2023-03-21
[Bug]: Korifi default domain guid must be a valid uuid


danail-branekov:
### Blockers/Dependencies

No response

### Background

Currently the default cfdomain object is crated by the helm chart with a hardcoded `metadata.name` of `default-domain`. Strictly speaking the `metadata.name` of each CF CR should represent its CF guid and should be a vaild uuid.

### Acceptance Criteria

GIVEN I have korifi deployed
WHEN I run `kubectl get -n cf cfdomains.korifi.cloudfoundry.org`
THEN I see a list of domains and every domain name is a valid uuid

### Dev Notes

No response

PR #4168, opened 2025-09-17
made metadata.name dynamic

thomasbscj:
Issue #2328

## What is this change about?
This PR makes every Domain have a unique ID

## Acceptance Steps
deploy korifi
cf create-shared-domain mynew.domain
cf create-shared-domain other.domain
kubectl get cfroutes -A

every domain should have unique IDs

### DIFF
--- a/helm/korifi/controllers/crds/korifi.cloudfoundry.org_appworkloads.yaml
+++ b/helm/korifi/controllers/crds/korifi.cloudfoundry.org_appworkloads.yaml
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.19.0
+    controller-gen.kubebuilder.io/version: v0.18.0
   name: appworkloads.korifi.cloudfoundry.org
 spec:
   group: korifi.cloudfoundry.org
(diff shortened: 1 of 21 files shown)


georgethebeatle (commented):
While reviewing this PR we have realised that there is no easy way to change the default domain name without breaking all existing deployments of Korifi.

There are a few facts that lead to the above conclusion:
- It is not allowed to create domains with overlapping DNS names, e.g. `example.com` and `my.example.com` (it would result in validation error). 
- Every domain "owns" the routes associated with it, which means that when the domain is deleted, all its routes are deleted as well.
- Every route has an immutable reference to its parent domain, meaning that once created a route cannot move to another domain.

If we want to change the default domain name to a valid uuid we should thin about existing deployments. 
- On an existing deployment there will already be a domain called `default-domain` that is associated to a DNS name, e.g. `example.com` as well as routes that belong to that domain.
- On update the post install job will try to create a domain with name some uuid and DNS `example.com` (or whatever is passed as value to the helm chart) and this will result in a validation error as explained in the facts section above.
- Therefore we need to invent a migration procedure for existing deployments.
- One strategy could be to delete the domain before creating the new one. However this is unacceptable since it means that all routes will be wiped out.
- Alternatively we could remove the finalizer of the existing `default-domain` so that we "leak" the routes and then create the new domain. This would work, because there will be no conflict in the DNS names. However in order to restore routing we would have to patch all existing routes so that they point to the new domain. This is not possible, because the domain reference of the routes is an immutable field.

Because of all this we think we should park this story for now and rethink our approach. Sorry about that.

Issue #4149, opened 2025-09-01
As a Korifi API user I want to be able to filter service bindings by display name


georgethebeatle:
### Background

According to the cf v3 api [docs](https://v3-apidocs.cloudfoundry.org/version/3.159.0/index.html#get-a-service-credential-binding), service bindings can be filtered by display name

### Acceptance

**GIVEN** I have created a managed service instance myservice
**WHEN** I create bindings `b1`, `b2` and `b3` by running `cf bind-service myapp myservice --binding-name BINDING_NAME`
**WHEN** I run `cf curl /v3/service_credential_bindings?names=b1,b3`
**THEN** I should only see bindings `b1` and `b3` in the response

Alternatively

**GIVEN** I have created a managed service instance
**WHEN** I create a service key by running: `cf create-service-key myservice mykey`
**WHEN** I run `cf service-key myservice mykey`
**THEN** I see the service instance credentials

### Dev Notes
- This feature is needed in order to enable the `cf service-key` command
- Add smoke test, as this is a newly supported command
- WIP commit: https://github.com/cloudfoundry/korifi/commit/57bca5ef581dd60458c112a2913a49758892faff

PR #4169, opened 2025-09-18
Filter service bindings by display name

uzabanov:
#4149

## What is this change about?
Filter service bindings by display name

No


tests/smoke/service_keys_test.go
```diff
@@ -0,0 +1,57 @@
...
+		)).To(Exit(0))
+		DeferCleanup(func() {
+			broker.NewDeleter(sharedData.RootNamespace).ForBrokerName(brokerName).Delete()
+		})
+
+		Expect(helpers.Cf("enable-service-access", "sample-service", "-b", brokerName)).To(Exit(0))
+		serviceKeysSession = helpers.Cf("create-service", "sample-service", "sample", serviceName, "-b", brokerName)
+		Expect(serviceKeysSession).To(Exit(0))
+
+		serviceKeysSession = helpers.Cf("create-service-key", serviceName, serviceKeyName)
```
georgethebeatle, line 43: The `create-service-key` command is asynchronous, so it is a good idea to call it with the `--wait` option to avoid races. Also it will be more idiomatic if we extract the `create-service-key` into a `JustBeforeEach` as it is the code being tested.

PR #4341, opened 2026-04-14
Implement rename org

ybykov-a9s:
## Summary

This PR completes the org rename fix and hardens metadata protection across Korifi objects described in this [Issue](https://github.com/cloudfoundry/korifi/issues/4245)

It includes three parts:

- Fix org rename handling for PATCH requests.
- Keep Danail’s context-aware [repository validation](https://github.com/cloudfoundry/korifi/commit/7965e315c14954aa614d61a5d51f402ad5ac12f4#diff-676de4e8a003e16a84e31ba23846cb034285b416424aca069fabe0c5ab9b55e6) for metadata patching.
- Add label-signing and a validating webhook to protect Korifi-managed labels at the Kubernetes API level.

## What changed

- Fixed org rename issues in the API patch flow.
- Kept metadata patch validation in repositories, where the code can compare incoming values with the current object and allow unchanged echoed metadata.
- Extended that repository-side validation/error propagation to other metadata patch paths.
- Added label signing for Korifi-managed labels.
- Added a validating webhook that rejects tampering with internal [korifi.cloudfoundry.org/*](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) labels and rejects non-Korifi [*.cloudfoundry.org](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) label keys.
- Added migration/backfill support and Helm wiring for the label-signing secret.
- Updated repository, webhook, e2e, and smoke coverage accordingly.

## Why repository-side validation should co-exist with the webhook solution

Both solutions protect different layers and should both remain in place.

Repository-side validation is still needed because:

- it gives the correct behavior for requests coming through the Korifi API
- it has object context, so it can distinguish between “client echoed existing metadata back unchanged” and “client is trying to modify protected metadata”
- it returns proper Korifi API errors from the handler/repository flow

The webhook solution is also needed because:

- repository validation only runs for requests that go through the Korifi API
- it does not protect direct writes to Korifi CRDs through the Kubernetes API
- the validating webhook blocks those direct mutations at admission time

In short:

- repository validation protects the Korifi API path
- webhooks protect the Kubernetes API path

Together they provide correct API semantics and cluster-level enforcement.

## Testing

- Added/updated repository tests for invalid metadata patch handling
- Added validating webhook integration tests
- Added e2e and smoke coverage for protected labels

### DIFF
--- a/controllers/webhooks/label_indexer/signer/signer.go
+++ b/controllers/webhooks/label_indexer/signer/signer.go
@@ -1,0 +1,37 @@
+package signer
+
+import (
+	"crypto/hmac"
+	"crypto/sha256"
+	"encoding/hex"
+	"sort"
+	"strings"
+)
+
+const ReservedLabelDomain = "korifi.cloudfoundry.org/"
+
+func Sign(secret []byte, labels map[string]string) string {
+	mac := hmac.New(sha256.New, secret)
+	for _, k := range sortedReservedKeys(labels) {
+		mac.Write([]byte(k))
+		mac.Write([]byte("="))
+		mac.Write([]byte(labels[k]))
+		mac.Write([]byte("\n"))
+	}
+	return hex.EncodeToString(mac.Sum(nil))
+}
+
+func Verify(secret []byte, labels map[string]string, sig string) bool {
+	return hmac.Equal([]byte(Sign(secret, labels)), []byte(sig))
+}
+
+func sortedReservedKeys(labels map[string]string) []string {
+	keys := make([]string, 0, len(labels))
+	for k := range labels {
+		if strings.HasPrefix(k, ReservedLabelDomain) {
+			keys = append(keys, k)
+		}
+	}
+	sort.Strings(keys)
+	return keys
+}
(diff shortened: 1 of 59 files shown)


controllers/webhooks/label_indexer/signer/signer.go
```diff
@@ -0,0 +1,37 @@
...
+	"crypto/sha256"
+	"encoding/hex"
+	"sort"
+	"strings"
+)
+
+const ReservedLabelDomain = "korifi.cloudfoundry.org/"
+
+func Sign(secret []byte, labels map[string]string) string {
+	mac := hmac.New(sha256.New, secret)
```
danail-branekov, line 14: I wonder whether using hmac signature with a secret is an overkill. Would it not be sufficient to sha256sum the reserved keys map? Then we would not need a secret that is seeded on helm install

controllers/webhooks/label_validator/webhook.go
```diff
@@ -0,0 +1,79 @@
...
+	// On first create (no existing signature yet) the label indexer has already
+	// run and written a fresh signature, so storedSig will be present.
+	// If it is still absent (e.g. pre-existing objects during rollout) we allow
+	// the request through to avoid blocking legitimate operations.
+	if storedSig == "" {
+		return admission.Allowed("no existing label signature")
+	}
+
+	if !signer.Verify(r.signingSecret, obj.GetLabels(), storedSig) {
+		return admission.Denied(fmt.Sprintf(
```
danail-branekov, line 64: Provided that in korifi we have a concept of returning webhook errors to the cli, see this [example](https://github.com/ybykov-a9s/korifi/blob/296b7ac2f7c2c7551ffd4fe0116bc07700cef879/controllers/webhooks/validation/duplicate_validator.go#L35), I wonder could we not do the same here? Then, if the error from the webhook is propagated to the cli, then we would not need to validate the metadata patch in the repositories. This would significantly reduce the amount of changes, and all the metadata patch validation tests in the api would become redundand

danail-branekov (commented):
I did not have the time to look in the change throughly, but it is great that you managed to implement a webhook validation.

I have two major comments:
* Computing the labels signature requires a secret which might be awkward with regards to lifecycle management
* If we manage to propagate the webhook error to the cli (see comments), then repositories should not explicitly need to validate the metadata patch (that would make repositories even simpler)

